Issue #248, opened 2026-04-17
feat: detect /workspace virtio-fs mount drop mid-run (silent disconnect)


aviadshiber:
## Problem

The `/workspace` virtio-fs mount can silently disconnect **during** container execution on macOS (Apple Hypervisor / applehv). This is different from #221, which covers the mount being missing at startup.

### Observed behavior (2026-04-17, agent `4fe40c`)

1. Container launched successfully, worktree created, `/workspace` fully accessible
2. ~30 minutes into execution, the mount silently dropped
3. All files under `/workspace/` became inaccessible
4. Every Bash command returned exit code 1 with no output (even `echo hello`)
5. `rg` (ripgrep) failed, Git unreachable — all tools broken
6. Agent detected the broken state, reported it, and exited cleanly (code 0)
7. User had to manually: wipe VM storage, restart Podman, rebuild images

### Root cause

macOS Apple Hypervisor Framework (applehv) Podman VM occasionally drops virtio-fs mounts under load. The VM process stays running but the mount goes stale. This is a known Podman-on-macOS issue.

### 1. Mount liveness check in container

Add a periodic health check that verifies `/workspace` is accessible:

```bash
# Simple check: stat a known file
stat /workspace/.git/HEAD >/dev/null 2>&1
```

This could run:
- As part of the existing liveness monitoring (`liveness:` config section)
- Via a lightweight background loop in the entrypoint
- In the PostToolUse hook (piggyback on existing hook infrastructure)

### 2. Specific exit code for mount failures

When mount drop is detected, exit with a **distinct exit code** (e.g., `exit 4`) so callers (bot, kapsis-status, etc.) can distinguish mount failures from normal agent failures:

| Exit code | Meaning |
|-----------|---------|
| 0 | Success |
| 1 | Agent failure |
| 2 | Push failed |
| 3 | Uncommitted changes |
| **4** | **Workspace mount failure (new)** |

### 3. Recovery guidance in kapsis-status

When exit code 4 is detected, `kapsis-status` should display:
```
MOUNT FAILURE: /workspace disconnected during execution.
Fix: podman machine stop && podman machine start
Then re-launch the agent.
```

## Environment

- macOS Darwin 25.3.0 (arm64)
- Podman 5.8.1 (applehv)
- Kapsis 2.18.0
- Container image: kapsis-slack-bot:latest

PR #280, opened 2026-04-24
fix(mount): prevent and detect /kapsis-status virtio-fs failures (#276)

### LINKED ISSUE #276: [overlay mode] `/kapsis-status/` bind mount intermittently inaccessible at container start — virtio-fs degradation, exits 1 with no output
### Summary

In overlay mode (launched without `--branch`), the `/kapsis-status/` directory is intermittently not accessible inside the container. When this happens, bash cannot open the redirect files specified by the agent command, the agent process never runs, and the container exits immediately with code 1. No response file or debug log appears on the host.

### Reproducing evidence

**Agent IDs of confirmed failures (April 22, 2026):**
- `632091` — overlay mode, exit code 1, container runtime ~31s, no `debug-632091.log` created
- `4d1ef1` — overlay mode, exit code 1, container runtime ~330s, no `debug-4d1ef1.log` created
- `a230ab` — overlay mode, exit code 1, container runtime ~44s, no `debug-a230ab.log` created (April 21)

**Comparison: successful overlay runs (same host, same config):**
- `e0f973`, `65b546`, `03efd4` — overlay mode, exit code 0, `debug-*.log` files present as 0-byte files at `~/.kapsis/status/`

The presence/absence of the debug log is the discriminating signal:
- When mount works: `~/.kapsis/status/debug-AGENT_ID.log` is created (even if empty — bash opened the fd successfully)
- When mount fails: no file is created at all — bash could not open `/kapsis-status/debug-AGENT_ID.log`

**Kapsis status JSON for failed agents:**
```json
{
  "exit_code": 1,
  "error": "Agent exited with error code 1",
  "error_type": "agent_failure",
  "heartbeat_at": null,
  "gist": null
}
```
`heartbeat_at: null` confirms the agent process never started — the container exited before Claude could run.

### Root cause (investigated against Kapsis source)

`/kapsis-status/` is a **bind mount** (not a named volume) — `add_common_volume_mounts()` in `scripts/launch-agent.sh` adds `-v ~/.kapsis/status:/kapsis-status`. The mount mechanism is **identical** between overlay and worktree modes (both call `add_common_volume_mounts()`).

**Actual cause: macOS virtio-fs transport degradation.** The `/kapsis-status/` bind mount goes through the same virtio-fs transport as `/workspace`. When the Podman VM's virtio-fs connection degrades (typically after host sleep/wake), both `/workspace` and `/kapsis-status/` become inaccessible inside the container.

This appears to happen **at container start** — the bash shell opens the output redirect files (`> /kapsis-status/response-ID.md`) before the agent runs, and if virtio-fs is already degraded at that moment, the file open fails silently and the shell exits 1. No `KAPSIS_EXIT_MOUNT_FAILURE` (exit code 4) is returned because `KAPSIS_MOUNT_CHECK_ENABLED` defaults to `false` in `scripts/lib/liveness-monitor.sh:68`.

**Related to #248/#250** ("Detect /workspace virtio-fs mount drop mid-run"), but that fix covers drops *during* a run. This is a startup-time failure — virtio-fs is already degraded when the agent command first tries to open the redirect files.

**Evidence confirming this:**
- Failures cluster after host sleep/wake cycles
- A bot restart (which resets Podman VM state) restores normal operation
- Worktree-mode containers have longer startup sequences that may allow virtio-fs to re-establish; overlay containers start faster and hit the degraded window

### Expected behavior

`/kapsis-status/` should be reliably writable at agent command startup. When it is not, the container should exit with `KAPSIS_EXIT_MOUNT_FAILURE` (4) rather than 1, so callers can distinguish mount failures from agent crashes.

### Environment

- Kapsis version: 2.21.3 (homebrew)
- Platform: macOS (M2 Pro, Darwin 25.3.0)
- Container runtime: Podman (VM-based, virtio-fs)
- Sandbox mode: overlay (auto-detected, no branch)
- Frequency: ~1-3 failures per day under normal load

### Suggested fixes

1. **Enable `KAPSIS_MOUNT_CHECK_ENABLED=true` by default** — detect degraded virtio-fs before the agent command runs and return exit code 4 instead of silently exiting 1.
2. **Add a pre-agent mount probe** — before executing the agent command, verify `/kapsis-status/` is writable (attempt a test write/unlink). If the probe fails, exit with `KAPSIS_EXIT_MOUNT_FAILURE` (4).
3. **Extend #248/#250 coverage** to include startup-time `/kapsis-status/` accessibility checks, not just mid-run drops.

aviadshiber:
Closes #276.

## Problem

On macOS, the Podman VM's virtio-fs transport intermittently degrades (typically after host sleep/wake). Because `/kapsis-status/` was a host bind mount over the same transport (`scripts/launch-agent.sh:1232`), bash's output redirects (`> /kapsis-status/response-<id>.md`) failed at container start and the agent exited 1 with no output and no `debug-*.log` — indistinguishable from a real agent crash. ~1–3 failures per day per affected host.

Only overlay-mode containers tended to hit this because they start faster; worktree containers usually outlasted the degraded window. The underlying cause is identical in both modes.

## Approach

Two complementary layers so the agent **succeeds** rather than merely reports the failure better:

1. **Pre-launch host probe + auto-heal** (`scripts/lib/podman-health.sh`, macOS only). Spins up a short-lived busybox container against a real bind mount and verifies it is writable. On failure, auto-restarts the Podman VM when no other `kapsis-*` containers are live; otherwise fails fast with exit 4 and a clear remediation message (never kills another agent's work).
2. **Off-virtio-fs storage for `/kapsis-status/`** (`scripts/launch-agent.sh` + `scripts/lib/status-sync.sh`, macOS only). Backs the mount with a per-agent Podman named volume (`kapsis-<id>-status`) that lives inside the VM, so agent status writes survive virtio-fs drops. A background worker mirrors the volume into `~/.kapsis/status/` every 2s to preserve live `kapsis-status.sh --watch` compatibility, with a synchronous final drain at exit so post-container logic reads the definitive state.

Plus two smaller, defensive changes:

3. **Container-side readiness probe** (`scripts/entrypoint.sh::probe_mount_readiness`). Runs immediately before `exec`ing the agent command — writes and unlinks a sentinel file under `/workspace` using `timeout` so a hung virtio-fs cannot freeze the container. Emits `KAPSIS_MOUNT_FAILURE:` to stderr and exits 4.
4. **Sentinel detection relaxed** (`scripts/launch-agent.sh:2540`). Honors the sentinel on exit code 1 in addition to 143/137 — covers the original bash-redirect failure mode. Exit 0 is still never overridden (security gate preserved).

## What doesn't change

- **Linux**: `/kapsis-status/` is still a direct bind mount; probe, auto-heal, and sync are `is_macos`-gated no-ops.
- **K8s backend**: uses CR status fields, not bind mounts — unaffected.
- **Mid-run `#248` liveness probe**: untouched.
- **`KAPSIS_MOUNT_CHECK_ENABLED` default**: still `false` — runtime liveness probe semantics are orthogonal to the startup prevention these changes add.

## Files changed

| File | Change |
|------|--------|
| `scripts/lib/podman-health.sh` | **new** — `probe_virtio_fs_health`, `count_running_kapsis_containers`, `maybe_autoheal_podman_vm` |
| `scripts/lib/status-sync.sh` | **new** — `start_status_sync`, `stop_status_sync` (background export + final drain) |
| `scripts/lib/constants.sh` | Exit code constants unchanged; adds VFS probe timeouts, status volume suffix, sync interval |
| `scripts/launch-agent.sh` | Named volume on macOS, pre-launch probe call, sync start/stop, cleanup, sentinel gate covers exit 1 |
| `scripts/entrypoint.sh` | Adds `probe_mount_readiness`, calls it immediately before `exec` |
| `tests/test-podman-health.sh` | **new** — 12 tests (probe, count, auto-heal), all mocked |
| `tests/test-status-sync.sh` | **new** — 6 tests (noop, spawn/pid, final drain, idempotent, double-start) |
| `tests/test-workspace-mount-validation.sh` | +6 tests for `probe_mount_readiness` + sentinel-on-exit-1 |
| `CLAUDE.md` | Expands the "Mount Failure Detection" section |

## Verification

All new + extended suites pass on this branch:

```
tests/test-podman-health.sh                   12/12
tests/test-status-sync.sh                      6/6
tests/test-workspace-mount-validation.sh      22/22
```

Regression spot-checks (unrelated suites) also pass:

```
tests/test-liveness-monitor.sh                56/56
tests/test-input-validation.sh                13/13
```

`shellcheck --severity=warning` is clean across all touched files (project config: `severity=warning` in `.shellcheckrc`).

A live `--dry-run` invocation against a throwaway project completes in ~17s and emits the expected `-v /root/.kapsis/status:/kapsis-status` line on Linux, confirming the macOS branch is correctly gated.

## Test plan

- [x] Unit tests for probe / count / auto-heal decision logic (mocked `podman`)
- [x] Unit tests for status-sync worker start/stop, PID file, final drain
- [x] Unit tests for `probe_mount_readiness` (healthy, missing, read-only, probe-container skip, sentinel format)
- [x] Sentinel detection test covers exit 1 path AND continues to reject exit 0
- [x] Shellcheck clean
- [ ] Live macOS verification on an affected host (sleep/wake + launch overlay agent) — needs reviewer with Mac + affected Podman VM

## Related

Related to #248/#250 (mid-run virtio-fs detection). Those still apply — this PR extends coverage to **startup** and **prevention**, and moves `/kapsis-status/` off the vulnerable transport entirely.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BGxUVEUceMkefK8vWEhRa2)_


aviadshiber (commented):
## Ensemble Review — PR #280

Reviewed by 5 parallel agents (correctness, security, conventions, test-coverage, macOS/virtio-fs). High-level verdict: **solid layered defense with clean platform gating and strong convention adherence**; two findings merit changes before merge, plus a handful of medium/low follow-ups.

### Overview
Three-layer virtio-fs failure handling for macOS Podman VMs (pre-launch probe + auto-heal, in-container startup probe, off-virtio-fs named volume for `/kapsis-status/` with a 2s mirror worker), plus relaxed sentinel detection to cover the original bash-redirect exit-1 failure mode. 1387 additions across 10 files, 12 new + 6 extended tests.

---

### 🔴 Must-fix

**1. Sentinel false-positive on exit 1** — `scripts/launch-agent.sh:~2609`
The sentinel override now triggers on `EXIT_CODE == 1` if `KAPSIS_MOUNT_FAILURE:` appears anywhere in the container output. Exit 1 is the most common, least-specific failure mode. An agent that legitimately fails (missing binary, project permission error, etc.) and happens to emit the sentinel string — even in a log line echoed from its own diagnostics — will be reclassified as exit 4, which masks real failures and flips caller retry semantics (slack-bot treats mount failures as retriable infra errors). The security property "exit 0 is never overridden" is preserved, but the failure→mount-failure misclassification is a DoS/observability hazard.
**Suggested fix:** narrow the pattern (anchor, require format like `^KAPSIS_MOUNT_FAILURE: probe_mount_readiness:`), and/or require the sentinel to be the **last** non-empty line of stderr — only the probe emits it there. Consider adding a per-launch nonce that the entrypoint embeds in the sentinel so the agent cannot spoof it.

**2. Tar extraction without symlink hardening** — `scripts/lib/status-sync.sh:~62`
`podman volume export | tar -C "$host_status_dir" --no-same-owner -xf -` doesn't guard against symlink entries. A buggy or hostile agent inside the container can write `ln -s /some/host/file /kapsis-status/link`; on the host, tar will materialize the symlink under `~/.kapsis/status/` and a subsequent write through it escapes the status dir.
**Suggested fix:** add `--no-same-owner --no-overwrite-dir` and use `tar -xp --no-symlinks` where supported, or extract into a staging dir and `cp -a --no-dereference` into place. BSD tar on macOS needs its own flag form — please gate via `compat.sh`.

---

### 🟠 Should-fix

**3. PID file written after `disown`** — `scripts/lib/status-sync.sh:~134`
Between `disown` and the `printf '%s' "$worker_pid" > "$pid_file"`, the parent can be interrupted and run `stop_status_sync` against a missing/empty pid file, leaking the worker. Write the PID file **before** `disown`.

**4. Probe's own `timeout` binary is unbounded** — `scripts/entrypoint.sh` `probe_mount_readiness`
If `timeout` itself is stuck on a degraded FS, the entrypoint can hang before `exec`. Consider (a) resolving `timeout` via absolute path once at script start, (b) adding an outer `SIGALRM`/`&` + `kill` fallback, or (c) skipping gracefully if `timeout` is missing (as `podman-health.sh` already does — the two paths should match).

**5. `count_running_kapsis_containers` prefix filter is too loose** — `scripts/lib/podman-health.sh:~149`
Matches any `kapsis-*` container. Any user-created container sharing that prefix (test harnesses, stale probe leftovers) will refuse the auto-heal. Use the agent-ID regex (`^kapsis-[a-z0-9-]{8,}-`) or a label filter (`--filter label=kapsis.managed=true`) and tag containers accordingly.

---

### 🟡 Nice-to-have / Follow-ups

- **Status volume cleanup on Linux** (`launch-agent.sh:~2307`): removal is a no-op but issues a `podman volume rm` on a name that never existed; gate the cleanup list with `is_macos` to avoid noise.
- **Final-drain ordering** (`status-sync.sh`): the background loop + TERM-triggered drain is correct to within one interval, but a status write racing the SIGTERM can still be lost if `tar` is mid-stream. Document the "up to one sync interval of staleness" contract or add a post-drain verification read.
- **Probe image**: `busybox:latest` (podman-health.sh:~75) depends on an image cache; on cold VMs the pull latency can itself look like VM degradation. Pin a digest and/or pre-warm on first run.
- **Sentinel logging**: also tag `KAPSIS_MOUNT_FAILURE:` writes with `status_record_error "mount_failure"` so the JSON status is the source of truth and the stderr grep is defence-in-depth rather than the primary signal.

---

### ✅ Strengths

- **Platform gating is comprehensive.** Every macOS-only path (`probe_virtio_fs_health`, `maybe_autoheal_podman_vm`, named volume, status-sync worker) is correctly guarded by `is_macos`/`is_linux`; Linux CI and runtime are unaffected.
- **Exit-0 invariant preserved.** The sentinel override explicitly cannot upgrade a clean success into a mount failure — the most important security property here.
- **Layered detection is well-designed.** Pre-launch probe (host) → `probe_mount_readiness` (container start) → liveness monitor (mid-run) is complementary, not duplicative, and all three share the same sentinel.
- **Auto-heal is conservative.** Refusing to restart the VM while other `kapsis-*` containers run prevents collateral data loss even at the cost of occasional false negatives.
- **Conventions are clean.** `#!/usr/bin/env bash` + `set -euo pipefail`, `[[ ]]`, 4-space indentation, quoted vars, proper `logging.sh` usage with test-friendly fallbacks, library double-source guards, constants lifted into `constants.sh`, Conventional Commits, test-framework assertions throughout. No lint violations detected on inspection.
- **Tests are well-structured.** Mock fidelity on the podman wrapper is good; happy-path + critical negative cases (exit 0 never overridden, sentinel honored on 143/137/1) are all covered; macOS paths forced-on during Linux CI. Suggested gaps to fill: worker respawn on stale PID, final-drain on SIGTERM with partial tar stream, autoheal retry-exhaustion path, `timeout`-missing fallback, volume-export failure during steady state.

### Suggested verdict
Request changes on (1) and (2); address (3)–(5) in this PR if cheap, otherwise as a fast-follow. The architecture is the right answer for #276.

<sub>_Ensemble review: 5 parallel agents, Claude Code on the web._</sub>

---

aviadshiber:
Thanks for the thorough review. Pushed `abb675b` addressing all must-fix and should-fix items.

### 🔴 Must-fix — addressed

**1. Sentinel false-positive on exit 1** → new tagged form, anchored grep, last-N-lines restriction.
- `probe_mount_readiness` now emits `KAPSIS_MOUNT_FAILURE[probe_mount_readiness]: ...`
- `liveness_monitor` now emits `KAPSIS_MOUNT_FAILURE[liveness_monitor]: ...`
- Host-side grep in `launch-agent.sh` is anchored to `^...KAPSIS_MOUNT_FAILURE\[(probe_mount_readiness|liveness_monitor)\]:` and only inspects the **last 10 lines** of container output (the sentinel is always the last thing emitted before exit).
- Two new tests cover the hazards the review flagged: `test_sentinel_rejects_arbitrary_log_line` (agent logs the literal phrase mid-run) and `test_sentinel_rejects_spoofed_early_line` (tagged sentinel far from tail). Skipped the per-launch nonce — the agent already has its own `AGENT_ID`, so a nonce in env is moot against active spoofing; the tightened pattern + position check defeats accidental matches, which was the real observability hazard.

**2. Tar extraction without symlink hardening** → staging-dir + regex-whitelisted `mv`.
`_status_sync_once` now:
- extracts into a per-call `.staging-XXXXXX/` **inside** the host status dir, using `--no-same-owner --no-same-permissions --no-overwrite-dir`,
- uses `find -mindepth 1 -maxdepth 1 -type f` + an explicit `[[ -L "$src" ]] && continue` guard to mirror only regular files,
- whitelists basenames via `^[a-zA-Z0-9._-]+$`,
- `rm -rf`s the staging dir regardless of outcome.

A new test plants a symlink `attacker-link -> /tmp/secret.txt` in the fake volume and asserts the symlink is **not** materialized on the host and the target is untouched.

### 🟠 Should-fix — addressed

**3. PID file after `disown`** → now written atomically via `mktemp + mv` **before** `disown`.

**4. `timeout` binary unbounded** → `probe_mount_readiness` resolves `command -v timeout` once and falls back to direct invocation if missing (matches `podman-health.sh`).

**5. Loose name-prefix match** → containers now carry `--label kapsis.managed=true` and `--label kapsis.agent-id=<id>`. `count_running_kapsis_containers` filters by `label=kapsis.managed=true` via `podman ps --filter`. Test updated to assert the `--filter` flag is present in the `podman ps` argv.

### 🟡 Nice-to-have

- **Status volume cleanup on Linux**: `cleanup_agent_volumes` now conditionally appends `status` to the suffix list only when `is_macos`, silencing the no-op `podman volume rm` on Linux.
- **Final-drain / probe image / sentinel → status.json**: left as follow-ups. Happy to address in a separate PR if you'd like — documented below.

### Deferred (follow-ups worth considering)

- **Final-drain staleness contract**: the "up to one interval" window still exists when the worker is mid-`tar` on SIGTERM. Acceptable IMO because the *volume* is authoritative and `stop_status_sync` performs one more synchronous sync after `kill -TERM`; documenting this in a header comment would tighten the contract without code changes.
- **Probe image pinning**: agree `busybox:latest` is suboptimal. A sha-pinned digest is the right fix but needs a maintenance story. I'd prefer to tackle that alongside the other pinned-digest work rather than do it ad-hoc here.
- **`status_record_error` mirror**: `status_complete 4 "..."` already writes exit_code=4 into `status.json` from the container side (via the liveness path and the new entrypoint probe's best-effort call). The stderr sentinel is already defense-in-depth; happy to add an explicit `error_type: "mount_failure"` field if useful.

Tests: 12 (podman-health) + 8 (status-sync, +2 new) + 24 (workspace-mount, +2 new) = 44 unit tests passing locally, shellcheck clean.

---

aviadshiber:
Pushed `22de775` — deferred nice-to-haves and the suggested test-gap coverage, so everything from the review is now addressed.

### 🟡 Nice-to-haves — now addressed

- **Final-drain staleness contract** — `stop_status_sync` now documents the "up-to-one-interval" window explicitly in its header, and the post-drain sync retries once with 200ms backoff when the first attempt fails (catches SIGKILL-leaving-podman-in-a-weird-state races).
- **Probe-image hardening** — `probe_virtio_fs_health` now:
  - prefers `kapsis-sandbox:latest` (always cached on any host that has launched an agent) over `busybox:latest`;
  - pre-checks with `podman image exists` before `podman run`;
  - when the image isn't cached, skips the probe by default (`KAPSIS_VFS_PROBE_SKIP_IF_MISSING=true`) rather than racing a network pull against a possibly-degraded VM — image-availability and mount-health are now fully decoupled;
  - opt-in: set `KAPSIS_VFS_PROBE_SKIP_IF_MISSING=false` to force a pull with a timeout.
- **Sentinel → `status.json`** — all three mount-failure code paths now call `status_set_error_type "mount_failure"` before `status_complete 4 ...`:
  - host pre-launch auto-heal refusal (new),
  - in-container `probe_mount_readiness` (new),
  - in-container `liveness_monitor` (new).
  The stderr sentinel is now defense-in-depth; JSON status is the authoritative signal for slack-bot and `--watch` consumers.
- **Sentinel grep** — extended to include the `probe_virtio_fs_health` subsystem tag.

### Test-gap coverage (reviewer's suggested gaps)

All added as unit tests, all passing locally:

| Gap | New test |
|---|---|
| Worker respawn on stale PID | `test_start_respawns_when_stale_pid_file_present` |
| Final-drain on SIGTERM with partial tar stream | `test_stop_retries_final_sync_on_transient_failure` |
| Autoheal retry-exhaustion | `test_autoheal_retry_exhaustion_returns_1` |
| `timeout`-missing fallback | `test_probe_without_timeout_binary_still_runs` |
| Volume-export steady-state failure | `test_volume_export_steady_state_failure_does_not_break_worker` |
| (New) image-preflight skip | `test_probe_skipped_when_image_missing_and_skip_enabled` |
| (New) image-preflight opt-in pull | `test_probe_pulls_image_when_skip_disabled` |

**Totals:** 51 unit tests passing (16 podman-health + 11 status-sync + 24 workspace-mount), shellcheck clean at `severity=warning`.

---

PR #284, opened 2026-04-25
fix(mount): complete virtio-fs detection suite with podman-health probe and status-sync (#276)

aviadshiber:
## Summary

Builds on the squash-merged PR #280 with the full implementation of virtio-fs mount failure detection (Issue #276), including the pre-launch probe, named-volume status isolation, and status-sync worker that were deferred or partially landed.

- **`scripts/lib/podman-health.sh`** (new, 318 lines) — pre-launch probe container on macOS: spins up a short-lived container, verifies a bind mount is writable, auto-heals by restarting the Podman VM when no other kapsis containers are running, fails fast with exit 4 + remediation message otherwise.
- **`scripts/lib/status-sync.sh`** (new, 247 lines) — host-side worker that mirrors `kapsis-${AGENT_ID}-status` named volumes into `~/.kapsis/status/` every 5 s so `kapsis-status.sh --watch` consumers work unchanged while `/kapsis-status/` is backed by a VM-resident named volume instead of a virtio-fs bind mount.
- **`scripts/entrypoint.sh`** — `probe_mount_readiness()` writes/unlinks a sentinel file under `/workspace` immediately before exec-ing the agent; emits `KAPSIS_MOUNT_FAILURE:` and exits 4 on degradation detected during container setup.
- **`scripts/launch-agent.sh`** — wires in podman-health pre-launch probe, named-volume status backend (macOS), `caffeinate -i -s` sleep prevention, and `KAPSIS_MOUNT_FAILURE:` sentinel override of exit codes.
- **`tests/test-podman-health.sh`** (new, 602 lines) and **`tests/test-status-sync.sh`** (new, 487 lines) — full unit + integration coverage for the two new libraries.
- **`tests/test-workspace-mount-validation.sh`** — extended with probe_mount_readiness and sentinel-parsing cases.
- **`CLAUDE.md`** / **`agent-sandbox.yaml.template`** — updated docs for exit code 4 sentinel, sleep prevention config, and named-volume storage isolation.

## Test plan

- [ ] `./tests/run-all-tests.sh --quick` passes (no containers needed)
- [ ] `shellcheck scripts/lib/podman-health.sh scripts/lib/status-sync.sh` passes cleanly
- [ ] `./tests/test-podman-health.sh` and `./tests/test-status-sync.sh` pass
- [ ] `./tests/test-workspace-mount-validation.sh` passes with extended probe cases

https://claude.ai/code/session_01Y57qNLhrtcYiSCQuNRJgNC

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y57qNLhrtcYiSCQuNRJgNC)_


aviadshiber:
## Ensemble Review (5 reviewers, majority vote)

Five independent reviewers analyzed this PR across complementary lenses (general correctness, macOS lifecycle, test quality, security, architecture). Findings below are aggregated by **majority vote** — a finding appears as "Consensus" only if **≥3 of 5** reviewers raised the same issue. Solo findings are listed separately under "Specialist observations."

---

### 🟥 Consensus: Action Recommended (≥3 reviewers)

**1. `caffeinate` orphan risk on hard exit** — *4/5 reviewers (R1, R2, R4, R5)*
- **Where:** `scripts/launch-agent.sh:2354-2566`
- **Issue:** `caffeinate -i -s &` is reaped only via the `EXIT` trap. SIGKILL of `launch-agent.sh`, host crash, or shell-level abort leaves `caffeinate` running indefinitely, holding the host awake with no PID file or `kapsis.managed` label to find it later. The `kill` on cleanup also has no SIGKILL fallback and no `wait`.
- **Suggested fix:** write a host-side pidfile (`~/.kapsis/state/caffeinate-${AGENT_ID}.pid`) so a future launch can reap orphans; add `kill -KILL` fallback after a short grace; consider `setsid` so the process group can be signaled atomically.

**2. `/workspace` probe file may leak into the agent's worktree** — *4/5 reviewers (R1, R2, R4, R5)*
- **Where:** `scripts/entrypoint.sh:1621-1633`
- **Issue:** `mktemp` creates `.kapsis-probe-XXXXXX` directly in `/workspace`. The cleanup `rm -f` runs *outside* the inner `timeout`, so if `/workspace` is degraded the rm itself may hang or be skipped. The dotfile can then be staged by an agent's aggressive `git add .` and committed.
- **Suggested fix:** place the probe under `/workspace/.kapsis/probe/` (a directory excluded by Kapsis' default `.gitignore` pattern) and wrap the cleanup `rm` in `timeout` as well.

**3. `_status_sync_worker` silently masks errors and lacks bounds** — *4/5 reviewers (R1, R4, R5, R3-via-tests)*
- **Where:** `scripts/lib/status-sync.sh:81-191`
- **Sub-issues:**
  - Worker subshell does not `set -o pipefail`, so a broken `podman volume export | tar -xf` pipeline (e.g. `podman` segfaults, `tar` reports success on empty input) is undetectable (R1).
  - `podman volume export` runs without a `timeout` wrapper inside the per-tick loop — exactly the hang condition the rest of the PR is designed to detect (R4).
  - Both ends of the pipeline send `2>/dev/null`, so genuine permission/OOM failures are unreportable for triage (R5).
  - Worker has no parent-liveness check; if `launch-agent.sh` is SIGKILL'd between `start_status_sync` and the EXIT trap, the worker becomes a permanent orphan re-exporting the volume (R5).
- **Suggested fix:** wrap `podman volume export` in `_vfs_timeout_cmd 30`, set `pipefail` in the subshell, capture stderr to a per-call temp and log on debug, and add a `kill -0 "$KAPSIS_PARENT_PID" || exit 0` watchdog to the worker loop.

**4. PID-file race between background spawn and PID write** — *4/5 reviewers (R1, R3, R4, R5)*
- **Where:** `scripts/lib/status-sync.sh:174-213`
- **Issue:** The worker subshell starts (`&`) before `printf "$worker_pid" > pid_tmp && mv pid_tmp pid` completes. A concurrent `start_status_sync` for the same `agent_id` can pass the `kill -0` check and spawn a second worker; a concurrent `stop_status_sync` in the gap (~ms) sees no PID file and skips the kill. R1 also notes that `disown "$worker_pid"` after the background spawn races a fast-exiting subshell. R3's test `test_double_start_does_not_spawn_twice` (line 200-231) is itself susceptible to this race and may pass trivially when `pid1=""`.
- **Suggested fix:** capture the PID inside the subshell and write the pid file *from within the worker* before the loop body begins; or guard `start_status_sync` with `flock` on the pid-file directory.

**5. Double `stop_status_sync` invocation pattern is fragile** — *3/5 reviewers (R1, R2, R5)*
- **Where:** `scripts/launch-agent.sh:2596-2610`
- **Issue:** The codebase calls `stop_status_sync` once inline and again from the `EXIT` trap, using `KAPSIS_STATUS_VOLUME=""` as the suppression mechanism. This couples ordering between the trap and the main flow; an early `exit` or non-zero between launch and the explicit stop would orphan the worker because the trap path now reads an empty volume name and short-circuits.
- **Suggested fix:** make `stop_status_sync` truly idempotent on its own (using a `_KAPSIS_STATUS_SYNC_STOPPED` guard variable) and remove the env-var clearing as a control-flow signal.

---

### 🟩 Consensus: Praise (≥3 reviewers)

**Symlink-hardened tar extraction** — *4/5 (R1, R3, R4, R5)*
`scripts/lib/status-sync.sh:58-119` — staging-dir extraction with `[[ -L ]]` rejection and a tight basename allowlist closes the path-escape vector that raw `tar -x` would otherwise expose. R3 highlights the matching regression test (`test_symlink_in_volume_is_not_materialized_on_host`) as strong end-to-end coverage.

**Label-based container identification for auto-heal gating** — *3/5 (R1, R4, R5)*
`scripts/launch-agent.sh:2153-2160` and `scripts/lib/podman-health.sh:259-271` — switching `count_running_kapsis_containers` from name-prefix to `label=kapsis.managed=true` correctly prevents user containers named `kapsis-*` from spoofing/blocking auto-heal decisions. R1 also commends the explicit refusal to restart the VM when other agents are running.

---

### 🟨 Sub-Majority Findings (2 reviewers — not consensus, but worth a quick fix)

- **CLAUDE.md ↔ code mismatch on exit-1 sentinel override** *(R1, R2, both MAJOR)* — `CLAUDE.md:387` and the code comment at `scripts/launch-agent.sh:2649-2655` claim the host overrides exit code **1** with the sentinel, but the actual conditions at `launch-agent.sh:2661/2672` only honor **143** and **137**. Either fix the code or correct the docs.
- **Status-sync interval default mismatch (2s vs 5s)** *(R2, R5, both MAJOR)* — `scripts/lib/status-sync.sh:131` falls back to `:-2` when the constant isn't sourced, but `scripts/lib/constants.sh:349` and `CLAUDE.md` both say 5s. Align the inline fallback with `KAPSIS_DEFAULT_STATUS_SYNC_INTERVAL=5`.
- **`sleep 0.3` portability** *(R1, R2)* — fractional `sleep` works on macOS/GNU but not POSIX/BusyBox. Acceptable for host-only code, but worth a comment.
- **Fixed `sleep 2` after `podman machine start`** *(R1 NIT, R2 MINOR)* — consider polling `podman info` instead of a hardcoded wait on slow Macs.

---

### 🔵 Notable Specialist Observations (1 reviewer, but high signal)

**Reviewer disagreement to surface:**
- The **tagged-sentinel exit-code override** drew opposite calls: R1 PRAISE (`launch-agent.sh:2645-2671` — anchored grep + tail-10 + signal-only override is well-defended); R4 MAJOR (`launch-agent.sh:2641-2664` — a compromised agent OOM-killed with `137` could still emit `KAPSIS_MOUNT_FAILURE[liveness_monitor]:` in its last-10 stderr lines and spoof exit-4). The team should weigh whether the threat model includes a malicious agent inside the sandbox.

**Test quality (R3, test specialist):**
- `tests/test-status-sync.sh:104-144` and `:408-440` and `:200-231` — three tests have timing-flaky assertions that can pass trivially when the worker hasn't yet written its PID file. Loop polls only ~3s for the first sync.
- `tests/test-status-sync.sh:283-318` — the whitelist regex at `status-sync.sh:103-106` rejects leading-dot filenames, which would silently drop `.progress-monitor-state` (mentioned in the in-code comment as legitimate). No test catches this.
- `tests/test-status-sync.sh:322-353` — uses sentinel PID `999999`, which is not guaranteed unused.

**Security (R4, security specialist):**
- `scripts/lib/podman-health.sh:158` — probe runs without `--userns=keep-id` and the production security profile, broadening trust on a tampered cached image.
- `scripts/lib/status-sync.sh:171-180` — worker subshell inherits the full parent environment (potentially `ANTHROPIC_API_KEY`, etc.) into a long-lived background process visible via `/proc/<pid>/environ`.
- `scripts/lib/podman-health.sh:286-289` — auto-heal `podman machine stop` halts *every* container in the VM, not just kapsis ones. A user's unrelated long-running container is killed without warning even when the kapsis count is zero.

**Architecture (R5):**
- `scripts/lib/podman-health.sh:42-50` duplicates timeout-cmd detection that `compat.sh:269-275` already caches in `_KAPSIS_TIMEOUT_CMD` — should reuse the existing helper.
- `kapsis.managed=true` label appears as a string literal in two locations (`launch-agent.sh:2160`, `podman-health.sh:197`) — promote to a constant in `lib/constants.sh`.

---

### Vote Tally Summary

| Severity | Distinct findings | Consensus (≥3) | Sub-majority (2) | Solo (1) |
|---|---|---|---|---|
| BLOCKER | 0 | 0 | 0 | 0 |
| MAJOR | ~14 | 4 | 4 | 6 |
| MINOR | ~22 | 1 | 2 | ~19 |
| PRAISE | ~9 | 2 | 1 | ~6 |

**Recommendation:** address the 5 consensus issues plus the 2 sub-majority MAJOR doc/code mismatches before merge; treat the specialist observations as follow-up issues to triage individually.

---
*Generated by ensemble review: 5 independent reviewers, findings aggregated by majority vote (≥3 reviewer agreement required for consensus).*

---

PR #293, opened 2026-04-27
fix(worktree): surface git worktree add error in logs when dir not created (#282)

### LINKED ISSUE #282: bug: git worktree add failure is silently swallowed, no error surfaced to user
## Problem

When `git worktree add` fails during agent launch, the error is silently discarded and the user only sees a generic message with no actionable information.

### Root cause

In `worktree-manager.sh` around the branch-creation path:

```bash
run_git git worktree add "$worktree_path" -b "$branch" "$base_branch" || true
```

The `|| true` swallows the failure. `run_git` captures git's stderr but only logs it at `DEBUG` level — invisible in normal logs. The script then detects the directory was not created and prints:

```
Worktree directory was not created: /path/to/worktree
This can happen when:
  1. The branch is already checked out in the main repository
  2. There's a git lock file preventing operations
  3. Insufficient disk space or permissions
```

…but never shows *which* of these actually caused the failure.

### Expected behavior

The actual `git worktree add` stderr output should be logged at `ERROR` level when the worktree directory is not created, so the user can diagnose the root cause immediately without needing `--debug` or manual re-running.

### Suggested fix

After detecting that the worktree directory was not created, re-run (or re-surface) the captured git error output at ERROR level:

```bash
# capture stderr explicitly
git_err=$(run_git git worktree add ... 2>&1) || true
if [[ ! -d "$worktree_path" ]]; then
    log_error "git worktree add failed: $git_err"
    ...
fi
```

### Impact

Without this, failures caused by low disk space, network issues fetching the base branch, or permission errors are completely opaque in the logs.

aviadshiber:
## Problem

When `git worktree add` fails during agent launch, the error is silently discarded. `run_git` captures stderr into a local variable and logs it only at `DEBUG` level — invisible in normal operation. The verification block then prints a generic 3-bullet list with no actual root cause.

Closes #282

## Fix

`run_git` now sets a module-level `_KAPSIS_LAST_GIT_ERR` variable whenever a git command fails. The "worktree directory was not created" error block reads this variable and logs it at `ERROR` level.

**Before:**
```
ERROR: Worktree directory was not created: /path/to/worktree
ERROR:
ERROR: This can happen when:
ERROR:   1. The branch is already checked out in the main repository
ERROR:   2. There's a git lock file preventing operations
ERROR:   3. Insufficient disk space or permissions
```

**After:**
```
ERROR: Worktree directory was not created: /path/to/worktree
ERROR: git worktree add error: fatal: '/dev/sda1': no space left on device
ERROR:
ERROR: This can happen when:
...
```

## Files changed

- `scripts/worktree-manager.sh` — 6 lines

## Test plan

- [ ] `bash -n scripts/worktree-manager.sh` passes
- [ ] Trigger a worktree creation failure (e.g. full disk, branch already checked out) and verify the actual git error appears in logs at ERROR level

https://claude.ai/code/session_01BGr5iZkgBCFL1ht3ougQ8g

---
_Generated by [Claude Code](https://claude.ai/code/session_01BGr5iZkgBCFL1ht3ougQ8g)_

### DIFF
--- a/scripts/worktree-manager.sh
+++ b/scripts/worktree-manager.sh
@@ -36,6 +36,9 @@ source "$WORKTREE_SCRIPT_DIR/lib/constants.sh"
 
 # Note: logging functions are provided by lib/logging.sh
 
+# Last git error output — reset before each run_git call, read by create_worktree
+_KAPSIS_LAST_GIT_ERR=""
+
 #===============================================================================
 # GIT COMMAND HELPER
 #
@@ -48,6 +51,7 @@ source "$WORKTREE_SCRIPT_DIR/lib/constants.sh"
 run_git() {
     local git_output
     local git_exit_code
+    _KAPSIS_LAST_GIT_ERR=""
 
     # Capture both stdout and stderr
     git_output=$("$@" 2>&1) && git_exit_code=0 || git_exit_code=$?
--- a/scripts/worktree-manager.sh
+++ b/scripts/worktree-manager.sh
@@ -58,9 +58,10 @@ run_git() {
         log_debug "git output: $git_output"
     fi
 
-    # Log failure if non-zero exit
+    # Log failure if non-zero exit; expose for callers via module-level var (Fix #282)
     if [[ $git_exit_code -ne 0 ]]; then
         log_debug "git command failed with exit code: $git_exit_code"
+        _KAPSIS_LAST_GIT_ERR="$git_output"
     fi
 
     return $git_exit_code
@@ -356,6 +357,10 @@ create_worktree() {
     # Verify worktree was actually created
     if [[ ! -d "$worktree_path" ]]; then
         log_error "Worktree directory was not created: $worktree_path"
+        # Surface the actual git error instead of leaving the user to guess (Fix #282)
+        if [[ -n "${_KAPSIS_LAST_GIT_ERR:-}" ]]; then
+            log_error "git worktree add error: $_KAPSIS_LAST_GIT_ERR"
+        fi
         log_error ""
 
         # Check if another worktree has this branch (Fix #1: enhanced diagnostics)


aviadshiber (commented):

**Verdict: APPROVE_WITH_NITS — unanimous (3/3)**

The fix is small, focused, and a clear net improvement over the silent-failure status quo. All three reviewers (correctness, conventions, security) converge on the same primary concern, with a couple of secondary items worth folding in.

---

### Consensus issue (3/3 reviewers)

**Stale `_KAPSIS_LAST_GIT_ERR` between calls** — `scripts/worktree-manager.sh:64`, surfaced at `:361`

The variable is set only on failure and never reset. Failure modes:

- Several git invocations in `create_worktree` bypass `run_git` entirely (e.g. `git fetch` ~`:297`, `git ls-remote` ~`:315`, `git rev-parse` ~`:330`). If one of those is what actually leaves `worktree_path` missing, the error surfaced will be from an unrelated earlier `run_git` call.
- `|| true` chains around `git worktree add` (~`:319`, `:325`, `:340`, `:351`) can cause an earlier failure's stderr to be reported even when a later attempt succeeded but the directory still isn't there.
- A retry / second invocation in the same shell could surface a stale error from a previous attempt.

**Suggested fix:** add a module-scope `_KAPSIS_LAST_GIT_ERR=""` declaration and clear it at the top of `create_worktree` (and ideally at the top of `run_git` itself). Optionally also stash the originating command name so the log line says *which* git call failed.

```bash
# top of file
_KAPSIS_LAST_GIT_ERR=""

# top of run_git
_KAPSIS_LAST_GIT_ERR=""
```

---

### Secondary items (1–2 reviewers each)

- **Convention nit (Reviewer 2):** the file otherwise prefers locals + `echo`/return-value passing (see `find_worktree_for_branch`, `get_worktree_metadata`). Module globals are reserved for `_LOADED` guards and long-lived state. Acceptable here, but capturing the error into a local right after each `run_git git worktree add` call would be tighter.
- **Missing test (Reviewer 2):** no regression test asserts the underlying git stderr appears in the failure output. Worth a short addition under `tests/test-worktree-*.sh` to lock the fix in.
- **Log-injection / terminal escapes (Reviewer 3):** `git_output` can contain attacker-influenceable content (branch names, remote ref messages). It now flows through `log_error` → `echo -e`, which interprets backslash escapes. Consider stripping control chars before logging, e.g. `tr -d '\000-\010\013\014\016-\037'`.
- **Credential exposure surface change (Reviewer 3):** error output moves from DEBUG to ERROR, which audit/webhook consumers may pick up. `sanitize_secrets` in `lib/logging.sh` covers `Bearer`/`Authorization` headers but not URL-embedded `https://user:token@host` patterns that can appear in git error output. Worth extending the sanitizer as a follow-up.

---

### Bottom line

Ship-able as-is for the immediate diagnostic win, but please add the variable reset (the unanimous concern) before merge — it's a 2-line change that prevents misleading error attribution. Control-char stripping and URL-credential scrubbing are good follow-ups but not blocking.

---

aviadshiber:
Addressed the consensus blocking item in 544ac46:

- Added `_KAPSIS_LAST_GIT_ERR=""` as a module-scope declaration after the CONFIGURATION section
- Added `_KAPSIS_LAST_GIT_ERR=""` at the top of `run_git` so each call starts clean — prevents stale errors from unrelated earlier calls (including direct `git fetch`/`git ls-remote`/`git rev-parse` calls that bypass `run_git`) from appearing as the failure cause

Deferred as follow-ups (non-blocking per the review):
- Control-char stripping before `log_error` (log injection via branch names)
- URL-embedded credential scrubbing in `sanitize_secrets`
- Regression test asserting git stderr surfaces in failure output

---

PR #295, opened 2026-04-27
fix(podman): recover zombie VM by killing vfkit when machine stop hangs (#273)

### LINKED ISSUE #273: Preflight: handle zombie Podman VM where all podman commands hang
## Problem

Kapsis v2.20.5 preflight SSH recovery (#255) handles **stale** SSH tunnels — where `podman run` hangs but other podman commands return errors. 

However, there's a more severe failure mode where the Podman VM enters a **zombie state**:
- `podman version` hangs indefinitely (never returns)
- `podman machine stop` hangs
- `podman machine start` returns "already running"
- `podman ps` hangs
- `podman exec` hangs
- Only `pkill -9 -f vfkit && podman machine start` recovers

In this state, the preflight check itself hangs (it can't detect the problem because the detection command never returns).

## Observed Scenario

- macOS with applehv backend, Podman Machine
- Agent `3a1a65` launched, container shows "Up 5 hours" 
- All `podman exec` commands to the container hang
- Bot marked agent as `error` (liveness timeout)
- Subsequent agent launches also hang (preflight hangs)
- `podman machine list` still returns (shows VM as "running, true, applehv") but all container operations hang

## Suggested Fix

The preflight check should use a **hard timeout** on its detection command (e.g., `timeout 10 podman version`). If the command is killed by the timeout, declare the VM as zombie and attempt recovery:

1. `timeout 10 podman version` → killed by timeout
2. `timeout 10 podman machine stop` → killed by timeout  
3. `pkill -f vfkit` (kill the hypervisor process)
4. Wait 3s
5. `podman machine start`

This is more aggressive than the current preflight (which assumes podman commands return) but covers the zombie VM case.

## Relationship to Existing Issues

- Extends #255 (SSH preflight recovery) to cover the more severe case
- Related to the virtio-fs mount drop bug (macOS applehv)

aviadshiber:
## Problem

When the Podman VM enters a zombie state (all podman commands hang), the existing SSH tunnel recovery in `_recover_podman_ssh_tunnel` calls `podman machine stop` — which also hangs. After the 30s timeout kills it, the code goes straight to `podman machine start`, which fails because the vfkit hypervisor process is still running. The preflight check itself then hangs on subsequent launches.

Closes #273

## Fix

Track whether `podman machine stop` succeeded. On failure or timeout, kill the vfkit hypervisor with `pkill -9 -f vfkit` before attempting `podman machine start`. This matches the only known manual recovery (`pkill -9 -f vfkit && podman machine start`) and makes it automatic.

Both the `$_KAPSIS_TIMEOUT_CMD` (gtimeout/timeout available) and no-timeout code paths are handled consistently.

**Recovery flow — before:**
```
timeout 30 podman machine stop   # hangs, killed by timeout
podman machine start              # fails: VM still alive
# → preflight returns error, user left stranded
```

**Recovery flow — after:**
```
timeout 30 podman machine stop   # hangs, killed by timeout → _stop_ok=false
pkill -9 -f vfkit                # kill hypervisor
sleep 3
podman machine start              # succeeds: clean start
```

## Files changed

- `scripts/lib/compat.sh` — 16 lines (stop-failure detection + vfkit kill)

## Test plan

- [ ] `bash -n scripts/lib/compat.sh` passes
- [ ] macOS only: simulate zombie VM (`pkill -STOP -f vfkit`), trigger preflight, verify recovery completes without manual intervention

https://claude.ai/code/session_01BGr5iZkgBCFL1ht3ougQ8g

---
_Generated by [Claude Code](https://claude.ai/code/session_01BGr5iZkgBCFL1ht3ougQ8g)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -318,8 +318,11 @@ jobs:
       - name: Generate cache key
         id: cache-key
         run: |
-          # Cache key based on Containerfile, build script, and runner architecture
-          echo "key=kapsis-image-${{ runner.arch }}-${{ hashFiles('Containerfile', 'scripts/build-image.sh', 'scripts/entrypoint.sh', 'scripts/lib/**') }}" >> $GITHUB_OUTPUT
+          # Hash only files actually COPY'd into the container (avoids cache
+          # invalidation when non-container scripts like podman-health.sh change)
+          readarray -t copied < <(grep '^COPY' Containerfile | awk '{print $2}' | grep -v '^--')
+          hash=$(sha256sum Containerfile "${copied[@]}" scripts/build-image.sh 2>/dev/null | sha256sum | cut -d' ' -f1)
+          echo "key=kapsis-image-${{ runner.arch }}-${hash}" >> $GITHUB_OUTPUT
 
       - name: Restore cached container image
         id: cache-restore
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -340,7 +340,21 @@ jobs:
             echo "Cached image loaded successfully"
           else
             echo "No cache found, building fresh image..."
-            ./scripts/build-image.sh --name kapsis-test --tag ci
+            # Retry up to 2 times on transient network failures (base image pull)
+            for attempt in 1 2; do
+              echo "Build attempt ${attempt} of 2..."
+              if ./scripts/build-image.sh --name kapsis-test --tag ci; then
+                echo "Build succeeded on attempt ${attempt}"
+                break
+              fi
+              if [[ ${attempt} -lt 2 ]]; then
+                echo "Build failed on attempt ${attempt}, retrying in 15s..."
+                sleep 15
+              else
+                echo "Build failed after ${attempt} attempts"
+                exit 1
+              fi
+            done
           fi
         timeout-minutes: 30
 
--- a/scripts/hooks/tool-phase-mapping.sh
+++ b/scripts/hooks/tool-phase-mapping.sh
@@ -76,8 +76,12 @@ _load_config() {
         done < <(yq -r ".patterns.${category}[]" "$config_file" 2>/dev/null || true)
     done
 
-    # Load phase ranges
+    # Load phase ranges.
+    # Strip surrounding quotes to handle both mikefarah/yq (no quotes) and
+    # python-yq/jq-backed wrappers which quote the concatenated string output.
     while IFS='=' read -r phase range; do
+        phase="${phase#\"}"; phase="${phase%\"}"
+        range="${range%\"}"
         [[ -n "$phase" && -n "$range" ]] && _PHASE_RANGES["$phase"]="$range"
     done < <(yq '.phase_ranges | to_entries | .[] | .key + "=" + (.value | @json)' "$config_file" 2>/dev/null || true)
 
--- a/scripts/lib/compat.sh
+++ b/scripts/lib/compat.sh
@@ -343,15 +343,28 @@ _recover_podman_ssh_tunnel() {
         return 0
     fi
 
-    # Tunnel is stale — attempt recovery via stop + start
+    # Tunnel is stale — attempt recovery via stop + start.
+    # For zombie VMs where all podman commands hang, kill the vfkit hypervisor
+    # directly when machine stop times out (Issue #273).
+    local _stop_ok=true
     if [[ -n "$_KAPSIS_TIMEOUT_CMD" ]]; then
-        "$_KAPSIS_TIMEOUT_CMD" 30 podman machine stop podman-machine-default &>/dev/null || true
+        "$_KAPSIS_TIMEOUT_CMD" 30 podman machine stop podman-machine-default &>/dev/null || _stop_ok=false
+        if [[ "$_stop_ok" == "false" ]]; then
+            declare -f log_warn &>/dev/null && log_warn "podman machine stop timed out — killing vfkit hypervisor (zombie VM recovery)"
+            pkill -9 -f vfkit &>/dev/null || true
+            sleep 3
+        fi
         if ! "$_KAPSIS_TIMEOUT_CMD" 60 podman machine start podman-machine-default 2>/dev/null; then
             # Log start failure for diagnosis (visible with KAPSIS_DEBUG=1)
             declare -f log_debug &>/dev/null && log_debug "podman machine start failed during SSH recovery"
         fi
     else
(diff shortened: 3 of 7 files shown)


aviadshiber (commented):

**Verdict: COMMENT (3/3 reviewers)** — the fix correctly addresses Issue #273 for the common case, but several gaps surfaced in 2+ reviews are worth addressing before merge.

### Majority findings (≥2 reviewers)

**1. `pkill -9 -f vfkit` is too broad — 3/3 unanimous**

Substring-match kills *every* vfkit process for the user, including:
- Other Podman machines (e.g. `podman-machine-other`) — kapsis hardcodes `podman-machine-default` everywhere, but power users with multiple machines lose unrelated VMs without warning.
- Non-Podman vfkit VMs (user-launched vfkit, krunkit-adjacent tooling).
- A concurrent kapsis launch's freshly-started VM. Two parallel launches both hitting the recovery path can race: launch A kills vfkit and `start`s; launch B's `pkill` lands *after* A's start completes, killing A's fresh VM. `scripts/lib/podman-health.sh:191` already guards its auto-heal with an "other kapsis containers running" check; this path lacks the equivalent.

Suggested tightening: scope the pattern to the actual machine, e.g. `pkill -9 -f 'vfkit.*podman-machine-default'` (vfkit's argv includes the machine name), or pin to the binary path Podman uses.

**2. The non-timeout branch is dead on a true zombie VM — 2/3**

Without `gtimeout`/`timeout`, `podman machine stop podman-machine-default &>/dev/null || _stop_ok=false` will *hang forever* on a zombie VM rather than returning non-zero. `_stop_ok=false` is never reached, the `pkill` block is unreachable, and the entire fix is bypassed on systems without coreutils. Options:
- At minimum, `log_warn` when `_KAPSIS_TIMEOUT_CMD` is empty pointing at `brew install coreutils`.
- Or skip `stop` entirely in the no-timeout branch and go straight to `pkill` + `start`.

**3. Duplicated kill block + sibling has the same bug — 2/3**

Both branches duplicate 5 lines (kill + sleep + log). More importantly, `scripts/lib/podman-health.sh:_podman_machine_restart` (lines ~233–237) does the same `stop || true; start` dance and has the *same* zombie-VM bug left unfixed. A shared helper (e.g. `_kill_vfkit_zombie()` in `compat.sh`) called from both call sites would deduplicate and fix the sibling path.

### Minority findings (1/3 — flagged for consideration, not blockers)

- **Logging asymmetry:** `log_debug` for `podman machine start` failure is too quiet — if start fails after a vfkit kill, the next launch is broken with no surfaced signal. Consider `log_warn` on the post-kill path.
- **Naming nit:** `_stop_ok` leading underscore is inconsistent with other locals in `compat.sh` (`probe_timeout`, `timeout_secs`, `i`). Rename to `stop_ok`.
- **State cleanup:** Stale `vfkit.pid` / gvproxy / socket files under `~/.local/share/containers/podman/machine/` are not handled; if `podman machine start` fails on stale state after the kill, the fix won't cover it.
- **Docs:** `CLAUDE.md` still describes manual recovery (`podman machine stop && podman machine start`); worth a note that SSH-tunnel recovery now auto-handles the vfkit zombie case.
- **Test coverage:** PR test plan is `bash -n` + manual macOS repro. A unit-style test mocking `podman` and `pkill` to verify the kill is invoked when stop fails would catch regressions and live alongside existing `compat.sh` tests.

### Consensus on what is correct

- ShellCheck clean, `[[ ]]`, quoting, 4-space indent, `declare -f` log guard pattern matches surrounding code.
- `|| _stop_ok=false` interacts safely with `set -e` (`||` short-circuits errexit).
- `pkill` availability on macOS is fine (BSD base).
- `sleep 3` is a reasonable settling delay; `pkill` no-match (exit 1) is correctly absorbed by `|| true`.

### Suggested order of fixes

1. Scope the `pkill` pattern to `podman-machine-default` *(unanimous, highest impact)*.
2. Extract a shared `_kill_vfkit_zombie()` helper and apply it in `podman-health.sh:_podman_machine_restart`.
3. Add a `log_warn` in the no-timeout branch about installing coreutils, or skip `stop` and go straight to kill+start there.

---
*Ensemble review: 3 independent reviewers, majority vote. Each reviewer received the same diff and project context but did not see the others' assessments.*

---

aviadshiber:
Addressed all three majority findings in 20323a3:

**1. pkill scope (unanimous)** — pattern changed from `-f vfkit` to `-f "vfkit.*${machine}"` in both call sites, scoped to the specific machine name to avoid killing unrelated VMs or racing a concurrent launch.

**2. Deduplication + sibling fix** — extracted `_kill_vfkit_zombie(machine)` helper into `compat.sh`; both the timeout and no-timeout branches in `_recover_podman_ssh_tunnel` now call it. Added a fallback stub in `podman-health.sh` (same `declare -f` pattern as existing stubs). `_podman_machine_restart` now tracks `stop_ok` and calls `_kill_vfkit_zombie` on stop failure — same zombie recovery as `compat.sh`.

**3. No-timeout branch** — added `log_warn` before `podman machine stop` in the no-timeout path pointing to `brew install coreutils`, so users know why zombie recovery may hang. Kept the `stop || stop_ok=false` → kill pattern rather than skipping stop, so non-zombie VMs still get a clean stop.

Minority findings (not addressed here):
- `log_debug` → `log_warn` on post-kill start failure — keeping as-is; start failure is already surfaced via the probe retry loop
- State cleanup (stale pid/socket files) — separate issue, not in scope
- Docs/test coverage — good follow-ups for separate PRs

---

aviadshiber:
@claude open issues of the followups

aviadshiber:
Opened three follow-up issues from the deferred minority findings:

- #296 — Promote `log_debug` → `log_warn` on `podman machine start` failure after zombie kill
- #297 — Clean up stale pid/socket files left by hard-killed vfkit before `machine start`
- #298 — Add integration test + docs for zombie VM recovery flow

---

aviadshiber:
@claude ci fails, fix it

aviadshiber:
**Root cause:** The container image cache key in CI used `scripts/lib/**`, which includes files like `podman-health.sh` that are never `COPY`'d into the container image. When this PR added `_kill_vfkit_zombie` to `podman-health.sh`, it invalidated the cache unnecessarily — forcing a full image rebuild that failed due to network issues in the GitHub Actions environment.

**Fix (commit `8404d35`):** The cache key now dynamically extracts the list of files actually `COPY`'d into the container from the `Containerfile` itself, then hashes only those files plus `build-image.sh`. The cache will now only invalidate when the image content genuinely changes.

---

Issue #298, opened 2026-04-27
test/docs: zombie VM recovery — add integration test and document manual steps


aviadshiber:
## Context

PR #295 added automated zombie VM recovery (`_kill_vfkit_zombie`) for the macOS/applehv case, but test coverage and documentation were explicitly deferred.

## Missing: Integration test

Add a test in `tests/test-security.sh` or a new `tests/test-zombie-recovery.sh` that:

1. Suspends vfkit with `pkill -STOP -f vfkit` to simulate a frozen VM (all podman commands hang).
2. Invokes `_recover_podman_ssh_tunnel` (or `_podman_machine_restart`) via a test harness.
3. Asserts that:
   - `_kill_vfkit_zombie` was called (detectable via a `pkill` stub/mock).
   - `podman machine start` was called after the kill.
   - The function returned success (exit 0).

Because this requires macOS + Podman VM, gate it with:
```bash
is_macos || skip_test "zombie VM recovery is macOS-only"
command -v podman &>/dev/null || skip_test "podman not installed"
```

## Missing: Documentation

Update `docs/ARCHITECTURE.md` or a new section in `docs/SETUP.md` to document:

- The zombie VM symptom (all podman commands hang, `podman machine list` still returns).
- The automated recovery flow introduced in #295.
- Manual recovery steps when auto-recovery fails:
  ```bash
  pkill -9 -f vfkit
  sleep 3
  podman machine start
  ```
- The `KAPSIS_PODMAN_MACHINE` env var (controls which machine name is used in recovery).

## Files to create / modify

- `tests/test-zombie-recovery.sh` (new) or extend `tests/test-security.sh`
- `docs/ARCHITECTURE.md` — add zombie VM recovery section
- Optionally `docs/NETWORK-ISOLATION.md` — note that recovery temporarily drops network isolation

## Related

- Spawned from PR #295 review (minority finding, deferred as separate issue)
- Parent fix: #273

PR #300, opened 2026-04-28
fix(compat): clean up stale pid/sock files after zombie vfkit kill

### LINKED ISSUE #297: feat(compat): clean up stale pid/socket files after zombie VM kill
## Context

PR #295 kills the vfkit hypervisor (`pkill -9 -f "vfkit.*${machine}"`) to recover from a zombie Podman VM, then waits 3 s and calls `podman machine start`. However, after a hard `pkill -9`, the Podman machine may leave behind stale pid/socket files that prevent a clean `machine start`.

## Problem

On macOS with the applehv backend, Podman stores runtime state (pid files, Unix sockets) under paths like:
- `~/.local/share/containers/podman/machine/applehv/<machine>/`
- `/var/folders/.../podman-machine-default.sock`

After a hard kill these files are not cleaned up. `podman machine start` may refuse to start or silently re-use the stale socket, leading to another zombie state.

## Proposed Fix

In `_kill_vfkit_zombie()` (after the `pkill` and before `sleep 3`), add a targeted cleanup step:

```bash
# Remove stale runtime artefacts left by hard-killed vfkit
local runtime_dir="${XDG_RUNTIME_DIR:-/tmp}/podman"
local machine_sock="${runtime_dir}/${machine}.sock"
local machine_pid="${runtime_dir}/${machine}.pid"
rm -f "$machine_sock" "$machine_pid" 2>/dev/null || true
```

The exact paths vary by backend (applehv vs qemu) and OS version — investigation needed to find the canonical locations across Podman 4.x / 5.x.

## Investigation needed

- [ ] Identify stale file paths for applehv and qemu backends on Podman 4 and 5
- [ ] Check whether `podman machine reset --force` is a safer alternative
- [ ] Verify the cleanup doesn't interfere with parallel Kapsis agent runs

## Related

- Spawned from PR #295 review (minority finding, deferred as separate issue)
- Parent fix: #273

aviadshiber:
## Summary

Fixes #297 — after `pkill -9 vfkit`, Podman leaves behind stale `.pid` and `.sock` files that can cause `podman machine start` to refuse to start or silently reuse the stale socket, potentially looping back into a zombie state.

- Extends `_kill_vfkit_zombie()` in `scripts/lib/compat.sh` to remove `*.pid` and `*.sock` from the machine's state directory for all known backend layouts (applehv for Podman 5.x, qemu for Podman 4.x, and the legacy flat layout) before the existing 3-second settle sleep
- Config and disk files (`.json`, `.raw`, `.qcow2`, `.ign`) are never touched — only runtime-only artefacts
- Cleanup is fully non-fatal (`|| true`) so a missing directory or failed `rm` never breaks recovery
- Expands the minimal fallback stub in `scripts/lib/podman-health.sh` to match, so test contexts that load that file without `compat.sh` get the same behaviour

## Ensemble brainstorm summary

| Perspective | Finding |
|---|---|
| Architect | Scoping cleanup to `${machine}` subdirectory makes it parallel-safe; three backend paths cover all Podman 4.x/5.x layouts |
| Contrarian | `podman machine reset --force` is too nuclear (destroys config); XDG_RUNTIME_DIR socket path is session-unpredictable and best skipped |
| Safety | `rm -f` on globs that match nothing is safe even without `nullglob` — `-f` suppresses ENOENT |

## Test plan

- [x] `bash -n` syntax check on both modified files
- [x] `./tests/run-all-tests.sh --quick` — all tests pass, 0 failures
- [ ] Manual macOS verification: simulate zombie VM with `pkill -STOP -f vfkit`, trigger recovery, assert stale files are removed and `podman machine start` succeeds (tracked in #298)

https://claude.ai/code/session_013vtS1fmyMz2jESzLF4JKgr

---
_Generated by [Claude Code](https://claude.ai/code/session_013vtS1fmyMz2jESzLF4JKgr)_

### DIFF
--- a/scripts/lib/compat.sh
+++ b/scripts/lib/compat.sh
@@ -321,23 +321,26 @@ _kill_vfkit_zombie() {
     declare -f log_warn &>/dev/null && log_warn "Killing vfkit hypervisor for '$machine' (zombie VM recovery)"
     pkill -9 -f "vfkit.*${machine}" &>/dev/null || true
 
-    # Remove stale pid/socket files left by the hard-killed hypervisor so that
+    # Remove stale runtime artefacts left by the hard-killed hypervisor so that
     # the subsequent `podman machine start` does not refuse to start or silently
     # reuse a stale socket (Issue #297).  Covers applehv (Podman 5.x), qemu
-    # (Podman 4.x), and the legacy flat layout — never touches config or disk
-    # images (.json, .raw, .qcow2, .ign).
-    local state_base="${HOME}/.local/share/containers/podman/machine"
+    # (Podman 4.x), and the legacy flat layout.
+    # Removed: *.pid  *.sock  *.lock   (runtime-only, safe to delete)
+    # Kept:    *.json *.raw *.qcow2 *.ign  (config/disk — must not be deleted)
+    local state_base="${XDG_DATA_HOME:-${HOME}/.local/share}/containers/podman/machine"
     local machine_dir
     for machine_dir in \
             "${state_base}/applehv/${machine}" \
             "${state_base}/qemu/${machine}" \
             "${state_base}/${machine}"; do
         if [[ -d "$machine_dir" ]]; then
-            rm -f "${machine_dir}"/*.pid "${machine_dir}"/*.sock 2>/dev/null || true
+            rm -f "${machine_dir}"/*.pid "${machine_dir}"/*.sock "${machine_dir}"/*.lock 2>/dev/null || true
             declare -f log_debug &>/dev/null && log_debug "Cleaned stale runtime files in ${machine_dir}"
         fi
     done
 
+    # 3 s for the kernel to fully reap the vfkit process and release any
+    # file-descriptor locks before podman machine start re-opens them.
     sleep 3
 }
 
--- a/scripts/lib/compat.sh
+++ b/scripts/lib/compat.sh
@@ -320,6 +320,24 @@ _kill_vfkit_zombie() {
     local machine="${1:-podman-machine-default}"
     declare -f log_warn &>/dev/null && log_warn "Killing vfkit hypervisor for '$machine' (zombie VM recovery)"
     pkill -9 -f "vfkit.*${machine}" &>/dev/null || true
+
+    # Remove stale pid/socket files left by the hard-killed hypervisor so that
+    # the subsequent `podman machine start` does not refuse to start or silently
+    # reuse a stale socket (Issue #297).  Covers applehv (Podman 5.x), qemu
+    # (Podman 4.x), and the legacy flat layout — never touches config or disk
+    # images (.json, .raw, .qcow2, .ign).
+    local state_base="${HOME}/.local/share/containers/podman/machine"
+    local machine_dir
+    for machine_dir in \
+            "${state_base}/applehv/${machine}" \
+            "${state_base}/qemu/${machine}" \
+            "${state_base}/${machine}"; do
+        if [[ -d "$machine_dir" ]]; then
+            rm -f "${machine_dir}"/*.pid "${machine_dir}"/*.sock 2>/dev/null || true
+            declare -f log_debug &>/dev/null && log_debug "Cleaned stale runtime files in ${machine_dir}"
+        fi
+    done
+
     sleep 3
 }
 
--- a/scripts/lib/podman-health.sh
+++ b/scripts/lib/podman-health.sh
@@ -27,7 +27,23 @@ declare -f log_debug   &>/dev/null || log_debug()   { [[ "${KAPSIS_DEBUG:-}" == 
 declare -f log_success &>/dev/null || log_success() { echo "[OK] $*"; }
 declare -f is_macos    &>/dev/null || is_macos()    { [[ "$(uname -s)" == "Darwin" ]]; }
 declare -f is_linux    &>/dev/null || is_linux()    { [[ "$(uname -s)" == "Linux" ]]; }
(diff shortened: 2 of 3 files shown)


aviadshiber (commented):
## Ensemble review — majority vote

Three independent reviewers (general / convention-focused / security-focused) reviewed this PR. Findings below are those that **at least 2/3 agreed on**; minority findings are noted separately at the end so you can judge.

**Aggregate verdict: COMMENT** (2× COMMENT, 1× REQUEST_CHANGES)

The fix is small, well-scoped, and the comment-block safety contract is excellent. The findings below are about robustness completeness, not correctness of what's there.

---

### Findings (consensus)

#### 🟠 Major — `*.pid`/`*.sock` may not be exhaustive (2/3)
`scripts/lib/compat.sh:336`, `scripts/lib/podman-health.sh:41`

Modern Podman 5.x with `applehv` writes `.lock` files alongside the pid/socket pairs (and the `.lock` is exactly the kind of stale artefact that causes "machine refused to start"). The current globs catch `vfkit.sock`/`gvproxy.sock`/`*.pid` but skip `.lock`. Two reviewers independently flagged this as the most likely scenario where #297 could re-manifest.

**Suggestion:** extend the glob set to include `*.lock` (the safety comment already correctly excludes `.json`/`.raw`/`.qcow2`/`.ign`, so this stays in the "runtime artefact" allowlist). Worth empirically confirming the file list against an actual degraded Podman 5.x machine state dir.

#### 🟡 Minor — hardcoded `${HOME}` ignores `XDG_DATA_HOME` (3/3)
`scripts/lib/compat.sh:329`, `scripts/lib/podman-health.sh:32`

Podman honors `XDG_DATA_HOME`. Users who redirect it (or invoke under `sudo` without `-H`) will silently no-op because `[[ -d "$machine_dir" ]]` rejects the wrong path — masking misconfiguration rather than fixing the bug.

**Suggestion:**
```bash
local state_base="${XDG_DATA_HOME:-$HOME/.local/share}/containers/podman/machine"
```

#### 🟡 Minor — stub duplication will drift from canonical impl (3/3)
`scripts/lib/podman-health.sh:30-46`

The fallback stub used to be a one-line `pkill+sleep`; this PR turns it into a 15-line near-clone of the real implementation. Other stubs in this file (`log_info`, `is_macos`, etc.) are intentionally trivial. The next Podman layout change will require editing both copies, and they will drift (already do — the stub omits the `log_debug` call).

**Suggestion:** either (a) extract the cleanup body into a small `scripts/lib/vfkit-cleanup.sh` both files source, or (b) have podman-health.sh `source` compat.sh lazily and keep the stub as the trivial `pkill+sleep` fallback.

#### 🟡 Minor — no test coverage for the new cleanup branch (3/3)
`tests/test-compat.sh`, `tests/test-podman-health.sh`

Neither test file references `_kill_vfkit_zombie`. The path-construction and selective-removal logic is fully mockable without a real VM — `CLAUDE.md` emphasizes test coverage and a unit test would lock in the "never delete `.json`/`.raw`/`.qcow2`/`.ign`" guarantee from the comment.

**Suggestion:** add a `--quick` test that:
1. Points `HOME` at a `mktemp -d`
2. Stubs `pkill` and `sleep`
3. Seeds `applehv/podman-machine-default/{foo.pid,bar.sock,disk.raw,config.json}`
4. Calls `_kill_vfkit_zombie`
5. Asserts `.pid`/`.sock` removed and `.raw`/`.json` preserved

#### 🔵 Nit — `sleep 3` placement vs. new rm work (2/3)
`scripts/lib/compat.sh:341`

The 3s budget was originally for the kernel to reap vfkit. The rm loop now runs *before* the sleep, which is fine, but the magic number is inherited without justification.

**Suggestion (optional):** poll for absence of the killed pid (`pgrep -f "vfkit.*${machine}"`) instead of fixed sleep, or add a one-line comment explaining the 3s rationale.

---

### Positives (3/3 agreement)
- The inline comment **explicitly enumerates what is NOT deleted** (`.json`, `.raw`, `.qcow2`, `.ign`) — exactly the safety contract a reviewer wants at the `rm` site.
- Correctly **scopes** to the named machine and guards each candidate dir with `[[ -d ]]`, avoiding cleanup of unrelated machines.
- Uses the `declare -f log_warn &>/dev/null && log_warn …` pattern consistent with the rest of compat.sh, keeping the function safe when sourced standalone.

---

### Open questions
1. Has the file list been **empirically verified** on a real Podman 5.x applehv state dir post-zombie-kill, or inferred from documentation? Worth noting in the PR description if confirmed.
2. Should `gvproxy` be killed alongside `vfkit`? On `applehv` it's a separate process owning the API socket — a leftover gvproxy could keep the socket bound even after vfkit dies.
3. Is `KAPSIS_PODMAN_MACHINE` ever populated from user-controlled config? See minority finding below.

---

### Minority findings (1/3, noted but not blocking)
- **Security reviewer flagged input validation on `$machine`**: regex metacharacters flow into `pkill -f "vfkit.*${machine}"` and path components flow into `${state_base}/.../${machine}`. Likely fine in practice (callers pass a literal default), but a defensive `[[ "$machine" =~ ^[A-Za-z0-9_.-]+$ ]] || return 2` guard would harden the function for a small cost.
- **Security reviewer flagged TOCTOU/symlink-following on `rm -f`**: replacing `rm -f *.pid` with `find "$machine_dir" -maxdepth 1 -type f \( -name '*.pid' -o -name '*.sock' \) -delete` would skip symlinks. Low probability inside `$HOME`, but explicitly defensible.
- **One reviewer flagged uncleaned sockets under `XDG_RUNTIME_DIR`/`~/.cache/containers/`** — explicit scope statement in the comment would close this.

---

_Generated via ensemble review (3 independent reviewers, majority vote)._

---

PR #301, opened 2026-04-28
fix(liveness): close 330s mount-check blind window + add evidence logging

aviadshiber:
## Summary

- **P0 fix**: Close the 330-second blind window where a dropped virtio-fs mount went undetected because the mount check was embedded in the liveness grace period sleep
- **H1 evidence**: Add wall-clock timing to `gc_stale_worktrees_by_age` to reveal whether concurrent GC I/O overlaps with container startup
- **H3 fix**: `probe_virtio_fs_health` now tests the actual `/Users` virtio-fs mount (via `KAPSIS_WORKTREE_PATH`) instead of `$TMPDIR` → `/var/folders`, which is a separate transport in the Podman VM

## Root Cause

Forensic analysis of a mount-failure incident (agent `9e9488`, exit code 4) revealed a precise timing match:

```
started_at → updated_at = 341 seconds
300s (liveness grace) + 30s (first interval) + 11s (2 retries × 5s) = 341s ✓
```

The mount check was integrated into the liveness loop which first sleeps the **full grace period (300s)** before any iteration. The standalone `_mount_check_loop` (30s delay) is explicitly disabled when liveness is enabled (`entrypoint.sh:1795-1796`), so the mount check inherited the 300s grace — defeating its intended 30s detection window.

The loop body order (`sleep → kill-check → mount-check → heartbeat-write`) also explains why `heartbeat_at: null`: the mount check killed the agent before the heartbeat line ever executed.

## Fix (`liveness-monitor.sh`)

Split the grace period when `mount_check_delay < grace`:

```
Before:  sleep(300s) → loop: sleep(30s) → first mount check at 330s ❌
After:   sleep(30s) → early mount check → sleep(270s) → loop at 300s ✅
```

If the early probe fails, the agent is killed at ~30s instead of ~330s — 10× faster retry.

## Evidence Logging

**H1 (GC I/O concurrency):** `gc_stale_worktrees_by_age` now logs elapsed time per worktree removal and total GC duration. Future incidents will show whether heavy GC overlaps with a running container's first minutes.

**H3 (wrong probe mount point):** The pre-launch `probe_virtio_fs_health` now mounts `KAPSIS_WORKTREE_PATH` (the actual `/Users/…` path) instead of a `mktemp -d` directory which resolves to `/var/folders` — a separate virtio-fs transport in the Podman VM. A degraded `/Users` mount could previously pass the pre-launch probe silently.

## Test Plan

- [x] All 57 existing liveness monitor tests pass
- [x] New regression test `test_mount_check_early_probe_fires_before_grace` verifies early probe fires at `mount_check_delay` (1s in test), not `grace+interval` (35s)
- [ ] Manual: rebuild bot image and verify `kapsis-status` shows mount failure detected within 30s on next virtio-fs drop


aviadshiber:
Related: #303 tracks the complementary D-state detection problem for **mid-run** drops (after grace). This PR covers the first 300s; #303 covers the rest of the agent lifetime. Together they close the full detection lifecycle.

PR #306, opened 2026-04-30
feat(liveness): host-side vfkit watchdog for fast mid-run virtio-fs drop detection

### LINKED ISSUE #303: feat(liveness): host-side vfkit watchdog for fast mid-run virtio-fs drop detection
## Problem

When virtiofsd/vfkit dies mid-run, the liveness monitor running **inside** the container takes ~180s to detect the mount loss instead of the intended ~55s (30s interval + 2×5s probe timeout + 2×5s retry delay).

**Root cause:** `timeout 5 stat /workspace` does not actually timeout in 5s when the mount is dead. The `stat(2)` syscall enters the Linux kernel's FUSE layer, which puts the process into **D state (TASK_UNINTERRUPTIBLE)** waiting for a response from virtiofsd. `SIGKILL` cannot interrupt D-state processes. The process stays stuck until the kernel's internal FUSE cleanup runs (~50–60s per attempt).

Observed timeline for agent `5489e3` (confirmed via status.json):
```
started_at:   2026-04-28T15:54:15Z
heartbeat_at: 2026-04-28T16:16:47Z  (last successful heartbeat)
updated_at:   2026-04-28T16:19:55Z  (exit code 4 written)

gap = 188s ≈ sleep(30) + D-state-probe(~50s) + retry-delay(5) + D-state-probe(~50s) + retry-delay(5) + D-state-probe(~50s)
```

During those 188s the agent process itself is also stuck in D state — no work is happening, no recovery is possible, and the user sees nothing until the liveness monitor finally kills the container.

## Related

- **PR #301** (`fix/mount-check-blind-window`, open): closes the 330s blind window during the grace period (drops detected at 30s instead of 330s). That PR addresses drops in the **first 300s**. This issue addresses drops **after** the grace period where D-state makes the in-container probe unreliable.

Together: #301 + this issue cover the full lifecycle — early drops AND mid-run drops.

## What's Missing

A **host-side vfkit/virtiofsd process watchdog** in `launch-agent.sh`. From outside the container, `kill -0 <vfkit-pid>` is instant — if vfkit is gone, all virtio-fs mounts are gone. Detection drops from ~188s to ~5s.

## Proposed Fix

After `podman run` is launched (background `&`), capture the vfkit PID and start a lightweight host-side watcher:

```bash
# After container starts:
_vfkit_pid=$(pgrep -n -f "vfkit.*${KAPSIS_PODMAN_MACHINE:-podman-machine-default}" 2>/dev/null)

if [[ -n "$_vfkit_pid" ]]; then
    (
        while kill -0 "$KAPSIS_CONTAINER_PID" 2>/dev/null; do
            if ! kill -0 "$_vfkit_pid" 2>/dev/null; then
                log_warn "vfkit process (PID $_vfkit_pid) exited — virtio-fs mounts lost"
                status_complete 4 "Workspace mount lost: vfkit exited (virtio-fs drop)" 2>/dev/null || true
                kill -TERM "$KAPSIS_CONTAINER_PID" 2>/dev/null || true
                break
            fi
            sleep 5
        done
    ) &
    _VFKIT_WATCHDOG_PID=$!
fi
```

Cleanup: kill `$_VFKIT_WATCHDOG_PID` in the trap/cleanup path alongside the existing caffeinate cleanup.

## Why the In-Container Probe Can't Be Fixed

The D-state hang is a kernel-level constraint — `SIGKILL` is explicitly not delivered to processes in `TASK_UNINTERRUPTIBLE`. The only way to unblock them is to write to `/sys/fs/fuse/connections/<N>/abort`, which requires root and knowledge of the connection number. A host-side watchdog avoids the problem entirely by detecting the root cause (vfkit gone) rather than the symptom (mount inaccessible).

## Evidence

- Agent `9e9488`: 330s blind window (grace-period bug, fixed in #301)
- Agent `5489e3`: 188s detection delay (D-state issue, this issue)
- Both failed with `exit_code: 4`, `error_type: mount_failure`
- `prevent_sleep: true` was active for `5489e3` — sleep/wake cycle ruled out

## Acceptance Criteria

- [ ] `launch-agent.sh` starts a host-side watcher loop after container launch (macOS only, skip if `_vfkit_pid` not found)
- [ ] On vfkit exit: writes `exit_code: 4` + `error_type: mount_failure` to status.json, sends SIGTERM to container
- [ ] Watchdog PID cleaned up in existing trap
- [ ] Log line: `vfkit watchdog active (vfkit PID: <N>, poll: 5s)`
- [ ] Detection time: ≤10s from vfkit exit to status.json update (vs current ~188s)

aviadshiber:
## Summary

Closes #303. Adds a host-side bash watchdog that detects vfkit exit (which kills all virtio-fs mounts in the Podman VM) within ~10s, vs the current ~188s in-container detection. Complements #301, which closed the grace-period blind window in the in-container probe.

## Why

When `vfkit` dies mid-run, every `stat`/`open` on `/workspace` enters the FUSE layer and the calling process is parked in `TASK_UNINTERRUPTIBLE` (D-state). `SIGKILL` cannot interrupt D-state, so each in-container probe attempt burns ~50s waiting on the kernel's FUSE cleanup. Confirmed against agent `5489e3` in the issue: 188s gap between last heartbeat and the eventual exit-4.

A host-side `kill -0` check is instantaneous and detects the root cause (vfkit gone) instead of the symptom (mount inaccessible).

## What

- **`scripts/launch-agent.sh`** — declares `_VFKIT_WATCHDOG_PID` next to `_CAFFEINATE_PID`, starts a background subshell after caffeinate that polls the vfkit PID every 5s, and on vfkit exit:
  1. Writes `exit_code=4` + `error_type=mount_failure` to `status.json` via `status_set_error_type` + `status_complete`.
  2. Logs `KAPSIS_MOUNT_FAILURE[vfkit_watchdog]: …` (matches the existing tagged sentinel format).
  3. SIGTERMs the local podman client (`pkill -TERM -f "podman run .*--name kapsis-${AGENT_ID}\b"`).
- Post-container override mirrors the existing exit-5 (hung-after-completion) pattern at L2688-2695: when `EXIT_CODE` is 143/137 and `status.json` reports exit_code=4, override to 4 so the canonical mount-failure branch runs unchanged.
- Cleanup added to `_cleanup_with_completion` next to the caffeinate kill, so the watchdog never outlives the container.
- macOS-only (`is_macos` gate); Linux falls through silently (no vfkit).
- Anchored `pkill` pattern (`podman run .*--name kapsis-…`) avoids matching incidental `podman ps`/`inspect` invocations.

## Why status.json (not container_output)

Writing a sentinel to `container_output` from a host process races against `tee`. Status.json uses an atomic temp+rename in `_status_write` and the same idiom is already used for exit-5 detection — so the new path is strictly safer.

## Configuration

- `KAPSIS_VFKIT_WATCHDOG_ENABLED` (default: `true`)
- `KAPSIS_VFKIT_WATCHDOG_INTERVAL` (default: `5` seconds)

Documented in `CLAUDE.md` next to the existing `KAPSIS_PREVENT_SLEEP` paragraph.

## Test plan

- [x] **New unit tests** in `tests/test-vfkit-watchdog.sh` (6 tests, all passing):
  - Watchdog writes `exit_code=4` + `error_type=mount_failure` when watched PID exits
  - `pkill` invoked with anchored pattern containing `kapsis-${AGENT_ID}` and `podman run`
  - Watchdog killed before fire leaves status.json untouched (idempotent cleanup)
  - Post-container override: 143 + status.json{exit_code=4} → EXIT_CODE=4
  - Post-container override: non-signal exit (1) is NOT silently upgraded
  - Post-container override: clean status.json leaves EXIT_CODE=143 untouched
- [x] `shellcheck -x --severity=warning scripts/launch-agent.sh tests/test-vfkit-watchdog.sh tests/run-all-tests.sh` — clean
- [x] `./tests/run-all-tests.sh --quick` — pre-existing failures only (keychain, code-sign, k8s, secret-store) reproduce on parent commit and are unrelated to this change
- [ ] **Manual macOS smoke test** (cannot CI; macOS runner with podman machine not available):
  1. `./scripts/launch-agent.sh ~/some-project --agent claude --task "sleep 600"`
  2. In another terminal: `pkill -9 -f 'vfkit.*podman-machine-default'`
  3. Within ~10s: log line `KAPSIS_MOUNT_FAILURE[vfkit_watchdog]` appears, agent exits 4, `status.json` shows `error_type=mount_failure`, time between vfkit kill and agent exit ≤10s.
- [ ] **Linux smoke**: confirm Linux runs are unchanged (watchdog block silently skipped via `is_macos`).

## Out of scope

- macOS CI runner with podman machine (no public CI offers this)
- Auto-respawn of vfkit / mid-run auto-heal
- Sub-second detection (current 5s satisfies the ≤10s SLA)

https://claude.ai/code/session_01UF69gFWVBtGixKEDjLdg1m

---
_Generated by [Claude Code](https://claude.ai/code/session_01UF69gFWVBtGixKEDjLdg1m)_


aviadshiber (commented):
## Ensemble Review (3-reviewer majority vote)

> **Effective verdict: REQUEST_CHANGES** — 2× REQUEST_CHANGES (high confidence) + 1× COMMENT (medium confidence). Posted as COMMENT only because GitHub disallows REQUEST_CHANGES on self-authored PRs.

Three independent reviewers analyzed this PR. The change is well-motivated and the diagnosis (FUSE D-state defeats in-container probes) is correct. However, the ensemble found several issues that converge across reviewers — the strongest signals are listed first.

---

### Critical findings (consensus across reviewers)

**1. `\b` word boundary in `pkill -f` is not portable to macOS BSD** &nbsp; *(flagged by all 3 reviewers)*
`scripts/launch-agent.sh:2620`
```bash
pkill -TERM -f "podman run .*--name kapsis-${AGENT_ID}\b"
```
This block is gated on `is_macos`, so it runs **only** on the platform where `\b` is unsupported. macOS `pkill` uses BSD ERE (regex(3)), which has no `\b` metacharacter — the regex effectively degrades to `kapsis-${AGENT_ID}` with no boundary anchor. Consequences:
- If a user passes `--agent-id` (validated only against `[a-zA-Z0-9_-]+`), prefix collisions become possible. AGENT_IDs `abc` and `abc12` both pass — and the watchdog firing in the `abc` session would also SIGTERM `abc12`'s `podman run`.
- Auto-generated 6-hex IDs avoid this in practice, but user-supplied IDs do not.
- **Fix:** anchor with `( |$)` or `[^a-zA-Z0-9_-]`, or capture the `podman run` PID at launch and `kill` it directly.

**2. Override-on-143/137 + watchdog/main race can silently lose `exit_code=4`** &nbsp; *(flagged by A, C)*
`scripts/launch-agent.sh:2747-2754` and `_cleanup_with_completion` at `:2369-2378`

Two coupled hazards:
- The override only triggers when `EXIT_CODE` is 143 or 137. If `pkill -TERM` doesn't reach the podman client (likely if finding #1 is true), the bash pipeline may return 1 (or even 0) from the `backend_run` path, and the override is bypassed.
- The watchdog's `status_complete 4` runs in a backgrounded subshell. `_STATUS_COMPLETE_SHOWN` is a parent-shell variable the subshell cannot mutate — so `_cleanup_with_completion` may later call `status_complete "$exit_code"` again with a different value and clobber the watchdog's `exit_code=4`. Last-writer-wins on the atomic rename.
- **Fix:** also accept `EXIT_CODE` 0/1 in the override when `status.json` reports `error_type=mount_failure`; or have the watchdog set a parent-visible filesystem sentinel (e.g. touch a flag file) and let the parent finalize status.json.

**3. Test ↔ production drift risk with no enforcement** &nbsp; *(flagged by A, B)*
`tests/test-vfkit-watchdog.sh:37-51` reproduces the watchdog body verbatim from `scripts/launch-agent.sh:2608-2621`. The header comment says "Production and test bodies must stay in sync", but there is no CI/lint gate. **Fix:** extract the watchdog body into a sourceable function (e.g. `scripts/lib/vfkit-watchdog.sh::run_vfkit_watchdog`) so production and tests share a single source of truth.

---

### Critical findings (single-reviewer, but high-impact)

**4. `KAPSIS_PODMAN_MACHINE` is regex-injected unvalidated** &nbsp; *(C)*
`scripts/launch-agent.sh:2603-2605`
```bash
local _vfkit_machine="${KAPSIS_PODMAN_MACHINE:-podman-machine-default}"
_vfkit_pid=$(pgrep -n -f "vfkit.*${_vfkit_machine}" 2>/dev/null || true)
```
The same env var **is** validated in `scripts/lib/podman-health.sh:257` against `^[a-zA-Z0-9_-]+$`. A malformed value (e.g. `.*`) makes the watchdog latch onto an arbitrary unrelated process, which then never exits, suppressing the watchdog entirely (silent under-fire).
**Fix:** apply the same regex guard before line 2605.

---

### Non-critical findings

- **`scripts/launch-agent.sh:2605`** — `pgrep -n` (newest) is fragile if vfkit was just restarted. Consider `pgrep -o`, or filter by parent process. *(A)*
- **`scripts/launch-agent.sh:2608-2621`** — backgrounded subshell has no `trap`, no `setsid`/`disown`, and no parent-liveness check. If `launch-agent.sh` is SIGKILLed or the terminal hangs up, the watchdog can outlive the parent and SIGTERM a future re-launched agent reusing the same AGENT_ID (resume mode at L2077/2143 does this). Add `trap 'exit 0' HUP TERM` and/or `kill -0 $PPID` in the loop. *(C)*
- **PID-reuse hazard** — `kill -0 "$_vfkit_pid"` can falsely succeed for hours if macOS reuses the PID. Periodically re-validate via `ps -o comm=` against expected `vfkit`. *(A, C)*
- **`scripts/launch-agent.sh:2376`** — cleanup `kill` is not followed by `wait`. If the subshell is mid-`_status_write`, `${status}.tmp.$$` can leak (rename is atomic, but temp files accumulate). *(A)*
- **`scripts/launch-agent.sh:2376`** — default SIGTERM may take up to `_watchdog_interval` (5s) to interrupt `sleep`; consider `kill -KILL` or a `trap 'exit 0' TERM` inside the subshell for prompt teardown. *(C)*
- **`scripts/launch-agent.sh:2619`** — `log_warn ... 2>/dev/null || true`: suppressing stderr from a logging call is unusual; the caffeinate path at L2592 doesn't do this. *(B)*
- **`scripts/launch-agent.sh:2747-2754` vs `:2733-2740`** — the new override block is near-identical to the existing hung-after-completion block. Consider extracting an `_override_exit_for_status_code` helper. *(B)*
- **`tests/test-vfkit-watchdog.sh`** — no `trap _teardown_test_env EXIT` per test; on assertion failure, `sleep 30` background processes and `$TEST_TMP` leak. The PATH override at L130-140 also leaks if the watchdog body errors. *(B)*
- **`tests/test-vfkit-watchdog.sh:86,90,132,233`** — the `sleep 30` + `sleep 1` racing pattern is timing-sensitive on slow CI runners. *(B)*
- **`tests/test-vfkit-watchdog.sh:62-64`** — re-sourcing `status.sh` by unsetting `_KAPSIS_STATUS_LOADED` couples to an internal sentinel name. *(B)*
- **`CLAUDE.md:403`** — "SIGTERMs the local podman client" is slightly imprecise; the `pkill` pattern targets `podman run` specifically. Doc otherwise matches the code (env var names, defaults, ~10s SLA). *(C)*

---

### What works well

- Diagnosis is well-grounded (D-state + 188s gap evidence).
- Status-volume vs `container_output` rationale is solid (atomic rename in `_status_write` is the safer path).
- `local _vfkit_pid` declared separately from assignment — correctly avoids SC2155.
- Linux dead-code path is clean (entire watchdog block is `is_macos`-gated).
- CLAUDE.md update is accurate on env-var names and defaults.

---

### Recommended path forward

Ranked by impact:
1. Fix the `\b` portability bug (quick, high impact).
2. Add `KAPSIS_PODMAN_MACHINE` validation (one line, mirrors existing `podman-health.sh`).
3. Decide on the watchdog/main race contract — either (a) broaden the override to include exit codes 0/1 when status.json says mount_failure, or (b) have the watchdog only set a sentinel and let the parent finalize status.json.
4. Extract the watchdog body to a sourceable helper to eliminate test/prod drift.
5. Add a parent-liveness guard inside the subshell (`trap` + periodic `kill -0 $PPID`).

— *Posted by ensemble review (3 reviewers, majority vote: REQUEST_CHANGES)*

---

aviadshiber:
## Ensemble Review #2 (3-reviewer majority vote)

> **Effective verdict: REQUEST_CHANGES → addressed in 80b424c.**
> Tally: 1× REQUEST_CHANGES (security, medium conf), 1× APPROVE (correctness, medium), 1× COMMENT (API/maintainability, high). The critical security finding has been resolved; non-critical items addressed where high-impact, deferred where cosmetic.

Three independent reviewers re-analyzed the prior fix at `a60442f`. Consensus: prior critical items (BSD `\b`, override race, test/prod drift, machine-name validation) were all correctly fixed. **One new critical finding emerged**.

### Critical (Reviewer B, security)

**Forgeable mount_failure path.** The widened override (`EXIT_CODE -ne 0` instead of 143/137) regressed the prior security property because status.json on macOS is mirrored verbatim from the container's `/kapsis-status` named volume into the host status dir (via `status-sync.sh:97-110`, no content validation). A compromised in-container agent could write `exit_code=4 + error_type=mount_failure` and exit 1, masking `agent_failure` as `mount_failure` — which per the documented orchestrator semantics is treated as a transient infra issue, hiding bad agent behavior and potentially causing retry of a compromised agent.

**Resolution (commit `80b424c`):** The override now requires a **host-only sentinel file** the container cannot forge:
- Watchdog touches `${TMPDIR}/kapsis-${AGENT_ID}.vfkit-fired` *before* `status_complete` (so the override has a strong signal even if `status_complete` fails — disk full, status disabled).
- `$TMPDIR` is host-private on macOS and **not bind-mounted into the container**.
- Override gates on: `EXIT_CODE != 0` **AND** sentinel exists. status.json's `mount_failure` is now defense-in-depth, not the trust anchor.
- Sentinel-only fallback: if sentinel exists but status.json doesn't agree, override still fires.
- Exit 0 still preserved.
- EXIT trap removes the sentinel.

### Non-critical, addressed (Reviewer A)

- `sleep "$interval"` → `sleep "$interval" & wait $!` so SIGTERM interrupts immediately. Previously the cleanup trap could block up to 5s on every normal exit (verified empirically: 9.9s vs 2ms with the fix).
- `basename "$PROJECT_PATH"` now cached in the override block (was invoked twice).

### Non-critical, deferred (with reviewer acknowledgement)

- PID-reuse hardening on `$parent_pid` via `ps -o etime=` snapshot (Reviewers A + B). Bound to the vfkit-died window; acceptable risk.
- Override DRY refactor with the exit-5 hung-after-completion block (Reviewer C, cosmetic).
- Extracting `_simulate_override` to a shared helper to fully eliminate test/prod drift (Reviewers A + C). The current helper mirrors production; full extraction is a follow-up.

### Skipped (not effective)

- Adding `vfkit_watchdog` to the `KAPSIS_MOUNT_FAILURE` sentinel regex allowlist (Reviewer A). The watchdog's `log_warn` writes to the host log, not container output, so the in-container-output regex grep wouldn't catch it.

### What works well (consensus)

- Library extraction eliminates body drift; tests source production directly.
- BSD-portable boundary class (`[^a-zA-Z0-9_-]`) correctly prevents AGENT_ID prefix collisions.
- Defense-in-depth input validation on AGENT_ID, machine, interval.
- Trap + `wait` pairing drains in-flight `_status_write` cleanly.
- `pgrep -o` (oldest) is well-justified over `-n`.
- AGENT_ID flag-injection check (R-B): pkill embeds AGENT_ID inside the `-f` PATTERN string, never as a separate argv — safe even with leading `-`.

### Tests

15 tests, all passing on Linux CI (forced `is_macos`). New cases:
- Watchdog touches sentinel on fire; killed-early watchdog leaves sentinel untouched
- Override does NOT fire when status.json reports `mount_failure` but no sentinel (forgery resistance)
- Override does NOT fire when sentinel arg is empty (watchdog disabled)
- Override fires on sentinel alone when `status_complete` failed

shellcheck clean at project severity.

— *Posted by ensemble review #2 (3 reviewers via Claude Code subagents)*

---

PR #307, opened 2026-04-30
test: fix quick-test failures in non-CI local environments

aviadshiber:
## Summary

- Add three skip helpers to `tests/lib/test-framework.sh` (`skip_if_not_mikefarah_yq`, `skip_if_root`, `skip_if_commit_signing_active`) to handle environment differences between local dev and CI
- Replace all `command -v yq` guards in 5 test files with `skip_if_not_mikefarah_yq` for tests that require mikefarah/yq v4 syntax (`select(kind == "seq")`, `yq eval`)
- Add `python3 -c yaml.safe_load` fallback for YAML validation in k8s config tests when mikefarah/yq is absent
- Skip `test_snapshot_file_fallback_on_failure` when running as root (chmod 000 has no effect for UID 0)
- Fix `test_sync_index_cache_tree_rebuild` by setting `git config commit.gpgsign false` on the local test repo (system-wide signing config broke test commits)

## Root causes fixed

| Test failure | Root cause | Fix |
|---|---|---|
| `test_keychain_array_account_parsing` + 10 others | Python yq uses jq 1.7 which removed `kind`; mikefarah/yq required | `skip_if_not_mikefarah_yq` guard |
| `test_generate_cr_valid_yaml` + 2 others | `yq eval '.'` is mikefarah syntax, fails with Python yq | `python3 yaml.safe_load` fallback |
| `test_snapshot_file_fallback_on_failure` | Running as root: `chmod 000` doesn't prevent reads | `skip_if_root` guard |
| `test_sync_index_cache_tree_rebuild` | Global `commit.gpgsign=true` rejects test commits | `git config commit.gpgsign false` on test repo |

**CI is unaffected**: GitHub Actions installs mikefarah/yq v4.52.5, runs as non-root, and has no commit-signing config — all tests that skip locally continue to run and pass on CI.

## Test plan

- [ ] `./tests/run-all-tests.sh --quick` passes with 0 failures locally
- [ ] All 6 previously-failing test scripts now report `Failed: 0`
- [ ] CI Quick Tests check passes on GitHub Actions

https://claude.ai/code/session_012bvK4yuP7FHxx2kEZJgeTx

---
_Generated by [Claude Code](https://claude.ai/code/session_012bvK4yuP7FHxx2kEZJgeTx)_


aviadshiber (commented):
## Ensemble code review (3 independent reviewers, majority vote)

Three reviewers analyzed the PR independently. Findings below are those that received a majority (2/3 or 3/3) vote. Severities reflect the consensus.

### Verdict: **Request changes** (3/3 reviewers — submitted as comment because GitHub does not allow REQUEST_CHANGES on own PR)

---

### F1 — Spurious `[SKIP]` log noise in `test-k8s-config-translation.sh` (3/3) — major

In the new `if skip_if_not_mikefarah_yq 2>/dev/null; then yq...; else python3...; fi` dispatch (lines 207, 309, 384), `2>/dev/null` does **not** suppress the helper's `log_skip` output. `log_skip` writes to **stdout** (`tests/lib/test-framework.sh:240` — `echo -e "${YELLOW}[SKIP]${NC} $*"`), so on systems lacking mikefarah/yq the runner prints a misleading `[SKIP] mikefarah/yq required ...` line and then runs the python3 fallback to completion. Reviewers see a SKIP that didn't actually skip.

**Fix:** Split into two helpers — a silent `has_mikefarah_yq()` (no `log_skip`) used for feature-detection branching, and `skip_if_not_mikefarah_yq()` (logs + returns) for unconditional skip sites.

### F2 — `skip_if_commit_signing_active` is dead code (3/3) — major

The helper is added at `tests/lib/test-framework.sh:1407` but `grep -rn skip_if_commit_signing_active tests/ scripts/` returns only the definition. The actual gpgsign fix in `tests/test-post-container-git.sh:194` uses an inline `git config commit.gpgsign false` instead. Either remove the helper, or use it (and drop the inline line).

### F3 — Mikefarah/yq detection regex is fragile (3/3) — minor/major

`(mikefarah|version v[0-9]+\.[0-9]+\.[0-9]+)` at `tests/lib/test-framework.sh:1382`. The second alternative is dead-or-broken: it matches any tool that prints `version vX.Y.Z` as a substring, and would fail to recognize older mikefarah builds that don't print "version v" (e.g. mikefarah v3 emits `yq version 3.4.1`). Mikefarah always emits its repo URL.

**Fix:** Drop the second alternative — `grep -q 'mikefarah'` is sufficient. For belt-and-suspenders, probe semantically (e.g. `echo '' | yq 'kind' &>/dev/null`).

### F4 — Incomplete scope: other commit-issuing tests still vulnerable (2/3) — minor/major

Only `test_sync_index_cache_tree_rebuild` got the gpgsign fix. `tests/test-full-workflow.sh` and `tests/test-parallel-agents.sh` still call `git commit` without disabling gpgsign (whereas `tests/test-commit-failure-handling.sh` and `tests/test-ephemeral-artifact-filtering.sh` already follow the disable pattern, confirming the idiom). Either move the disable into a shared setup helper, or apply it after every `git init` consistently.

### F5 — `--global` scope check misses system/XDG configs (2/3) — minor

`skip_if_commit_signing_active` checks only `git config --global commit.gpgsign`. System config (`/etc/gitconfig`), XDG config (`~/.config/git/config`), and `GIT_CONFIG_PARAMETERS` env are ignored.

**Fix:** Use `git config --get --type=bool commit.gpgsign 2>/dev/null` (no `--global`) to read the effective value the test will actually inherit.

### F6 — Inverted naming convention forces double-negative call sites (2/3) — minor

`skip_if_X` returning `1` = "should skip" forces all callers to write `if ! skip_if_X; then return 0; fi`. The double-negative is harder to read than `require_X || return 0`. The PR is consistent with the existing `skip_if_no_container` (line 1079), so this is not new debt — but new helpers could break the pattern.

**Fix (optional):** Rename to `require_mikefarah_yq` / `require_non_root` (returning 0 = available) so call sites become `require_mikefarah_yq || return 0`.

---

### Minority views (rejected by majority)

- **Reviewer A claimed `skip_if_root` logic is inverted.** Traced by both peer reviewers and verified by re-reading the file: the logic is correct. Bash `if ! skip_if_root` enters the body when `skip_if_root` returned non-zero (i.e., as root), causing the `return 0` to take effect — which is the intended skip. Non-root users fall through to the test body. **No bug here.**

### Single-reviewer items worth tracking (not blocking)

- PyYAML fallback in `test-k8s-config-translation.sh` assumes `python3 -c "import yaml"` works. macOS system Python often lacks PyYAML; the fallback would fail the test rather than skip. Consider a probe before using.
- The framework has no `TESTS_SKIPPED` counter; `log_skip` followed by `return 0` is currently counted as a pass. Out of scope for this PR but worth a follow-up.
- Em-dash characters in the new `log_skip` messages — minor consistency nit.

---

### What's correct in the PR

- The `if ! skip_if_not_mikefarah_yq; then return 0; fi` rewires across `test-keychain-retrieval.sh`, `test-secret-store-injection.sh`, and `test-git-credential-helper.sh` are correct and a clear improvement over raw `command -v yq` checks.
- `skip_if_root` usage in `test-snapshot-staging.sh` is logically correct (despite Reviewer A's confusion).
- The `git config commit.gpgsign false` line in `test_sync_index_cache_tree_rebuild` correctly fixes that specific test.
- Adding new guards to previously-unguarded tests (`test_inject_to_invalid_value_warning`, `test_build_config_secret_store_toggle`, `test_build_config_minimal_no_secret_store`) is a real correctness improvement.

### Recommended pre-merge fixes (in priority order)

1. **F1 (blocker):** Add silent `has_mikefarah_yq()` predicate; use it in the k8s-config dispatch sites.
2. **F2 (blocker):** Either delete `skip_if_commit_signing_active` or wire it into commit-issuing tests (replacing the inline `git config` line).
3. **F3:** Tighten the regex to just `grep -q 'mikefarah'`.
4. **F4:** Decide whether the gpgsign fix needs to extend to other test files; if yes, factor into shared setup.
5. **F5/F6:** Address as cleanups in the same PR.

_Reviewed by 3 independent agent reviewers; consensus reached via majority vote._

---

aviadshiber:
@claude resolve conflicts with main

aviadshiber:
## Ensemble review disposition (verified at current head)

The branch has been merged with latest `main` (`a6afe7d`, 0 behind) and the full `--quick` suite is green post-merge. Per-finding status:

- **F1** (spurious `[SKIP]` noise from `2>/dev/null` not suppressing stdout `log_skip`): **fixed** — the three dispatch sites in `tests/test-k8s-config-translation.sh` (lines 207/309/384) now branch on the silent `has_mikefarah_yq()` predicate (`tests/lib/test-framework.sh:1420`); verified zero `[SKIP]` lines are emitted when the python3 fallback runs, 23/23 tests pass.
- **F2** (`skip_if_commit_signing_active` dead code): **fixed** — the helper has been removed entirely; commit-issuing tests use the established inline `git config commit.gpgsign false` pattern instead.
- **F3** (fragile yq detection regex): **superseded** — the merge kept main's hardened detection (brand check + `eval --inplace` capability probe), dropping this PR's simpler duplicate.
- **F4** (other commit-issuing tests missing the gpgsign disable): **fixed** — `test-full-workflow.sh` (both repo setups), `test-parallel-agents.sh`, and `test-post-container-git.sh` all disable `commit.gpgsign`.
- **F5** (`--global`-scoped signing check misses system/XDG config): **moot** — the helper it applied to was removed (see F2).
- **F6** (`require_X` vs `skip_if_X` naming): **deferred** — keeping `skip_if_*` for consistency with the existing `skip_if_no_container` precedent, as the review itself allowed.

The rejected minority claim about `skip_if_root` being inverted was re-verified as rejected — the logic is correct and untouched.

---

Issue #256, opened 2026-04-19
Post-container commit fails silently — agent work lost (exit 126 + commit_changes:447)


aviadshiber:
## Summary

When a container exits with code 126, Kapsis detects 24 file changes in the worktree, passes scope validation and staged-file sanitization, but `git commit` fails silently. The 23 staged files are lost — the worktree is cleaned up with no commit on the branch. The caller (slack-bot) sees `container_crash` with no indication that the agent *did* produce useful work.

This triggered a cascade of 3 retries (7ace3a, 9daec0, db728e), each re-doing ~1h43m of work.

## Timeline (agent 41b963)

```
12:00:24  launch-agent starts, pre-flight passes
12:00:26  worktree created: products-41b963, branch: slack-bot/41b963
12:01:03  container starts
13:42:31  container exits with code 126 (after 6088s / ~1h42m)
13:42:41  post_container_worktree: "Agent made 24 file change(s)"
13:42:42  validate_scope_worktree: PASSED
13:42:46  commit_changes: staged 24 files
13:42:46  validate_staged_files: removed accidental submodule (pr-fix-clone), 23 files remain
13:43:02  sanitize_staged_files: "All staged files are clean"
13:43:03  commit_changes:447: "Commit failed or nothing to commit"  ← silent failure
13:43:03  post_container_git:864: "Commit failed"
           worktree cleaned up, branch slack-bot/41b963 left with 0 commits
```

### 1. Commit failure reason is swallowed

`commit_changes:447` logs "Commit failed or nothing to commit" but does NOT log:
- The `git commit` command that was run
- The stderr/stdout from git
- The actual error (was it a hook failure? permission issue? empty diff after staging?)

Without the git error output, it's impossible to diagnose why 23 staged files resulted in "nothing to commit".

### 2. Exit code 126 treated same as any non-zero exit

Exit code 126 typically means "command found but not executable" (e.g., a permission issue with the Claude CLI binary inside the container). But the agent *did* produce 24 file changes over 1h42m, meaning it ran successfully for most of its lifecycle. The exit code 126 likely came from a final command or cleanup step, not from the main agent work.

Kapsis should distinguish between:
- Container failed early (no file changes) → discard
- Container produced work but exited non-zero → **preserve the work**

### 3. File changes are lost on commit failure

When commit fails, the worktree is cleaned up and the branch is left empty. The 24 file changes are permanently lost. The agent's ~1h42m of work is unrecoverable.

[...]

PR #317, opened 2026-05-09
fix(launch-agent): guard critical unguarded function calls (closes #261)

### LINKED ISSUE #261: Evaluate removing set -e from launch-agent.sh in favor of explicit error handling
## Summary

`launch-agent.sh` is 2800+ lines with `set -euo pipefail`. Every bare function call is a potential silent-failure landmine. The script already has dozens of `|| true` and `|| rc=$?` workarounds scattered throughout. Issue #256 was a direct consequence of a missing `||` guard.

## Context

From the brainstorm on #256, the contrarian noted:

> This is the "Swiss cheese model" of error handling — you hope every layer has its holes covered, but inevitably one doesn't. The BashFAQ literally says: "In practice, set -e has a lot of surprising corner cases... it's better to handle each error explicitly."

The `set -e` paradigm works well for simple scripts but becomes fragile in complex ones where:
- Functions are called from multiple contexts (conditional vs non-conditional)
- Return codes need to be captured for branching logic
- Cleanup code must run regardless of failure
- The script has evolved organically with workarounds

## Scope

This is a significant refactor. The proposal is NOT to do it all at once, but to:

1. **Audit** all bare function calls in `launch-agent.sh` for missing `||` guards
2. **Evaluate** whether `set -e` is providing more protection or more footguns at this script size
3. **If warranted**, incrementally migrate to explicit error handling in critical paths (post-container operations, cleanup, status reporting)

## Risk

Removing `set -e` without adding explicit checks everywhere could introduce bugs where errors are silently ignored. This needs careful, incremental migration with test coverage at each step.

## Origin

Identified during brainstorm review of #256 (contrarian observation, acknowledged as out of scope).

aviadshiber:
## Summary

Closes #261.

`launch-agent.sh` runs under `set -euo pipefail` and is 3,195 lines. Bare function calls without `||` guards are silent-failure landmines — issue #256 (commit failure causing 3× repeated 1h43m of wasted agent work) was a direct consequence of exactly this pattern.

An ensemble of three parallel agents audited the script and identified 13 critical unguarded call sites. This PR applies guards to all of them using the three patterns already established in the codebase.

## Changes

**Fail-fast (`if ! func; then status_complete 1 ...; exit 1; fi`)** — failures here leave the agent in an unrecoverable state:

| Call | Why it must fail-fast |
|------|-----------------------|
| `validate_config_security` | Proceeding with an untrusted config is a security bypass |
| `handle_existing_worktree` | Two concurrent worktrees for the same branch causes data loss |
| `setup_sandbox` | `WORKTREE_PATH` is set inside this function; failure leaves it unbound, crashing every downstream call under `set -u` |
| `generate_volume_mounts` | Container launched without mounts runs against the wrong filesystem |
| `generate_env_vars` | Agent runs blind without its environment |
| `write_secrets_env_file` | Agent runs without credentials |
| `backend_build_spec` | Malformed spec produces opaque container engine errors |

**Warn-and-continue (`\|\| log_warn ...`)** — failures are recoverable; crashing here would hide the primary agent result:

| Call | Why it continues |
|------|-----------------|
| `start_vfkit_watchdog` | Defense-in-depth sentinel only; container still runs correctly |
| `cleanup_agent_volumes` | Post-exit maintenance; don't replace a clean exit code with a cleanup failure |
| `cleanup_worktree` (3 sites) | Stale worktree ≠ current-session data corruption; recovery hint included in warning |

**Propagate via `_pcg_rc`** — failure in a sub-function that must abort via the existing error chain:

| Call | Why rc propagation |
|------|--------------------|
| `repoint_sanitized_git_objects` | Must succeed before any `git status`/`sync_index_from_container`; uses `_pcg_rc` so the main exit-code chain (exit 6 / commit_failure) triggers correctly |

## Test plan

- [ ] Launch agent with a deliberately broken config path → should see `status_complete 1 "Config security validation failed"` instead of opaque crash
- [ ] Quick tests pass: `./tests/run-all-tests.sh --quick`
- [ ] `bash -n scripts/launch-agent.sh` passes (syntax clean — verified locally)
- [ ] No new shellcheck warnings on the modified lines

https://claude.ai/code/session_01XTUuspFLHFdAobTHcPVYK6

---
_Generated by [Claude Code](https://claude.ai/code/session_01XTUuspFLHFdAobTHcPVYK6)_

### DIFF
--- a/scripts/launch-agent.sh
+++ b/scripts/launch-agent.sh
@@ -2069,7 +2069,8 @@ handle_existing_worktree() {
     # Handle --force-clean: remove existing worktree
     if [[ "$FORCE_CLEAN" == "true" ]]; then
         log_warn "Force-clean requested: removing existing worktree..."
-        cleanup_worktree "$PROJECT_PATH" "$existing_agent_id"
+        cleanup_worktree "$PROJECT_PATH" "$existing_agent_id" \
+            || log_warn "Worktree cleanup failed for agent '$existing_agent_id' — manual removal may be needed: git worktree remove $existing_worktree"
         log_info "Existing worktree removed. Continuing with fresh start."
         return 0
     fi
@@ -2112,7 +2113,8 @@ handle_existing_worktree() {
                 ;;
             s)
                 log_warn "Starting fresh: removing existing worktree..."
-                cleanup_worktree "$PROJECT_PATH" "$existing_agent_id"
+                cleanup_worktree "$PROJECT_PATH" "$existing_agent_id" \
+                    || log_warn "Worktree cleanup failed for agent '$existing_agent_id' — manual removal may be needed: git worktree remove $existing_worktree"
                 log_info "Existing worktree removed. Continuing with fresh start."
                 return 0
                 ;;
@@ -2497,7 +2499,11 @@ main() {
 
     log_timer_start "config"
     resolve_config
-    validate_config_security "$CONFIG_FILE"
+    if ! validate_config_security "$CONFIG_FILE"; then
+        status_complete 1 "Config security validation failed"
+        _STATUS_COMPLETE_SHOWN=true
+        exit 1
+    fi
     parse_config
     validate_agent_command "$AGENT_COMMAND"
     log_timer_end "config"
@@ -2513,7 +2519,11 @@ main() {
 
     # Handle existing worktree for branch (Fix #1: resume/force-clean)
     if [[ -n "$BRANCH" ]] && [[ "$SANDBOX_MODE" != "overlay" ]] && [[ "$DRY_RUN" != "true" ]]; then
-        handle_existing_worktree
+        if ! handle_existing_worktree; then
+            status_complete 1 "Failed to resolve existing worktree for branch '${BRANCH}'"
+            _STATUS_COMPLETE_SHOWN=true
+            exit 1
+        fi
     fi
 
     # Run pre-flight validation for worktree mode (skip in dry-run)
@@ -2532,7 +2542,11 @@ main() {
     # Setup sandbox (worktree/overlay) — only for backends that support it
     if backend_supports "worktree" || backend_supports "overlay"; then
         log_timer_start "sandbox_setup"
-        setup_sandbox
+        if ! setup_sandbox; then
+            status_complete 1 "Sandbox setup failed (mode: ${SANDBOX_MODE:-unknown})"
+            _STATUS_COMPLETE_SHOWN=true
+            exit 1
+        fi
         log_timer_end "sandbox_setup"
     else
         log_info "Backend '$BACKEND' handles sandbox setup in-cluster"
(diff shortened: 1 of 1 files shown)


aviadshiber (commented):
## Ensemble Review — Majority Vote: **APPROVE** (2 APPROVE / 1 COMMENT, 0 REQUEST_CHANGES)

> Submitted as COMMENT because the same account opened the PR; the ensemble's tally was APPROVE.

Three independent reviewers analyzed this PR (correctness, conventions, security). No blocking issues from any reviewer.

| Reviewer | Focus | Verdict | Confidence |
|---|---|---|---|
| A | Correctness / control-flow | COMMENT | HIGH |
| B | Project conventions / style | APPROVE | HIGH |
| C | Security / risk | APPROVE | HIGH |

### What this PR does well
- **Style is clean** — uses `log_warn`/`log_error` from `logging.sh` (no raw `echo`), `[[ ]]` conditionals, proper quoting, 4-space indent. Backslash + `||` line-continuation matches existing patterns at lines 2425/2646/2696/2718.
- **Status pattern is consistent** — the fail-fast triplet `status_complete 1 "..." ; _STATUS_COMPLETE_SHOWN=true ; exit 1` matches existing convention at lines 2477/2507-2509/2647-2648/3091/3207.
- **No security regressions** — `write_secrets_env_file` partial-state risk is properly mitigated by `umask 0077` (mode 0600) plus the EXIT trap's `backend_cleanup` (`scripts/backends/podman.sh:117-124`) which `rm -f`s `$SECRETS_ENV_FILE`. `cleanup_agent_volumes` only touches build caches / per-agent status volume — no credential material. Agent IDs are validated `^[a-zA-Z0-9_-]+$` elsewhere, so log-injection risk in the warning strings is negligible.
- **Trap interaction is correct** — `_STATUS_COMPLETE_SHOWN=true` correctly suppresses double-`status_complete` from `_cleanup_with_completion` (line 2431) while still letting the trap clean up DNS pin file, secrets env file, snapshot dir, caffeinate, and vfkit watchdog.

### Non-blocking observations (worth considering for follow-up)

**Most fail-fast guards are belt-and-suspenders, not landmine defusal** *(Reviewer A)*
Several wrapped functions reach failure via direct `exit 1` rather than `return 1`:
- `validate_config_security` exits at lines 656, 682 (only returns 0)
- `generate_env_vars` exits at 1688/1695/1709/1716/1722/1734
- `setup_worktree_sandbox` (called via `setup_sandbox`) exits at 1210
- `build_container_command` (called via `backend_build_spec`) exits at 2160
- `write_secrets_env_file` only ever returns 0

The new `if ! func` blocks therefore catch only residual command-substitution / printf failures on these paths. Net effect: improved error messages on rare paths, no regression — but the PR description slightly oversells "silent-failure landmines" since the dominant failure mode was already a hard exit. Consider either (a) converting the inner `exit 1`s to `return 1` so the new guards do real work and the EXIT trap can run uniformly, or (b) a one-line comment that the guard is forward-compatible.

**`repoint_sanitized_git_objects` failure is misclassified** *(Reviewer A)*
When `_pcg_rc=1; return 1` fires from the new guard at line 3066, control reaches `POST_EXIT_CODE -ne 0` at lines 2970-2991. Because no commit was attempted, `commit_status != "failed"`, so the chain reports it as `push_failure` with exit code 1 — not exit 6 (`commit_failure`) as readers of the PR description might infer. Worktree preservation (line 3148) fires correctly. Consider either calling `status_set_commit_info "failed"` before `return 1`, or adding a dedicated `error_type` for objects-symlink failure so callers can disambiguate.

### Nits *(Reviewers B & C)*
- `scripts/launch-agent.sh:2073, 2117, 3170` — recovery hint `git worktree remove $existing_worktree` interpolates the path unquoted into the displayed command. If the worktree path contains spaces, the user copy-pasting the hint gets a broken command. Consider rendering as `git worktree remove '$existing_worktree'`.
- No regression test added for the new error paths. Defensive nature of the change makes this acceptable, but a single integration test that asserts "broken config → `status_complete` with code 1, no opaque crash" would lock in the behavior.

### Reviewers' raw verdicts
- **A (correctness):** *"Most fail-fast guards wrap functions whose real failure modes use direct exit 1, not return 1. The new blocks catch only the rare residual non-zero-return paths; net effect: improved error messages, no regression — but the PR description oversells the change."*
- **B (conventions):** *"Logging compliant, status reporting matches convention exactly, quoting/indentation/line-continuation all consistent. Conventional Commits well-formed."*
- **C (security):** *"No partial-state credential risk, no log-injection vector, no information disclosure beyond what's already routinely logged. Trap interaction correct — no temp-file or secret leakage."*

LGTM on majority. The two follow-up items above are worth a follow-up PR or a code comment, but should not block merge.

---
*Generated by ensemble code review (3 parallel reviewers, majority vote).*

---

PR #322, opened 2026-05-11
feat(conversations): persist agent JSONL via shared host volume (Issue #265)

### LINKED ISSUE #265: Mount conversation JSONL as shared volume for post-mortem preservation
## Problem

When a Kapsis container is killed, crashes, or vanishes (exit code 137, OOM, podman rm), the conversation JSONL stored inside the container overlay filesystem is permanently lost. This makes post-mortem debugging impossible — you can't determine what the agent was doing when it failed.

`podman cp` as a workaround only works on stopped containers, NOT on vanished ones (the exact scenario where post-mortem is most needed).

## Proposal

Mount a host directory as a shared volume for conversation JSONL:

```
-v ~/.kapsis/conversations/${AGENT_ID}:/home/developer/.claude/conversations
```

Configure Claude Code inside the container to write JSONL to this mounted path (or symlink the projects dir). This way JSONL survives container death — same pattern as the existing `/kapsis-status/` shared volume.

Add a TTL-based cleanup in `kapsis cleanup` (e.g., 7 days) to prevent unbounded growth.

## Context

Found during debugging session where 4 agents were investigated post-kill. None had preserved conversation data. The only forensics available were Kapsis launch logs, inbox JSON, and git worktree state.

## Labels
enhancement, reliability

aviadshiber:
## Summary

Closes #265

When a Kapsis container is killed (OOM, crash, `podman rm -f`), all conversation JSONL written inside the container overlay is permanently lost — making post-mortem debugging impossible. This PR mounts a host directory into the container so conversation history survives container death.

- **Mount** `~/.kapsis/conversations/<agent-id>/` into each container at `/home/developer/.claude/conversations` — the bind mount outlives the container overlay, exactly like `/kapsis-status` and `/kapsis-audit`
- **Cleanup** — adds `clean_conversations()` to `kapsis-cleanup` with a 7-day TTL (configurable via `KAPSIS_DEFAULT_CONVERSATIONS_TTL_DAYS`); runs automatically in the default cleanup pass alongside `clean_audit`
- **Isolation** — per-agent subdirectory (`<agent-id>/`) prevents one agent from reading another's conversations

## Design decisions

| Question | Decision | Rationale |
|---|---|---|
| Bind mount vs named volume? | Bind mount | Conversations are not crash-critical like status writes; sleep prevention already guards virtio-fs; avoids needing a sync worker |
| Per-agent or shared dir? | Per-agent (`~/.kapsis/conversations/<agent-id>/`) | Mirrors the per-agent build-cache volume pattern; easy to filter in cleanup |
| TTL? | 7 days default | Same as worktree max-age; long enough for post-mortem, short enough to prevent unbounded growth |

## Files changed

- `scripts/lib/constants.sh` — `CONTAINER_CONVERSATIONS_PATH`, `KAPSIS_DEFAULT_CONVERSATIONS_TTL_DAYS`
- `scripts/launch-agent.sh` — bind mount wired into `add_common_volume_mounts()`
- `scripts/kapsis-cleanup.sh` — `CONVERSATIONS_DIR`, `clean_conversations()`, wired into default cleanup pass

## Test plan

- [ ] Launch agent, confirm `~/.kapsis/conversations/<agent-id>/` is created on host and populated during the session
- [ ] Kill container mid-session (`podman rm -f <container>`), confirm JSONL files remain on host
- [ ] `kapsis-cleanup --dry-run` shows conversation directories older than 7 days
- [ ] `kapsis-cleanup --all` removes all conversation directories
- [ ] `bash -n` syntax check passes on all three modified files

https://claude.ai/code/session_01D68NhjR2rWvN4XU3CP7HrT

---
_Generated by [Claude Code](https://claude.ai/code/session_01D68NhjR2rWvN4XU3CP7HrT)_

### DIFF
--- a/scripts/kapsis-cleanup.sh
+++ b/scripts/kapsis-cleanup.sh
@@ -49,6 +49,7 @@ LOG_DIR="${KAPSIS_LOG_DIR:-$KAPSIS_DIR/logs}"
 SANDBOX_DIR="${KAPSIS_SANDBOX_DIR:-$HOME/.ai-sandboxes}"
 SANITIZED_GIT_DIR="${KAPSIS_SANITIZED_GIT_DIR:-$KAPSIS_DIR/sanitized-git}"
 AUDIT_DIR="${KAPSIS_AUDIT_DIR:-$KAPSIS_DIR/audit}"
+CONVERSATIONS_DIR="${KAPSIS_CONVERSATIONS_BASE_DIR:-$KAPSIS_DIR/conversations}"
 
 # Options
 DRY_RUN=false
@@ -115,6 +116,7 @@ WHAT GETS CLEANED:
     Status files    Completed status files in ~/.kapsis/status/
     Sanitized git   Temporary git dirs in ~/.kapsis/sanitized-git/
     Audit files     Old audit trail files in ~/.kapsis/audit/ (TTL-based or --all)
+    Conversations   Old agent conversation dirs in ~/.kapsis/conversations/ (7-day TTL or --all)
     Containers      Stopped kapsis-* containers (with --containers)
     Volumes         Build cache volumes (with --volumes)
     Images          Kapsis container images (with --images or --all)
@@ -848,6 +850,81 @@ clean_audit() {
     fi
 }
 
+# Clean per-agent conversation directories (Issue #265)
+# Removes directories older than KAPSIS_DEFAULT_CONVERSATIONS_TTL_DAYS (default 7 days).
+# In --all mode removes everything regardless of age.
+clean_conversations() {
+    section "Conversation Directories"
+
+    if [[ ! -d "$CONVERSATIONS_DIR" ]]; then
+        echo "  No conversations directory found"
+        return
+    fi
+
+    # Refuse to traverse a symlinked top-level dir (see clean_worktrees).
+    if [[ -L "$CONVERSATIONS_DIR" ]]; then
+        echo -e "  ${YELLOW}Conversations dir is a symlink; refusing to clean${NC}"
+        return
+    fi
+
+    local count=0
+    local total_size=0
+    local ttl_days="${KAPSIS_DEFAULT_CONVERSATIONS_TTL_DAYS:-7}"
+    local now
+    now=$(date +%s)
+    local ttl_seconds=$((ttl_days * 86400))
+
+    for conv_dir in "$CONVERSATIONS_DIR"/*/; do
+        [[ -d "$conv_dir" ]] || continue
+        # Security: skip symlinks to prevent following attacks (see clean_sandboxes).
+        [[ -L "$conv_dir" ]] && continue
+        local name
+        name=$(basename "$conv_dir")
+
+        # Apply project/agent filter when set
+        if [[ -n "$PROJECT_FILTER" ]] && [[ "$name" != *"${PROJECT_FILTER}"* ]]; then
+            continue
+        fi
+        if [[ -n "$AGENT_FILTER" ]] && [[ "$name" != *"${AGENT_FILTER}"* ]]; then
+            continue
+        fi
+
+        # In --all mode clean everything; otherwise only expire past TTL
+        if [[ "$CLEAN_ALL" != "true" ]]; then
+            local mtime
+            mtime=$(get_file_mtime "$conv_dir" 2>/dev/null) || continue
+            [[ -z "$mtime" ]] && continue
+            local age=$((now - mtime))
+            if [[ "$age" -le "$ttl_seconds" ]]; then
+                continue
+            fi
+        fi
+
+        local size
+        size=$(get_dir_size "$conv_dir")
+        local size_human
+        size_human=$(format_size "$size")
+
+        if [[ "$DRY_RUN" == "true" ]]; then
+            print_item "conversation" "$name" "$size_human"
+        else
+            rm -rf "$conv_dir"
+            print_item "conversation" "$name" "$size_human"
+        fi
+
+        ((total_size += size)) || true
+        ((count++)) || true
+    done
+
+    if (( count == 0 )); then
+        echo "  No expired conversation directories to clean"
+    else
+        echo -e "  ${BOLD}Total: $count conversation directories ($(format_size $total_size))${NC}"
+        ((TOTAL_SIZE_FREED += total_size)) || true
+        ((ITEMS_CLEANED += count)) || true
+    fi
+}
+
 # Clean SSH cache
 clean_ssh_cache() {
     section "SSH Host Key Cache"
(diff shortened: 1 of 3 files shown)


aviadshiber (commented):
## Ensemble Review — Majority Vote (3 independent reviewers)

**Verdict: REQUEST_CHANGES** (unanimous, 3/3 — posted as COMMENT because GitHub disallows self-REQUEST_CHANGES)

Three reviewers analyzed this PR independently with no cross-talk; findings below are filtered to those that received ≥2 votes, with a few high-confidence single-vote items kept for completeness.

---

### Blocking (consensus)

**1. The mount target is not where Claude Code writes JSONL** *(3/3 votes)*

`scripts/lib/constants.sh:361` sets `CONTAINER_CONVERSATIONS_PATH=/home/developer/.claude/conversations`, but Claude Code persists session JSONL under `~/.claude/projects/<encoded-cwd>/<session-id>.jsonl`, not `~/.claude/conversations/`. A grep across the repo finds no env var, settings entry, agent adapter, or entrypoint hook that redirects the writer. As shipped, the mount captures nothing in production and Issue #265's post-mortem goal is not delivered.

Suggested fix: either change the mount to `/home/developer/.claude/projects`, or add the redirection (env var / `settings.json`) that points Claude Code at the new path.

**2. TTL pruning is dead code** *(3/3 votes)*

In `scripts/kapsis-cleanup.sh:895`:
```bash
mtime=$(get_file_mtime "$conv_dir" 2>/dev/null) || continue
```

`get_file_mtime` (`scripts/lib/compat.sh:65`) starts with `[[ ! -f "$file" ]] && return 1`, so it returns exit 1 + empty string for every directory. The `|| continue` then skips every iteration in non-`--all` mode — the 7-day TTL never expires anything. Use `get_dir_mtime` (`compat.sh:87`) instead.

**3. Project/agent filters cannot match agent-id directory names** *(3/3 votes)*

`scripts/kapsis-cleanup.sh:885–890` filters via substring match against `name=$(basename "$conv_dir")`. But `generate_agent_id` (`launch-agent.sh:66`) emits a bare 6-char hex (e.g., `a3f2b1`) with no project/agent-type prefix. `*"${PROJECT_FILTER}"*` will silently never match — `kapsis-cleanup --project foo` skips conversations entirely. Compare `clean_sandboxes`/`clean_worktrees`, which iterate over `project-agent`-shaped names.

**4. Overlay collision with the staged `~/.claude`** *(1/3 votes, but high-impact if confirmed)*

The Claude agent profile stages `~/.claude` into the container via fuse-overlayfs after the bind mount is set up (`scripts/entrypoint.sh`). The overlay mounts on top of `/home/developer/.claude`, which can hide the bind-mounted `conversations` subdir; writes then land in the overlay upper layer (ephemeral) instead of the host bind mount, defeating post-mortem survival. Please verify mount ordering and add a test that asserts a file written to the path from inside the container shows up on the host.

---

### Non-blocking (consensus)

- **No test coverage** *(3/3)* — Project policy (CLAUDE.md) expects quick tests for new behavior; nothing covers the mount, TTL cleanup, symlink refusal, or filters. PR test plan boxes are unchecked.
- **Env-var naming asymmetry** *(2/3)* — `KAPSIS_CONVERSATIONS_DIR` (per-agent path, in `launch-agent.sh:1275`) vs `KAPSIS_CONVERSATIONS_BASE_DIR` (base, in `kapsis-cleanup.sh:52`). Other persistent dirs (`KAPSIS_AUDIT_DIR`, `KAPSIS_STATUS_DIR`) use one name everywhere. Suggest: `KAPSIS_CONVERSATIONS_DIR` for the base, derive per-agent inside `launch-agent.sh`.
- **No K8s backend parity** *(2/3)* — `operator/internal/controller/job_builder.go` wires `kapsis-status` and `kapsis-audit` volumes but no conversations volume/env. Feature is podman-only despite K8s being a first-class backend per CLAUDE.md.
- **No documentation updates** *(2/3)* — `docs/CLEANUP.md`, `docs/CONFIG-REFERENCE.md`, and `CLAUDE.md`'s persistent-dirs table should describe the new pass, TTL, and env vars.
- **Per-loop symlink guard ineffective** *(1/3, verified empirically)* — `kapsis-cleanup.sh:880`: `for conv_dir in "$CONVERSATIONS_DIR"/*/` yields trailing-slash entries; `[[ -L "$conv_dir" ]]` (with the trailing slash) follows the link and tests the target, so symlinks-to-dirs pass through. Either drop the trailing slash from the glob or test `${conv_dir%/}`.
- **Mount added unconditionally for non-Claude agents** *(1/3)* — Aider/Codex/Gemini don't use this path; the mount creates orphan empty dirs that `kapsis-cleanup` then has to scan. Consider gating on agent type or a `KAPSIS_CONVERSATIONS_ENABLED` toggle mirroring `KAPSIS_AUDIT_ENABLED`.
- **Active-agent race** *(1/3)* — `clean_conversations` has no liveness check (unlike `clean_status`); `kapsis-cleanup --all` during a live session could `rm -rf` the bind-mount source.

---

### Nits

- `KAPSIS_DEFAULT_CONVERSATIONS_TTL_DAYS` is read as if user-overridable, but `KAPSIS_DEFAULT_*` constants are conventionally `readonly` in `constants.sh`. Introduce a separate `KAPSIS_CONVERSATIONS_TTL_DAYS` override, mirroring `KAPSIS_AUDIT_TTL_DAYS`.
- TTL = 7 days vs `clean_audit`'s 30 days — pick a value with a stated rationale.
- Section comment in `constants.sh` ("Same pattern as /kapsis-status — a bind-mount") is outdated: on macOS `/kapsis-status` was deliberately moved to a named volume + sync worker to survive virtio-fs drops (Issue #276). This PR reintroduces the bind-mount failure mode; either document the trade-off or apply the same volume-+-sync pattern.
- "No expired conversation directories to clean" is misleading under `--all`; phrase like `clean_audit`'s "No audit files to clean".

---

### Positives

- Symlink defenses (top-level dir + per-entry) correctly mirror `clean_worktrees`/`clean_sandboxes`.
- Per-agent isolation in the path (`/<AGENT_ID>/`) is the right shape.
- `clean_conversations` faithfully mirrors `clean_audit`'s structure (section header, TTL handling, dry-run support, totals accounting).
- `((count++)) || true` and `((total_size += size)) || true` correctly handle the `set -e` arithmetic pitfall called out in CLAUDE.md.
- Constants are well-placed in `scripts/lib/constants.sh` with a clear comment block referencing Issue #265.
- `AGENT_ID` is guaranteed to be set by the time `add_common_volume_mounts` runs, and `ensure_dir` is invoked before the mount is appended.

---

*Method: 3 independent reviewers, no cross-talk; findings reported here received ≥2 votes (high-confidence single-vote items kept where empirically verified).*


---

aviadshiber:
@claude fix issues from review, change path to right dir.
Review with security agent these changes as well.

aviadshiber:

aviadshiber:
@claude fix all issues with consensus

PR #327, opened 2026-05-12
feat(plugins): inject Claude Code plugin hooks into settings.local.json

aviadshiber:
## Summary

- Adds `scripts/lib/inject-plugin-hooks.sh` — injects Claude Code plugin hooks (status, stop) into `settings.local.json` inside the container at startup
- Wires hook injection into `scripts/entrypoint.sh` and `scripts/launch-agent.sh`
- Adds comprehensive tests in `tests/test-plugin-hook-injection.sh`
- Adds `docs/PLUGINS.md` reference guide for plugin hook configuration

## Changes

- `scripts/lib/inject-plugin-hooks.sh` — new library for plugin hook injection
- `scripts/entrypoint.sh` — calls hook injection at container startup
- `scripts/launch-agent.sh` — pre-launch plugin hook setup
- `tests/test-plugin-hook-injection.sh` — 701-line test suite
- `docs/PLUGINS.md` — plugin hooks documentation

## Test plan

- [ ] `./tests/run-all-tests.sh --quick` passes
- [ ] `shellcheck scripts/lib/inject-plugin-hooks.sh` passes
- [ ] Plugin hooks are injected correctly inside container at startup

https://claude.ai/code/session_012Gmtu6W93gTxZM99TkxEiP

---
_Generated by [Claude Code](https://claude.ai/code/session_012Gmtu6W93gTxZM99TkxEiP)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -69,6 +69,7 @@ kapsis/
 │   │   ├── progress-monitor.sh # Progress tracking
 │   │   ├── inject-status-hooks.sh # Status hook injection
 │   │   ├── inject-lsp-config.sh   # LSP configuration injection
+│   │   ├── inject-plugin-hooks.sh # Plugin hook injection (Claude Code plugins)
 │   │   ├── build-config.sh     # Build configuration parser
 │   │   ├── atomic-copy.sh      # Atomic file copy operations
 │   │   ├── audit.sh            # Audit trail recording
@@ -144,6 +145,7 @@ kapsis/
 | Security vulnerability scan | `docs/SECURITY-VULNERABILITY-SCAN.md` |
 | Agent profiles design | `docs/designs/agent-profiles-architecture.md` |
 | Agent profiles | `configs/agents/*.yaml` |
+| Plugin hook injection | `docs/PLUGINS.md` |
 | Security policy | `SECURITY.md` |
 | AI agent guidelines | `AGENTS.md` |
 
@@ -240,6 +242,7 @@ Status is written as JSON to `~/.kapsis/status/`.
 | `scripts/lib/podman-health.sh` | Podman VM health probe & auto-heal (macOS pre-launch mount check) |
 | `scripts/lib/status-sync.sh` | Mirrors per-agent named volumes to `~/.kapsis/status/` (macOS) |
 | `scripts/lib/inject-lsp-config.sh` | LSP config injection — editor integration inside containers |
+| `scripts/lib/inject-plugin-hooks.sh` | Plugin hook injection — merges Claude Code plugin hooks into settings.local.json |
 | `scripts/lib/rewrite-plugin-paths.sh` | Plugin path rewriting — adjusts paths for container mounts |
 | `scripts/lib/ssh-config-compat.sh` | SSH config portability — normalizes SSH config across environments |
 | `scripts/backends/podman.sh` | Podman backend — local container execution |
--- a/docs/CONFIG-REFERENCE.md
+++ b/docs/CONFIG-REFERENCE.md
@@ -166,6 +166,20 @@ agent:
   # Default: 60
   gist_llm_interval: 60
 
+  # Inject Claude Code plugin hooks from the host into the container.
+  # When true, Kapsis merges hooks from each host-enabled (and whitelisted) plugin's
+  # hooks/hooks.json into ~/.claude/settings.local.json at container startup.
+  # Only applies to claude-cli / claude / claude-code agent types.
+  # Default: false (opt-in)
+  install_plugins: false
+
+  # Restrict which plugins may be injected (optional allowlist).
+  # Must be a JSON array of exact plugin ids (e.g. ["deeperdive-java-linter@m"]).
+  # Empty array [] or omitting the key allows all host-enabled plugins.
+  # Corrupted (non-array) value is fail-closed: no plugins are injected.
+  # Default: [] (allow all host-enabled plugins)
+  plugin_whitelist: []
+
 #===============================================================================
 # FILESYSTEM MOUNTS
 #===============================================================================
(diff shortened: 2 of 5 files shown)


aviadshiber (commented):

Three reviewers reviewed this PR independently with different lenses (shell correctness/security, project conventions, tests/docs). All three returned **COMMENT** — solid PR, no blockers. Findings below are aggregated by majority vote.

---

### Majority findings (flagged by ≥2 reviewers)

**1. `docs/PLUGINS.md` is invisible from the documented entry points** *(Reviewers B + C)*

The new doc is excellent but not linked from anywhere a reader following the project's documented map would land. Please add:

- `CLAUDE.md` — Documentation Map table (~L98–120): add a row for `docs/PLUGINS.md`
- `CLAUDE.md` — Repository Structure tree under `docs/` (~L71): add the file
- `CLAUDE.md` — Key Files table (~L242): add a row for `scripts/lib/inject-plugin-hooks.sh` (sibling `inject-lsp-config.sh` is listed; this one isn't)
- `docs/CONFIG-REFERENCE.md`: document the new `agent.install_plugins` and `agent.plugin_whitelist` YAML keys (currently only in `docs/PLUGINS.md`)

**2. Test gap: pre-existing malformed `settings.local.json`** *(Reviewers A + C)*

If the file already exists on disk with corrupted JSON (e.g., truncated by a prior crashed run), the inner `jq` merge at `scripts/lib/inject-plugin-hooks.sh:279–316` will hit the `else` branch *per plugin* and leave the file untouched while logging WARN. No test pins this behavior. A ~10-line regression test would protect a realistic failure mode.

---

### High-signal minority findings (one reviewer, but concrete)

**A. Half-failed split-jq leaks tmp file and silently misreports success** — `scripts/lib/inject-plugin-hooks.sh:316–328`

```bash
if jq ... > "$tmp_file" 2>/dev/null; then
    plugin_hook_count=$(jq -r '.plugin_hook_count' "$tmp_file")
    jq -c '.settings' "$tmp_file" > "${tmp_file}.settings" \
        && mv "${tmp_file}.settings" "$settings_local"
    rm -f "$tmp_file"
    ...
    log_info "Loaded plugin ..."
    merged_count=$((merged_count + 1))
fi
```

If the second `jq -c '.settings'` fails: (a) the `&&`-guarded `mv` is skipped, (b) we never enter `else`, (c) `${tmp_file}.settings` is orphaned (only `$tmp_file` is unlinked), and (d) we still log `Loaded plugin …` and increment `merged_count` — falsely claiming success. Additionally, `plugin_hook_count=$(jq …)` on the line above has no `|| true`, so a jq failure there aborts the entire loop under `set -e`. Suggested fix:

```bash
if plugin_hook_count=$(jq -r '.plugin_hook_count' "$tmp_file" 2>/dev/null) \
   && jq -c '.settings' "$tmp_file" > "${tmp_file}.settings" 2>/dev/null \
   && mv "${tmp_file}.settings" "$settings_local"; then
    rm -f "$tmp_file" "${tmp_file}.settings"
    chmod 600 "$settings_local"
    log_info "Loaded ..."
    merged_count=$((merged_count + 1))
else
    rm -f "$tmp_file" "${tmp_file}.settings"
    log_warn "..."
fi
```

**B. Likely typo in `docs/PLUGINS.md:3`** — opening sentence reads as "headless mode **does** load plugins" which contradicts the rest of the doc and the script header (`scripts/lib/inject-plugin-hooks.sh:4–9`). Probably missing a "not."

**C. No symlink-rejection test** — `scripts/lib/inject-plugin-hooks.sh:128–133` refuses if `settings.local.json` / `installed_plugins.json` / `settings.json` is a symlink (explicit security property per `docs/PLUGINS.md:92`), but no test exercises any branch. A regression silently dropping the check would not be caught.

---

### Other observations (non-blocking, single reviewer each)

- **`realpath -m` fallback defeats containment** — when `realpath` is absent, the fallback returns the unresolved input, so `~/.claude/plugins/../../tmp/evil` would pass the prefix check. `realpath` is normally present in the image, but the fallback is misleading-secure. Either hard-fail or normalize `..` in bash. *(`inject-plugin-hooks.sh:91, 238`)*
- **Cross-FS non-atomic `mv`** — `mktemp` lands in `/tmp` (tmpfs in containers) but `$settings_local` is under `$HOME`. Across filesystems, `mv` is copy+unlink → readers can observe a partial file. Use `mktemp -p "$(dirname "$settings_local")"`. *(Pre-existing pattern in `inject-status-hooks.sh:145`.)*
- **Agent-type normalization order** — `INSTALL_PLUGINS` defaults to `true` only for `claude|claude-cli|claude-code`, but the case value is raw YAML. `type: Claude-CLI` would default to `false`. Consider normalizing first (`normalize_agent_type` at `launch-agent.sh:1825–1827`). *(`launch-agent.sh:773–780`)*
- **Invocation style inconsistency** — `entrypoint.sh` *executes* `inject-status-hooks.sh` (sibling) but *sources* `inject-plugin-hooks.sh` and calls `inject_plugin_hooks`. Both are valid; mirroring the sibling would be more consistent.
- **Logging fallback duplication** — the `log_info`/`log_warn` fallback block is now copy-pasted across three injection libs. Worth extracting to `lib/logging.sh` as a follow-up (out of scope here).
- **JSON-validity post-condition** — tests use `assert_contains` on raw strings; none runs `jq -e . settings.local.json` after merge. Cheap addition.

---

### Strengths (consensus across all three reviewers)

- Threat-model thinking is excellent: symlink refusal, `installPath` containment with `realpath -m` normalization, exact-match (not `inside()`) whitelist matching, fail-closed on corrupted whitelist — each backed by a dedicated regression test (`test_whitelist_exact_match_not_substring`, `test_installpath_traversal_blocked`, `test_whitelist_corruption_fails_closed`).
- File shape mirrors `scripts/lib/inject-lsp-config.sh` closely (source-guard, logging fallback, banner-divided sections) — easy to navigate for future maintainers.
- Schema validation on `installed_plugins.json` before extracting fields is defensive in the right way.
- Single-jq passes for both candidate enumeration and merge+count avoid fork explosions; intra-plugin dedup inside one `jq reduce` is the right correctness move.

---

*Review produced by 3-reviewer ensemble with majority-vote aggregation. Single-reviewer findings included where the issue is concrete and high-signal.*

---

aviadshiber:
@claude merge pr

Issue #328, opened 2026-05-12
overlay sandbox mode broken: CoW overlay AND atomic_copy_dir fallback both fail for every filesystem.include path


aviadshiber:
## Summary

In `overlay` sandbox mode, the entrypoint's *"Setting up CoW overlays for staged configs"* phase fails for **every** entry in `filesystem.include`. The `fuse-overlayfs` mount fails, the `atomic_copy_dir` fallback also fails, and the agent eventually exits 1 when `inject-status-hooks.sh:277` tries to write `/workspace/CLAUDE.md`.

10 of the 10 most recent agents on my host failed identically. The only agents that succeed are ones that get auto-detected as `worktree` mode (i.e. caller passed `--branch`). Pure `overlay` mode is currently 100% broken on this host.

## Versions / environment

- **Kapsis**: 2.25.3 (installed via Homebrew → `/opt/homebrew/Cellar/kapsis/2.25.3`)
- **macOS**: Darwin 25.3.0, M2 Pro
- **Podman machine**: `applehv`/`vfkit`, currently running
- **virtio-fs**: probed at the machine level — `/Users` mount is RW from a fresh `podman run` (so this is NOT the broader virtio-fs mount-drop class)

## Reproduction (from one of my agent logs, 2026-05-12 10:21)

```
[INFO] detect_sandbox_mode: Sandbox mode: overlay (auto-detected: no branch or not git repo)
…
[INFO] [entrypoint] Setting up CoW overlays for staged configs...
[WARN] [entrypoint] Overlay failed for .claude, falling back to atomic copy
[WARN] [entrypoint] atomic_copy_dir: cp failed for .claude
[WARN] [entrypoint] atomic_copy_dir: using fallback copy for .claude
[WARN] [entrypoint] Atomic copy validation failed for dir: .claude
[WARN] [entrypoint] Overlay failed for .ssh, falling back to atomic copy
[WARN] [entrypoint] atomic_copy_dir: cp failed for .ssh
[WARN] [entrypoint] atomic_copy_dir: using fallback copy for .ssh
[WARN] [entrypoint] Atomic copy validation failed for dir: .ssh
… (same for .slack, .kube, .config/.jira, .config/bkt, .config/mp, .jk, .candlekeep)
[INFO] [entrypoint] Sandbox mode: overlay
[INFO] [entrypoint] Timer [fuse_overlay_setup]: 0s elapsed
[...]

PR #329, opened 2026-05-12
fix(overlay): make status tracking work with read-only workspace (#328)

aviadshiber:
## Summary

Fixes the entrypoint crash that makes pure overlay-mode launches 100% broken on macOS (issue #328) while preserving the read-only guarantee of overlay mode. The whole point of overlay is that `/workspace` is RO — so status tracking now lives off-workspace, on the always-writable `/kapsis-status` named volume that's already mounted for status JSON.

**Scope:** status side only. The companion `fuse-overlayfs` / `atomic_copy_dir` failures for HOME-side staged configs (`.claude`, `.ssh`, ...) are a separate code path and stay out of this PR — they need their own follow-up.

## What was broken

In overlay mode:
1. `inject-status-hooks.sh` unconditionally appended a gist preamble to `/workspace/CLAUDE.md`. Read-only mount → raw bash redirect error → entrypoint exits 1 before the agent runs.
2. `KAPSIS_GIST_FILE` defaulted to `/workspace/.kapsis/gist.txt` — also unwritable.
3. `KAPSIS_INJECTED_TASK_SPEC` was exported but no consumer read it, so even the existing progress-instructions injection never actually reached the agent.

## What changed

| File | Change |
|------|--------|
| `scripts/lib/inject-status-hooks.sh` | Short-circuit `inject_gist_instructions()` in overlay mode (mirrors existing pattern); `[[ -w ]]` guards on `CLAUDE.md` / `AGENTS.md` appends; new `render_gist_instructions()` helper (substitutes `@@KAPSIS_GIST_FILE@@`), exposed via `--render-gist-instructions` |
| `scripts/launch-agent.sh` | In overlay mode, also export `KAPSIS_GIST_FILE=/kapsis-status/gist.txt` |
| `scripts/lib/status.sh` | `status_read_gist_file()` allow-prefix accepts `/kapsis-status/` in addition to `/workspace/.kapsis/`; `realpath` canonicalization still blocks traversal |
| `scripts/lib/gist-instructions.md` | `/workspace/.kapsis/gist.txt` → `@@KAPSIS_GIST_FILE@@` placeholder |
| `scripts/entrypoint.sh` | `inject_progress_instructions()` writes to `/kapsis-status` in overlay mode and appends rendered gist instructions; now called from the hook-capable branch in overlay mode so claude/codex/gemini agents get gist guidance in-band when CLAUDE.md/AGENTS.md surfacing is unavailable |
| `configs/{claude,codex,aider}.yaml` | Read task spec from `${KAPSIS_INJECTED_TASK_SPEC:-/task-spec.md}` so the rendered spec actually reaches the agent |
| `tests/test-status-hooks.sh` | +3 tests: overlay short-circuit, placeholder substitution, allow-prefix accepts `/kapsis-status/` and still rejects traversal |

## After this change

| Path | Worktree mode | Overlay mode |
|------|---------------|--------------|
| Status JSON | `/kapsis-status/...json` | `/kapsis-status/...json` |
| Agent gist file | `/workspace/.kapsis/gist.txt` | `/kapsis-status/gist.txt` |
| Task spec (rendered) | `/workspace/.kapsis/task-spec-with-progress.md` | `/kapsis-status/task-spec-with-progress.md` |
| `CLAUDE.md` / `AGENTS.md` gist preamble | appended (existing behaviour) | skipped (workspace stays untouched) |
| Entrypoint exit on RO `/workspace` | n/a | clean — no more "Read-only file system" |

## Test plan

- [x] `bash -n` syntax check on all modified scripts — clean
- [x] `bash ./tests/test-status-reporting.sh` — 27/27 pass
- [x] `bash ./tests/test-status-hooks.sh` — 64/64 pass (was 61, +3 new)
- [x] Manual render check: `KAPSIS_GIST_FILE=/kapsis-status/gist.txt ./scripts/lib/inject-status-hooks.sh --render-gist-instructions` produces the right path
- [x] Manual security check: `status_read_gist_file /kapsis-status/../etc/passwd` returns 0 with no state change
- [ ] **Container smoke (needs Podman host)**: launch any agent in overlay mode against a directory containing a `CLAUDE.md`; expect entrypoint completes with `Skipping gist instructions injection (overlay mode — read-only workspace)` and the agent runs to completion (`status.json` reaches phase `complete`, no `Read-only file system` error).
- [ ] **Worktree regression**: same launch with `--branch foo --base-branch main` still appends the gist preamble to `/workspace/CLAUDE.md` and writes `/workspace/.kapsis/gist.txt`.
- [ ] **Gist flow in overlay**: inside the container, `echo "doing X" > "$KAPSIS_GIST_FILE"` succeeds and the next `kapsis-status.sh --json` reports the gist text.
- [ ] **Task-spec delivery**: in overlay mode, `cat "$KAPSIS_INJECTED_TASK_SPEC"` shows the original task + progress instructions + gist instructions with `/kapsis-status/gist.txt` substituted.

## Out of scope

- HOME-side `fuse-overlayfs` / `atomic_copy_dir` failures for staged configs (issue #328 fix candidates #1–#3). Those need their own PR.
- Surfacing rendered instructions via `~/.claude/settings.local.json` system-prompt addition — the task-spec channel covers all three hook-capable agents uniformly, so no per-agent surfacing is needed for this fix.

https://claude.ai/code/session_01Sbaa3HjXG5RoraUN65iQ1s

---
_Generated by [Claude Code](https://claude.ai/code/session_01Sbaa3HjXG5RoraUN65iQ1s)_

### DIFF
--- a/configs/aider.yaml
+++ b/configs/aider.yaml
@@ -3,7 +3,10 @@
 
 agent:
   # Aider with message file input and auto-yes
-  command: "aider --yes-always --message-file /task-spec.md"
+  # KAPSIS_INJECTED_TASK_SPEC is set by entrypoint.sh when status injection
+  # appends progress/gist guidance to a writable copy of the task spec. Falls
+  # back to the read-only /task-spec.md mount when no injection happened.
+  command: "aider --yes-always --message-file \"${KAPSIS_INJECTED_TASK_SPEC:-/task-spec.md}\""
   workdir: /workspace
 
 filesystem:
--- a/configs/claude.yaml
+++ b/configs/claude.yaml
@@ -3,7 +3,10 @@
 
 agent:
   # Claude Code with spec file input
-  command: "claude --dangerously-skip-permissions -p \"$(cat /task-spec.md)\""
+  # KAPSIS_INJECTED_TASK_SPEC is set by entrypoint.sh when status injection
+  # appends progress/gist guidance to a writable copy of the task spec. Falls
+  # back to the read-only /task-spec.md mount when no injection happened.
+  command: "claude --dangerously-skip-permissions -p \"$(cat \"${KAPSIS_INJECTED_TASK_SPEC:-/task-spec.md}\")\""
   workdir: /workspace
 
 filesystem:
--- a/configs/codex.yaml
+++ b/configs/codex.yaml
@@ -3,7 +3,10 @@
 
 agent:
   # Codex CLI with full auto approval
-  command: "codex --approval-mode full-auto \"$(cat /task-spec.md)\""
+  # KAPSIS_INJECTED_TASK_SPEC is set by entrypoint.sh when status injection
+  # appends progress/gist guidance to a writable copy of the task spec. Falls
+  # back to the read-only /task-spec.md mount when no injection happened.
+  command: "codex --approval-mode full-auto \"$(cat \"${KAPSIS_INJECTED_TASK_SPEC:-/task-spec.md}\")\""
   workdir: /workspace
 
 filesystem:
--- a/docs/STATUS-TRACKING.md
+++ b/docs/STATUS-TRACKING.md
@@ -245,9 +245,20 @@ Each write overwrites the previous gist, so `kapsis-status` always shows the age
 ### Configuration
 
 ```bash
-# Default path (can be overridden)
+# Worktree mode default (writable /workspace)
 export KAPSIS_GIST_FILE=/workspace/.kapsis/gist.txt
+
+# Overlay mode default — /workspace is read-only, so the gist file lives on
+# the per-agent /kapsis-status volume (set automatically by launch-agent.sh)
+export KAPSIS_GIST_FILE=/kapsis-status/gist.txt
 ```
+
+In overlay mode, agents are told the active path via the injected task spec
+(rendered from `scripts/lib/gist-instructions.md` with `@@KAPSIS_GIST_FILE@@`
+substituted). `kapsis-status-hook.sh` reads whichever path
+`$KAPSIS_GIST_FILE` points to, and `status_read_gist_file()` in
+`scripts/lib/status.sh` accepts both `/workspace/.kapsis/` and
+`/kapsis-status/` (traversal still blocked by `realpath -m` canonicalization).
 
 ### Viewing Gists
 
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -1334,15 +1334,19 @@ inject_progress_instructions() {
 
     # Append instructions to task spec (copy to writable location first)
     local injected_spec="$kapsis_dir/task-spec-with-progress.md"
+    local write_err
+    write_err=$(mktemp)
     if ! {
         cat "$task_spec"
         echo ""
         echo ""
         cat "$instructions"
-    } > "$injected_spec" 2>/dev/null; then
-        log_warn "Could not write $injected_spec — skipping progress injection"
+    } > "$injected_spec" 2>"$write_err"; then
+        log_warn "Could not write $injected_spec — skipping progress injection: $(tr '\n' ' ' <"$write_err")"
+        rm -f "$write_err"
         return 0
     fi
+    rm -f "$write_err"
 
     # If gist tracking is enabled, render gist instructions with the live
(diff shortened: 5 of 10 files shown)


aviadshiber (commented):
## Ensemble Review (3-reviewer, majority vote)

**Verdict: APPROVE (3/3 unanimous)** — clean, scoped fix. No blocking issues.

Three independent reviewer agents audited the diff with different focus areas (conventions, security/cross-platform, correctness/regressions). Findings below are de-duplicated and ranked by reviewer consensus.

---

### Blocking
- none (3/3)

### Important — majority consensus (2+ reviewers flagged)

**1. `awk gsub` treats `&` as special in the replacement string** — `scripts/lib/inject-status-hooks.sh:292`
   Today `KAPSIS_GIST_FILE` is hardcoded to `/kapsis-status/gist.txt` or `/workspace/.kapsis/gist.txt`, so this is safe. But if a future user/config sets a path containing `&` or `\`, the substitution will be silently wrong. Either escape before substituting:
   ```bash
   gsub(/[\\&]/, "\\\\&", gf)
   ```
   or document the constraint.

**2. `rendered_gist=$(...)` is not declared `local`** — `scripts/entrypoint.sh:~1354`
   Leaks into the caller's scope. Pre-existing style violation per CLAUDE.md. Add `local rendered_gist` alongside the other locals.

### Important — singleton (1 reviewer)

- **Stderr suppression in overlay writer** (`scripts/entrypoint.sh:1342,1360`): `2>/dev/null` on `> "$injected_spec"` swallows the underlying errno (EROFS, ENOSPC, virtio-fs drop). Given this PR exists *because* of silent mount-related failures, suppressing stderr is the wrong default here — capture stderr to a temp and include it in the `log_warn`.
- **Test coverage gap**: the +3 new tests cover the skip-and-render side, but the new overlay-mode `inject_progress_instructions` writer (writes `/kapsis-status/task-spec-with-progress.md`, appends rendered gist when `KAPSIS_INJECT_GIST=true`, exports `KAPSIS_INJECTED_TASK_SPEC`) is uncovered.
- **Docs lag**: `docs/STATUS-TRACKING.md` and `docs/PLUGINS.md` still hardcode `/workspace/.kapsis/gist.txt` — accurate for worktree mode but worth a note that overlay mode now uses `/kapsis-status/gist.txt`.
- **`--render-gist-instructions` CLI verb** is undocumented in the script's `Usage:` header.

### Nits
- Defensive `mkdir -p "$kapsis_dir"` in the overlay branch would harden against future runtime changes (today /kapsis-status is always pre-created by the named volume).
- `render_gist_instructions` defaults gist_file to `/workspace/.kapsis/gist.txt` — if `KAPSIS_SANDBOX_MODE=overlay` but `KAPSIS_GIST_FILE` is somehow unset (direct manual exec), rendered text would point at a read-only path. `launch-agent.sh` always sets it, so this is theoretical.
- `test_status_read_gist_accepts_kapsis_status_volume` only asserts rejection cases; a positive acceptance assertion (a real file under `/kapsis-status/...` is accepted) would match the test name.
- `realpath -m` is GNU-only but only runs inside the (Linux) container — not a regression.

---

### Positives (unanimous)
- **Pattern consistency**: overlay short-circuit in `inject_gist_instructions()` mirrors the existing `inject_progress_instructions()` pattern.
- **Worktree path is byte-equivalent**: marker preserved, trailing-newline equivalent, `grep -q "$marker"` re-entry guard still works, `cp` → `printf '%s\n'` produces equivalent file permissions under standard umask.
- **`[[ -w ]]` guards** on CLAUDE.md / AGENTS.md prevent the original crash mode defensively.
- **Traversal safety preserved**: `realpath -m` canonicalization + trailing-slash prefix check correctly rejects `/kapsis-status/../etc/passwd` and `/kapsis-status-evil/...`.
- **Graceful env-var fallback**: `${KAPSIS_INJECTED_TASK_SPEC:-/task-spec.md}` in all three agent configs means partial-injection failures still leave the agent with a usable spec.
- **awk-based `@@KAPSIS_GIST_FILE@@` substitution** sidesteps sed-delimiter escaping pitfalls.
- **CLI verb** (`--render-gist-instructions`) enables clean reuse from `entrypoint.sh` without re-sourcing.

---

*Generated by ensemble review (3 independent reviewer agents, majority vote on verdict).*

---

aviadshiber:
@claude review with ensemble and fix

PR #332, opened 2026-05-12
fix(overlay): harden HOME-side CoW + atomic-copy fallback (#328 followup)

aviadshiber:
## Summary

Follow-up to PR #329 (which fixed the workspace-side breakage). This PR addresses the remaining HOME-side breakage from issue #328 — fix candidates **#1**, **#2**, **#3** that PR #329 explicitly listed as out-of-scope.

In overlay mode every entry in `filesystem.include` was failing identically with a cryptic `cp failed` warning, and the actual `fuse-overlayfs` error was swallowed by `2>/dev/null`. This PR makes the failure debuggable and lets the atomic-copy fallback succeed in more cases.

### `scripts/lib/atomic-copy.sh`
- **Surface mktemp / cp stderr** in `log_warn` instead of redirecting to `/dev/null`. Previously the operator only saw `mktemp failed` or `cp failed`; now they see the underlying `Read-only file system` / `Permission denied` / etc.
- **Cross-filesystem scratch fallback** in `atomic_copy_dir` and `atomic_copy_file`: when mktemp inside `dirname(dst)` fails, stage the copy under a writable scratch directory (`KAPSIS_SCRATCH_DIR`, default `/tmp`) and then do a best-effort non-atomic copy into `$dst`. Atomicity is lost in this degraded path, but it lets the copy succeed when `$dst` itself is writable but its parent isn't — the exact case the issue describes.

### `scripts/entrypoint.sh::setup_staged_config_overlays`
- **Pre-check `$HOME` writability**: if HOME isn't writable, emit a single warning listing the likely causes (paranoid security profile, missing chown on container HOME, broken bind-mount over HOME) and skip the whole loop. Replaces N cryptic "cp failed" lines (one per staged config) with one actionable diagnostic.
- **Surface `fuse-overlayfs` stderr** in the overlay-failed warning. Previously redirected to `/dev/null`, which forced bisecting from `Overlay failed for X` to find the real cause (kernel/fuse compat, missing capability, LSM denial, etc.).

### `tests/test-atomic-copy.sh`
- +4 tests for the new fallback paths and surfaced-error behaviour:
  - `atomic_copy_dir` succeeds via scratch when `mktemp`-in-parent fails but `$dst` is writable.
  - `atomic_copy_dir` fails cleanly with surfaced stderr when both parent and dst are RO.
  - `atomic_copy_dir` warning includes the underlying mktemp stderr text (not a generic message).
  - `atomic_copy_file` warning surfaces the underlying mktemp/cp stderr.
- Tests use `mktemp`/`cp` function overrides rather than chmod-based RO simulation, because chmod 555 is bypassed by root and several CI environments run tests as root.

## Test plan

- [x] `bash -n` on all modified scripts — clean
- [x] `shellcheck scripts/entrypoint.sh scripts/lib/atomic-copy.sh tests/test-atomic-copy.sh` — clean
- [x] `bash tests/test-atomic-copy.sh` — 28/28 pass (was 24, +4 new)
- [x] `./tests/run-all-tests.sh --quick` — pre-existing failures only (none introduced by this PR; verified by re-running the failing scripts on `main`-state via `git stash`)
- [ ] **Container smoke (needs Podman host)**: launch an agent in overlay mode against a host project; expect the `Setting up CoW overlays for staged configs...` phase either succeeds, or — if `fuse-overlayfs` is genuinely broken on the host — emits a single warning per entry that includes the actual `fuse-overlayfs` stderr (instead of `Overlay failed for X, falling back to atomic copy`).
- [ ] **Container smoke with simulated RO HOME**: chmod 555 the container HOME via a debug entrypoint; expect a single `HOME (...) is not writable — staged configs cannot be installed` warning + the listed likely causes, and the entrypoint proceeds rather than exiting from cascading errors.
- [ ] **Worktree regression**: re-run a worktree-mode agent; expect identical behaviour (no scratch/RO path triggered, no regression).

## Out of scope

- The exit-6-on-no-commit observation from issue #328 comment 1 — the comment itself says "worth filing separately if that's not already an option". Not part of the listed fix candidates #1–#3.
- The root-cause investigation of *why* `fuse-overlayfs` fails inside the new RHEL-style image. With this PR the actual stderr is finally in the logs, so that investigation can happen on real failure traces.

https://claude.ai/code/session_01Sbaa3HjXG5RoraUN65iQ1s

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ex72Qxx8PQxKvPWYjbX89M)_

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -607,8 +607,23 @@ setup_staged_config_overlays() {
 
     log_info "Setting up CoW overlays for staged configs..."
 
-    # Create base directories for upper and work layers
-    mkdir -p "$upper_base" "$work_base" 2>/dev/null || true
+    # Issue #328: HOME must be writable for either the fuse-overlayfs mount
+    # or the atomic_copy fallback to succeed. If it isn't, every staged config
+    # fails identically with a cryptic "cp failed" warning. Detect it once
+    # up front and tell the operator what to look at.
+    if [[ ! -w "$HOME" ]]; then
+        log_warn "HOME ($HOME) is not writable — staged configs cannot be installed"
+        log_warn "Likely causes: read-only root filesystem (security profile=paranoid), missing chown on container HOME, or a broken bind-mount over HOME"
+        log_warn "Continuing without staged configs; agents that need credentials may fail to authenticate"
+        return 0
+    fi
+
+    # Issue #328: upper/work layers live on the container writable layer
+    # (not under HOME). Surface mkdir errors instead of swallowing them so a
+    # broken root layer fails loudly rather than cascading into every overlay.
+    if ! mkdir -p "$upper_base" "$work_base" 2>/dev/null; then
+        log_warn "Failed to create overlay state dirs ($upper_base, $work_base) — falling back to copy for every entry"
+    fi
 
     # Split comma-separated list
     IFS=',' read -ra configs <<< "$KAPSIS_STAGED_CONFIGS"
@@ -631,11 +646,22 @@ setup_staged_config_overlays() {
             # Directory: create overlay mount
             mkdir -p "$dst" 2>/dev/null || true
 
-            if fuse-overlayfs -o "lowerdir=${src},upperdir=${upper},workdir=${work}" "$dst" 2>/dev/null; then
+            # Issue #328: capture fuse-overlayfs stderr so failures are
+            # debuggable. Previously redirected to /dev/null, which forced
+            # users to bisect from a generic "Overlay failed" warning.
+            local _overlay_stderr _overlay_rc
+            _overlay_stderr=$(fuse-overlayfs -o "lowerdir=${src},upperdir=${upper},workdir=${work}" "$dst" 2>&1) && _overlay_rc=0 || _overlay_rc=$?
+
+            if [[ $_overlay_rc -eq 0 ]]; then
                 log_debug "CoW overlay: ${relative_path}"
             else
                 # Fallback: atomic copy with validation (fixes race condition #151)
-                log_warn "Overlay failed for ${relative_path}, falling back to atomic copy"
+                if [[ -n "$_overlay_stderr" ]]; then
+                    log_warn "Overlay failed for ${relative_path} (rc=${_overlay_rc}): ${_overlay_stderr}"
+                else
+                    log_warn "Overlay failed for ${relative_path} (rc=${_overlay_rc}, no stderr)"
+                fi
+                log_warn "Falling back to atomic copy for ${relative_path}"
                 atomic_copy_dir "$src" "$dst" || log_warn "Atomic copy validation failed for dir: ${relative_path}"
             fi
         else
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -611,10 +611,22 @@ setup_staged_config_overlays() {
     # or the atomic_copy fallback to succeed. If it isn't, every staged config
     # fails identically with a cryptic "cp failed" warning. Detect it once
     # up front and tell the operator what to look at.
+    #
+    # Review finding #3: downstream auth/credential paths need to know the
+    # configs were skipped so they can fail with a clear "auth will not work"
+    # message rather than a cryptic "key not found". Export a state flag and
+    # write a structured sentinel to /kapsis-status so kapsis-status.sh can
+    # surface it. Still return 0 — the entrypoint continues so the agent can
+    # report the auth-skipped state cleanly, but no caller mistakes the
+    # absence of staged configs for a successful install.
     if [[ ! -w "$HOME" ]]; then
-        log_warn "HOME ($HOME) is not writable — staged configs cannot be installed"
+        log_warn "KAPSIS-AUTH-WARNING: HOME ($HOME) is not writable — staged configs cannot be installed"
         log_warn "Likely causes: read-only root filesystem (security profile=paranoid), missing chown on container HOME, or a broken bind-mount over HOME"
-        log_warn "Continuing without staged configs; agents that need credentials may fail to authenticate"
+        log_warn "Downstream agents that need credentials (.ssh, .claude, .config/...) WILL fail to authenticate"
+        export KAPSIS_STAGED_CONFIGS_SKIPPED=1
+        if [[ -d /kapsis-status ]] && [[ -w /kapsis-status ]]; then
+            printf 'HOME_NOT_WRITABLE\n' > /kapsis-status/staged-configs-skipped 2>/dev/null || true
+        fi
         return 0
     fi
 
--- a/scripts/lib/atomic-copy.sh
+++ b/scripts/lib/atomic-copy.sh
@@ -117,18 +117,38 @@ atomic_copy_file() {
     # Ensure destination directory exists
     mkdir -p "$(dirname "$dst")" 2>/dev/null || true
 
-    # Create temp file in same directory as dst (required for atomic mv)
-    local tmp_file
-    tmp_file=$(mktemp "$(dirname "$dst")/.atomic-copy-XXXXXX") || {
-        log_warn "atomic_copy_file: mktemp failed for $(basename "$dst")"
-        # Fallback to direct copy
-        cp -p "$src" "$dst" 2>/dev/null || true
+    # Create temp file in same directory as dst (required for atomic mv).
+    # Issue #328: capture mktemp stderr so an unwritable parent (read-only
+    # HOME, restrictive perms, broken bind mount) gives a debuggable error
+    # instead of a silent "mktemp failed".
+    local tmp_file mktemp_err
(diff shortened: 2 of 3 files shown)


aviadshiber (commented):
## Ensemble Review (3 reviewers, majority vote ≥2/3)

Three independent reviewers analyzed the diff; only findings raised by ≥2 of 3 are reported below. Single-voter nits were filtered out.

### Findings

**1. `atomic-copy.sh:215-219` — major (B+C)**
The "last resort" branch runs `cp -rp "$src/." "$dst/"` and unconditionally `return 1`, even when the direct `cp` succeeded. The caller in `entrypoint.sh:665` then emits `Atomic copy validation failed for dir: …` while the data has in fact been placed correctly — a misleading warning that masks success and re-creates the exact "everything fails" symptom #328 was trying to eliminate.

*Fix:* Capture the cp rc; return 0 when the direct cp succeeded (or document this as intentionally pessimistic and adjust the caller log).

**2. `atomic-copy.sh:238-245` — major (A+C)**
The scratch-fallback validation compares `_atomic_count_files "$src"` to `_atomic_count_files "$dst"`, but `$dst` is not reset before the stage cp. If `$dst` pre-existed with files (CoW fallback into an already-populated `$HOME/.ssh`, retries, etc.), the counts mismatch and the function returns 1 despite every src file being copied. The symmetric case — stale files in `$dst` whose paths happen to overlap src — lets validation falsely pass, leaving stale state mixed with fresh.

*Fix:* Mirror the happy-path contract — either `find "$dst" -mindepth 1 -delete` before staging, or count `tmp_dir` vs `src` before the stage cp (not `dst` after).

**3. `entrypoint.sh:614-619` — major (B+C)**
When `HOME` is unwritable, `setup_staged_config_overlays` emits warnings and `return 0`. The caller has no way to know that no configs were installed — downstream auth steps proceed and the agent fails later with cryptic credential errors, the exact debuggability problem this PR is trying to fix.

*Fix:* Either `return 1` so the caller can decide, or set a state flag (e.g. `KAPSIS_STAGED_CONFIGS_SKIPPED=1`) consumed by credential-injection paths, or surface a single visible "auth will likely fail" warning at agent-start. Consider whether `paranoid` profile should treat this as fatal.

**4. `atomic-copy.sh:125, 197` — minor (A+B+C, unanimous)**
`mktemp_err=$(mktemp ... 2>&1) && tmp_file="$mktemp_err"` conflates stdout and stderr on success. If `mktemp` ever emits even one stderr byte while still returning 0 (libc warnings, locale messages, a custom wrapper, `set -x` trace leakage), `tmp_file` becomes `"garbage\npath"` and the subsequent `[[ ! -e "$tmp_file" ]]` test fails, silently routing real success into the fallback (which is non-atomic for `atomic_copy_file`).

*Fix:* Capture stderr only on failure, e.g.
```bash
tmp_file=$(mktemp "$(dirname "$dst")/.atomic-copy-XXXXXX" 2>/tmp/.mktemp.$$.err) \
  || mktemp_err=$(<"/tmp/.mktemp.$$.err")
rm -f "/tmp/.mktemp.$$.err"
```
or an `if … then … else … fi` form that doesn't reuse the captured-stderr variable for the success path.

**5. `atomic-copy.sh:206` — minor (A+B+C, unanimous)**
`KAPSIS_SCRATCH_DIR` defaults to `/tmp`, where credential filenames/sizes (SSH keys, API tokens) are visible in the directory listing on multi-user hosts. `mktemp -d` creates the scratch dir at `0700` so contents are protected, but the parent listing isn't. The variable is also a new public knob — not declared in `scripts/lib/constants.sh`, not documented in `docs/CONFIG-REFERENCE.md` or `agent-sandbox.yaml.template`.

*Fix:* Prefer `${XDG_RUNTIME_DIR:-${TMPDIR:-/tmp}}` as default (typically `/run/user/$UID`, mode `0700`); add the variable to constants + reference docs.

**6. `tests/test-atomic-copy.sh:574-579` — minor (A+B)**
`_reset_atomic_copy_lib` `unset -f mktemp` but does not unset `cp`. Tests that override `cp` (`test_atomic_copy_dir_ro_parent_ro_dst`, `test_atomic_copy_file_surfaces_cp_stderr`) rely on a manual `unset -f cp` after the reset. `run_test` invokes test functions in the same shell, so an early abort before the manual unset would leak the override into later tests.

*Fix:* Move `unset -f cp 2>/dev/null || true` into `_reset_atomic_copy_lib`, and ideally arm it via `trap … RETURN` inside each overriding test.

### Verified non-issues (consensus)

- `shellcheck scripts/entrypoint.sh scripts/lib/atomic-copy.sh tests/test-atomic-copy.sh` — clean.
- `local cp_err` re-declaration in the same function is legal bash; no shadowing bug.
- Bash function overrides for `mktemp`/`cp` correctly propagate into the sourced library (dynamic-scope lookup).
- `_KAPSIS_ATOMIC_COPY_LOADED` source-guard is present, so `_reset_atomic_copy_lib` correctly re-sources.
- `$tmp_dir` is cleaned up on every exit path inside the scratch branch.
- `set -e` does not abort on any of the new `x=$(cmd 2>&1) && y=…` constructs — all are in `if`/`||` contexts.
- `KAPSIS_ROOT` defined at `tests/lib/test-framework.sh` load.
- `local _overlay_stderr _overlay_rc; _overlay_stderr=$(…) && _overlay_rc=0 || _overlay_rc=$?` preserves the actual rc.
- The tail "Fallback: copy directly" block (`atomic_copy_dir`, ~line 280) still returns 1 — unchanged behavior.

### Coverage gaps (majority)

- No test exercises a pre-populated `$dst` going through the scratch-fallback (would catch finding #2).
- No test for `atomic_copy_file`'s scratch / fallback success path — only the warning text is asserted.
- No test for `setup_staged_config_overlays`'s HOME-unwritable early return.
- No assertion that the "last resort" branch returns the correct code when direct cp succeeds (finding #1).

### Overall recommendation

**comment** — A+B chose `comment`; C chose `request-changes`. Findings #1–#3 are non-trivial reliability concerns that meaningfully extend the original #328 symptom into adjacent scenarios; worth resolving before merge but not strict blockers given the PR is a clear net improvement to debuggability.

---
*Posted by an ensemble of 3 independent reviewers; only findings raised by ≥2/3 are included.*

---

aviadshiber:
Addressed the ensemble-review findings in commit 7e79b99 (rebased onto current branch tip).

| # | Finding | Resolution |
|---|---|---|
| 1 | Last-resort `cp` returns 1 even on success (major) | `atomic_copy_dir` now validates file count after the last-resort direct `cp -rp` and returns 0 when the copy actually succeeded. No more false "validation failed" warnings over intact data. |
| 2 | Scratch-fallback validates against pre-populated dst (major) | Scratch stage cp now `find -mindepth 1 -delete`s existing dst contents before staging (preserving dst's own perms), so retries and pre-populated `$HOME/.ssh` work correctly. |
| 3 | HOME-unwritable `return 0` hides failure (major) | Exports `KAPSIS_STAGED_CONFIGS_SKIPPED=1` and writes a `/kapsis-status/staged-configs-skipped` sentinel; warning tagged `KAPSIS-AUTH-WARNING:` for log scraping. Still returns 0 so the agent continues and can report cleanly. |
| 4 | mktemp stdout/stderr conflation (minor, unanimous) | New `_atomic_run_mktemp` helper captures stderr via a per-call temp file; stdout (the path) is no longer poisoned by stderr noise. Verified by a new test that injects stderr during a successful `mktemp`. |
| 5 | `KAPSIS_SCRATCH_DIR` defaults to `/tmp` (minor, unanimous) | New `_atomic_scratch_base` helper defaults to `XDG_RUNTIME_DIR` (typically `/run/user/$UID`, mode 0700) then `TMPDIR` then `/tmp`. Scratch base gets `chmod 0700`. Scratch fallback also now `chmod`s a newly-created dst to the source's dir mode so `.ssh` no longer drops from 0700 to 0755 on the fallback path. |
| 6 | `_reset_atomic_copy_lib` doesn't unset `cp` (minor) | Helper now `unset -f cp` too, so mocks can't leak across tests when an assertion aborts before a manual unset. |

**Test coverage added:**
- `test_atomic_copy_dir_scratch_resets_prepopulated_dst` (finding #2)
- `test_atomic_copy_dir_last_resort_returns_success_when_cp_works` (finding #1)
- `test_atomic_copy_dir_mktemp_stderr_does_not_corrupt_path_on_success` (finding #4)

All 31 tests pass; shellcheck clean.

**Not addressed (intentional):**
- "No test for `setup_staged_config_overlays` HOME-unwritable early return" — the function is tightly coupled to entrypoint state (sourced inside a `set -euo pipefail` script with many globals); a unit test would need either function extraction or a podman integration test. Filing separately if it doesn't get covered by the integration suite.
- "Document `KAPSIS_SCRATCH_DIR` in CONFIG-REFERENCE" — kept as an undocumented escape hatch for now since the new XDG-based default makes manual configuration unlikely.

---

PR #333, opened 2026-05-13
fix(staging): add :U mount flag + post-staging chmod for macOS UID mismatch (#328)

aviadshiber:
## Summary

Complements commit 6c11151 (#334) which already fixed the AF_UNIX socket `cp` ENOENT root cause. This PR adds two remaining gaps not covered by that commit:

- **`launch-agent.sh`**: Add `:U` Podman mount option to staging volume mounts on macOS. On macOS, the host user's UID (e.g. 501) maps to a different UID inside the container while the container process runs as `developer` (UID 1000). Directories with mode `0700` (`.ssh`, `.claude`) owned by the host UID appear inaccessible to UID 1000 — both `fuse-overlayfs` and `cp` fail with `EACCES`. The `:U` idmapped-mount option (Podman 4.3+, kernel 5.12+) remaps ownership to the container user so those dirs become readable. `:ro` is preserved; the host is never modified.

- **`entrypoint.sh`**: After `atomic_copy_dir`, run `find -type f -exec chmod u+rw` on the destination. Defense-in-depth: source files with mode `0444`/`0400` (copied verbatim via `cp -p`) would block downstream writers like `inject-lsp-config.sh` and `filter-agent-config.sh` that need to modify `settings.local.json`.

- **`tests/test-snapshot-staging.sh`**: Three new quick tests for the `:U` flag platform logic (macOS includes `U`, Linux does not, both `ro` and `U` present on macOS).

> **Note:** The original `atomic-copy.sh` changes from this branch have been dropped — they were superseded by 6c11151's more precise `_atomic_cp_stderr_is_stat_enoent_only` / `_atomic_cp_with_enoent_tolerance` implementation. This PR now contains only the two complementary changes above.

## Relationship to #334 / 6c11151

| Problem | Fixed by |
|---------|----------|
| `cp -rp` exits 1 on AF_UNIX socket ENOENT during readdir | 6c11151 (`atomic-copy.sh`) |
| mode-0700 dirs EACCES to container user (UID mismatch, macOS) | This PR (`:U` in `launch-agent.sh`) |
| Source files with mode `0444` block downstream config writers | This PR (`chmod u+rw` in `entrypoint.sh`) |

## Test plan

- [ ] Quick tests pass: `./tests/run-all-tests.sh --quick`
- [ ] New snapshot staging tests pass: `./tests/test-snapshot-staging.sh`
- [ ] On macOS: launch an agent against a project where `~/.claude` or `~/.ssh` is mode `0700` — confirm no `EACCES` in staging copy logs
- [ ] Verify `:ro` is still enforced (no writes reach host staging dirs)

https://claude.ai/code/session_01R8aTDCN9m81HUpQHyMwxbG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,14 +6,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
-## [2.28.1] - 2026-05-12
+## [2.28.2] - 2026-05-13
 
 ### Fixed
-- Harden HOME-side CoW + atomic-copy fallback (#328 followup) (#332)
+- Tolerate benign cp stderr from unstattable readdir entries (#328 root cause) (#334)
 
 ## [Unreleased]
 
 ## [2.1.1] - 2026-02-02
+
+### Fixed
+- Harden HOME-side CoW + atomic-copy fallback (#328 followup) (#332)
+
 
 ### Added
 - Persist agent JSONL via shared host volume (Issue #265) (#322)
@@ -924,7 +928,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Setup and installation guide
 - Contributing guidelines
 
-[Unreleased]: https://github.com/aviadshiber/kapsis/compare/v2.28.1...HEAD
+[Unreleased]: https://github.com/aviadshiber/kapsis/compare/v2.28.2...HEAD
+[2.28.2]: https://github.com/aviadshiber/kapsis/releases/tag/v2.28.2
 [2.28.1]: https://github.com/aviadshiber/kapsis/releases/tag/v2.28.1
 [2.28.0]: https://github.com/aviadshiber/kapsis/releases/tag/v2.28.0
 [2.27.0]: https://github.com/aviadshiber/kapsis/releases/tag/v2.27.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,14 +6,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
-## [2.28.2] - 2026-05-13
+## [2.28.3] - 2026-05-13
 
 ### Fixed
-- Tolerate benign cp stderr from unstattable readdir entries (#328 root cause) (#334)
+- Broaden benign-stderr classifier + chmod restrictive dst (#335) (#336)
 
 ## [Unreleased]
 
 ## [2.1.1] - 2026-02-02
+
+### Fixed
+- Tolerate benign cp stderr from unstattable readdir entries (#328 root cause) (#334)
+
 
 ### Fixed
 - Harden HOME-side CoW + atomic-copy fallback (#328 followup) (#332)
@@ -928,7 +932,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Setup and installation guide
 - Contributing guidelines
 
-[Unreleased]: https://github.com/aviadshiber/kapsis/compare/v2.28.2...HEAD
+[Unreleased]: https://github.com/aviadshiber/kapsis/compare/v2.28.3...HEAD
+[2.28.3]: https://github.com/aviadshiber/kapsis/releases/tag/v2.28.3
 [2.28.2]: https://github.com/aviadshiber/kapsis/releases/tag/v2.28.2
 [2.28.1]: https://github.com/aviadshiber/kapsis/releases/tag/v2.28.1
 [2.28.0]: https://github.com/aviadshiber/kapsis/releases/tag/v2.28.0
--- a/packaging/debian/debian/changelog
+++ b/packaging/debian/debian/changelog
@@ -1,3 +1,9 @@
+kapsis (2.28.3-1) unstable; urgency=medium
+
+  * Update to v2.28.3
+
+ -- Aviad Shiber <aviadshiber@gmail.com>  Wed, 13 May 2026 19:16:18 +0000
+
 kapsis (2.28.2-1) unstable; urgency=medium
 
   * Update to v2.28.2
--- a/packaging/debian/debian/changelog
+++ b/packaging/debian/debian/changelog
@@ -1,3 +1,9 @@
+kapsis (2.28.2-1) unstable; urgency=medium
+
+  * Update to v2.28.2
+
+ -- Aviad Shiber <aviadshiber@gmail.com>  Wed, 13 May 2026 08:49:18 +0000
+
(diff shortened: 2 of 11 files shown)


aviadshiber:
@claude is this pr still needed given pr https://github.com/aviadshiber/kapsis/pull/334 ?

aviadshiber:
Yes, this PR is still needed — but only partially overlaps with #334.

**What #334 already covered:** the `atomic-copy.sh` socket-file tolerance (`_atomic_cp_stderr_is_stat_enoent_only` helper + `_atomic_cp_with_enoent_tolerance` wrapper).

**What this PR still contributes that isn't in main:**

1. **`launch-agent.sh` — `:U` idmapped mount flag (primary fix):** Adds `ro,U` on macOS so mode-`0700` dirs (`.ssh`, `.claude`) appear owned by the container user. This is the root-cause fix for the EACCES on `fuse-overlayfs` — without it, overlay staging still fails even after #334.

2. **`entrypoint.sh` — `chmod u+rw` after copy:** Ensures `inject-lsp-config.sh` and `filter-agent-config.sh` can write `settings.local.json` regardless of source permissions.

3. **`test-snapshot-staging.sh` — 3 new tests:** Cover the `:U` mount option logic (macOS includes `U`, Linux does not, both `ro` and `U` present on macOS).

The PR currently has a merge conflict in `atomic-copy.sh` because both this branch and #334 touched that file. I can rebase to drop the overlapping `atomic-copy.sh` changes (main's implementation is strictly better) and keep the three unique contributions above. Want me to do that?

---

aviadshiber:
## Ensemble Review (5 independent reviewers, majority vote)

Five independent reviewer agents reviewed this PR with different framings (correctness, security, testing, portability, style). Verdict by majority vote.

### Vote tally

| Reviewer focus | Verdict | Confidence |
|---|---|---|
| Correctness | COMMENT | medium |
| Security | **REQUEST_CHANGES** | **high** |
| Testing | COMMENT | medium |
| Portability | APPROVE | medium |
| Style/conventions | COMMENT | medium |

**Majority: `COMMENT` (3/5).** One strong dissent (REQUEST_CHANGES, high confidence) on the security finding below.

---

### Consensus concerns (raised by ≥2 reviewers)

#### 1. `:U` semantics may be misrepresented in the code comment — 2/5 reviewers, both at medium–high confidence

The comment in `scripts/launch-agent.sh:1458-1467` states `:U` "creates an idmapped mount" and that "the host filesystem is never modified." Per `man podman-run`:

> `U`: This option tells Podman to chown the source volume to match the default UID and GID used within the container.

That is a **recursive chown of the source**, not an idmapped mount (the idmapped flag is `:idmap`). For directory mounts (`~/.ssh`, `~/.claude`, `~/.config/git`), `mount_source` is the live host path — only individual files are routed through `_snapshot_file` (see `launch-agent.sh:1452-1456`). Implications flagged:

- On macOS, the chown executes inside the Podman VM against the virtio-fs view; whether it propagates back to the macOS host depends on virtio-fs behavior and is undocumented. Worst case the user's host `~/.ssh` becomes unusable to `sshd`/`ssh` (which refuse keys with wrong owner/perms).
- Concurrency: parallel agents with the same staged source would race on the recursive chown, conflicting with this project's hermetic-isolation guarantee.
- Performance: every launch performs a recursive chown of every file under `~/.claude`.

The reviewer raising this at high confidence recommends one of: `--userns=keep-id` adjustments, `:idmap` (true kernel idmapping), or snapshotting directories the same way files are snapshotted.

**Action:** at minimum, correct the inline comment to accurately describe `:U`. Ideally, verify the actual on-macOS behavior empirically (does the chown propagate to the Mac host after a rootless-Podman run?) and document the outcome in the comment.

#### 2. New tests do not exercise the production code path — 4/5 reviewers

The helper `_staging_mount_opts_for_platform()` in `tests/test-snapshot-staging.sh:356-363` is an **inlined re-implementation** of the `if is_macos; then ...` block in `launch-agent.sh:1468-1471`. The three new tests assert against this duplicate. If someone removes the `is_macos` guard, drops the `,U`, or changes the production logic, the tests still pass green.

**Suggested fix:** extract a tiny helper (`staging_mount_opts_for_platform`) into `scripts/lib/compat.sh` (or similar) and call it from both `launch-agent.sh` and the tests. Alternatively, source `launch-agent.sh` with `is_macos` mocked and assert on the constructed `VOLUME_MOUNTS` array.

Bonus gap: the `chmod u+rw` block in `entrypoint.sh` (the more impactful half of the fix) has **zero test coverage** and is straightforwardly unit-testable in quick mode (create a 0444 file, simulate the fallback, assert perms became u+rw).

---

### Other notable items (single-reviewer)

- **No Podman version preflight (R4):** `:U` requires Podman 4.3+ / kernel 5.12+. `scripts/preflight-check.sh` only checks for Podman presence. On stale macOS Podman installs, `podman run` fails with an opaque "invalid mount option" rather than a clear remediation message.
- **Document the platform asymmetry (R4):** the root cause is `--userns=keep-id` in `scripts/lib/security.sh:208` combined with macOS host UID 501 vs. container `developer` UID 1000. A one-line "Linux host UID 1000 ≈ container developer UID 1000, so keep-id alone is sufficient" in the comment would prevent future confusion.
- **`find ... chmod ... 2>/dev/null || true` (R1):** swallows real chmod failures; the whole point is making files writable, so a silent failure leaves the downstream writer to hit EACCES with no breadcrumb. Consider logging at debug, or dropping the `|| true` since `find ... +` returns 0 if any invocation succeeded.
- **Chmod-after-failed-copy (R5):** the chmod runs even when `atomic_copy_dir ... || log_warn ...` failed. Likely intentional defence-in-depth, but worth a one-line note acknowledging the fall-through.
- **Non-home absolute-path branch at `launch-agent.sh:1488` (R1):** same `:ro` (no `:U`) — would have the identical UID-mismatch failure mode for any included mode-0700 directory outside `$HOME`. Not a regression; just a coverage gap.

### Nits (non-blocking)

- Root-skip uses `[[ "$(id -u)" == "0" ]]` — existing project pattern is numeric `[[ $(id -u) -eq 0 ]]` (e.g. `test-workspace-mount-validation.sh:397`). `[[ $EUID -eq 0 ]]` would avoid the subshell entirely.
- Underscore-prefixed `local _staging_mount_opts`: R4 noted this as anomalous, R5 found 22 such locals already in `launch-agent.sh` — actually consistent with the file's existing style.
- Conventional-commit prefix `fix(staging):` is valid.

---

### Summary

The fix's *intent* is correct and the chmod safety-net half is uncontroversial. The two substantive concerns the ensemble agrees on are **(a)** the `:U` flag's documented semantics (recursive source chown) don't match the in-code rationale, which deserves at minimum a corrected comment and ideally a verified-on-macOS behavior note, and **(b)** the new tests verify the test file's own re-implementation rather than the production code in `launch-agent.sh`, providing near-zero regression protection.

Majority verdict is `COMMENT` rather than `REQUEST_CHANGES`, but the security reviewer's dissent is worth addressing before merge.

_Generated by the ensemble review skill — 5 independent reviewer agents, majority vote._

---

Issue #164, opened 2026-02-15
Bind-mount staging race condition: host file writes corrupt container copies


aviadshiber:
## Summary

Files listed in `filesystem.include` that are actively written by host processes (e.g., `~/.claude.json` by Claude Code) can arrive corrupted inside the container. The `atomic-copy.sh` fix from PR #156 makes the **write** to the destination atomic, but does not protect against the **read** from the bind-mounted source being torn by a concurrent host write.

## Evidence

Two confirmed occurrences on the same host:

| Date | Agent | Error | Corruption Type |
|------|-------|-------|-----------------|
| 2026-02-13 | `e6c6f1` | `JSON Parse error: Unexpected EOF` | Truncated mid-write |
| 2026-02-15 | `f82937` | `JSON Parse error: Property name must be a string literal` | Partial/inconsistent content |

Debug logs:
```
# Feb 15 — agent f82937
Configuration error in /home/developer/.claude.json: JSON Parse error: Property name must be a string literal
Claude configuration file at /home/developer/.claude.json is corrupted: JSON Parse error: Property name must be a string literal
The corrupted file has been backed up to: /home/developer/.claude.json.corrupted.1771172255649

# Feb 13 — agent e6c6f1
Claude configuration file at /home/developer/.claude.json is corrupted: JSON Parse error: Unexpected EOF
The corrupted file has been backed up to: /home/developer/.claude.json.corrupted.1770933972721
```

The host file (`~/.claude.json`, 64KB) validates fine with both `node` and `python3` JSON parsers — the corruption only exists in the container copy.

### The staging flow

1. **Host side** (`launch-agent.sh:1089`): The live host file is bind-mounted read-only:
   ```
   -v ~/.claude.json:/kapsis-staging/.claude.json:ro
   ```

2. **Container side** (`entrypoint.sh:344`): `atomic_copy_file` copies from the bind mount:
   ```bash
   cp -p /kapsis-staging/.claude.json /home/developer/.atomic-copy-XXXXXX   # line 131
   mv /home/developer/.atomic-copy-XXXXXX /home/developer/.claude.json      # line 133
   ```

3. **Validation** (`atomic-copy.sh:137`): Compares destination against source:
   ```bash
   _atomic_validate_file "$src" "$dst"   # size check + jq validation
   ```

### The problem

[...]

PR #339, opened 2026-05-14
fix(staging): pre-filter directory snapshots on macOS to keep :U safe (#338)

### LINKED ISSUE #338: :U idmapped mount flag breaks container start on macOS hosts with virtio-fs-hidden AF_UNIX sockets (regression from #333)
## Summary

Follow-up to #333 (closed by 2.28.4). The new `:U` idmapped mount flag for macOS staging volumes triggers a NEW failure mode that prevents the container from starting at all: podman's chown-recursively step (required by `:U`) lstat's every entry in the host source tree and aborts on AF_UNIX socket files that virtio-fs returns from `readdir()` but ENOENT from `stat()` (the same virtio-fs pathology behind #328 / #335).

The host filesystem layout that previously worked with `:ro` only now fails immediately:

```
Error: preparing container <ID> for attach: failed to chown recursively host path:
  lstat /Users/aviad.s/.claude/.git/fsmonitor--daemon.ipc: no such file or directory
```

Container never starts. Exit 127.

## Reproduction (macOS host, Podman 5.x, kapsis 2.28.4)

Any agent launch whose `filesystem.include` covers `~/.claude` (or any dir containing host-side AF_UNIX sockets — `.ssh/agent/`, `.claude/.git/fsmonitor--daemon.ipc`, marketplace `.git/` clones with fsmonitor enabled, etc.) hits this failure.

Concrete probe: real repo (`~/git/products`), bot's trusted config, both overlay AND worktree mode tested:

```bash
kapsis ~/git/products \
  --config ~/.config/kapsis/slack-bot-agent.yaml \
  --spec /path/to/spec.md \
  --agent-id t2284d1007 \
  --branch test/x --base-branch master

# Result (both with and without --branch):
# Error: preparing container 616b23d1a06d... for attach: failed to chown
#   recursively host path: lstat /Users/aviad.s/.claude/.git/fsmonitor--daemon.ipc:
#   no such file or directory
# [kapsis] Failed (exit code: 127, elapsed: 0m 44s)
```

Agent logs: `~/.kapsis/logs/kapsis-t22841002.log` (overlay), `~/.kapsis/logs/kapsis-t2284d1007.log` (worktree). Both fail identically — the error is in podman's container-prepare phase, before the entrypoint runs.

## Root cause

The `:U` flag instructs podman to perform `chown -R <containerUID>:<containerGID>` on the source path (mapped onto an idmapped overlay). To do that, podman walks the source tree, lstat'ing each entry. virtio-fs on macOS exposes the AF_UNIX socket entries via `readdir()` but returns ENOENT from `stat()`/`lstat()` (same kernel issue documented in #328 and #335 reproduction). Podman's lstat fails, the chown step bails, and the whole container-prepare exits non-zero.

Distinct from #328/#335: those manifested INSIDE the container during entrypoint's staging step (cp -rp / fuse-overlayfs failing). This one happens on the HOST in podman's preparation step. The fixes from #332 / #334 / #336 all live in `atomic-copy.sh` / `entrypoint.sh` and run AFTER the container starts — they can't help when the container never starts.

## Suggested fixes

1. **Pre-filter source paths to skip non-stattable entries before applying `:U`.** Walk the source on the host (with the same kernel call podman would use), skip sockets/FIFOs/devices, and pass the filtered tree to podman. Expensive but precise.
2. **Use a different idmap mechanism.** `:U` is one option. Alternatives:
   - `--userns=keep-id:uid=0,gid=0` with explicit mappings, plus `:idmap` instead of `:U`.
   - Stage to a host-side scratch dir (without sockets, controlled by kapsis) and mount that with `:ro`. Loses the lazy CoW benefit but sidesteps the socket issue entirely.
3. **Conditionally drop `:U` on sources known to contain sockets** (specifically `.claude` and `.ssh`) and keep it for sources where the previous EACCES manifested.

## Local workaround (this host)

Inline-patched `/opt/homebrew/Cellar/kapsis/2.28.4/libexec/scripts/launch-agent.sh` line 1470 to drop the `:U` flag:

```diff
 local _staging_mount_opts="ro"
-if is_macos; then
-    _staging_mount_opts="ro,U"
-fi
+# kapsis#333 follow-up workaround — see <this-issue-URL>
+# if is_macos; then
+#     _staging_mount_opts="ro,U"
+# fi
```

Without `:U`, the host UID mapping behavior is back to pre-2.28.4. The in-image fixes from PR #336 (`_atomic_cp_stderr_is_benign_only` classifier + `_atomic_restore_modes` + defensive chmod) handle the mode-0700-dir-staging case at the cp level inside the container.

Lossy parts of this workaround (only matter if your kapsis CLI is upgraded via brew — patch is overwritten):
- macOS hosts with strict mode-0700 dirs may regress to the #328 EACCES cascade
- PR #336's classifier catches `cannot stat ... Operation not supported` and `preserving permissions ...` etc., so most cases should still succeed without `:U`

## Reverting the workaround

After the upstream fix ships:

```bash
brew upgrade kapsis  # blows away the local patch
```

## Consumer status

slack-bot's `elif _default_repo:` workaround in `claude_handler.py` is currently disabled (depends on the inline `:U` patch staying in place across brew restarts). If `brew upgrade kapsis` runs unintended, the `:U` patch will be lost and staging will break again until either the patch is reapplied or this issue lands a real fix.

aviadshiber:
## Summary

Fixes #338 — the `:U` mount flag added in #333 (2.28.4) trips podman's container-prepare step because its recursive chown does `lstat()` on every source entry, and virtio-fs on macOS returns `ENOENT` for AF_UNIX socket entries it lists from `readdir()` (e.g. `.claude/.git/fsmonitor--daemon.ipc`). Container never starts, exit 127. PR #336's in-container classifier can't help — it runs after the container is up.

**Fix approach** (issue's suggested fix #1 / #2(b)): on macOS, when a home-dir mount source is a directory containing AF_UNIX sockets / FIFOs / devices, snapshot it to `$HOME/.kapsis/snapshots/<agent-id>/` with those entries excluded, then mount the snapshot with the existing `:ro,U` flags. The snapshot is socket-free, so the `:U` chown traversal succeeds; the #328 fix (UID remap for mode-0700 dirs) is preserved. Directories without unstattable entries keep the live `:ro,U` mount and the fast fuse-overlayfs CoW path.

**Why this approach for the long term**:
- Removes the bug class — any virtio-fs lstat-quirk on any unusual entry type is pre-empted.
- Preserves the #328 fix instead of trading socket bug for EACCES bug.
- Generalizes the existing `_snapshot_file()` strategy (#164) to directories.
- No new Podman / kernel version dependency.
- Cost-controlled: `find -quit` pre-scan only triggers the slow path when sockets are actually present.
- Defense in depth: PR #336's atomic-copy classifier and `entrypoint.sh:685` chmod stay as the in-container safety net.

## Changes

- **`scripts/launch-agent.sh`** — two new helpers next to `_snapshot_file()`:
  - `_path_has_unstattable_entries()` — `find ... -print -quit` scan (returns on first match).
  - `_snapshot_dir_filtered()` — `cp -Rp` + post-prune of any sockets/FIFOs/devices.
  - `generate_filesystem_includes()` grows an `elif` branch alongside the existing file-snapshot path.
- **`tests/test-snapshot-staging.sh`** — 9 new cases covering FIFO detection (root + nested), clean-tree negative case, missing-path guard, FIFO exclusion from snapshots, mode preservation (0600 files), dry-run passthrough, empty-source handling, byte-identical regular file content. The existing `ro,U` mount-opts tests stay unchanged — mount opts on macOS remain `ro,U`.

## Test plan

- [x] `bash tests/test-snapshot-staging.sh` — 23/23 pass (14 existing + 9 new)
- [x] `bash tests/test-agent-config-mounts.sh` — 10/10 pass (adjacent regression check)
- [x] `bash -n` syntax check on both modified files
- [ ] Manual repro on the affected macOS host (per #338 repro command) — expect: container starts, agent runs, `$HOME/.kapsis/snapshots/<agent-id>/.claude/` exists during run with regular files but no `*.ipc` socket.
- [ ] Regression check on a macOS host that originally manifested #328 — expect: `.claude/settings.local.json` and `.ssh/known_hosts` readable inside the container, no `Permission denied` warnings from `atomic_copy_dir`.

## Notes

- shellcheck wasn't available in this environment; both files pass `bash -n` and match the existing project style.
- Quick test suite (`./tests/run-all-tests.sh --quick`) was killed in this environment after running for 10+ minutes on unrelated tests — recommend running the full suite in CI before merging.

https://claude.ai/code/session_012hmkMKfMNinrzK9D2ngz2E

---
_Generated by [Claude Code](https://claude.ai/code/session_012hmkMKfMNinrzK9D2ngz2E)_

### DIFF
--- a/scripts/launch-agent.sh
+++ b/scripts/launch-agent.sh
@@ -1456,6 +1456,13 @@ _snapshot_dir_filtered() {
     fi
 
     local snapshot_path="${SNAPSHOT_DIR}/${relative_name}"
+
+    # Pre-clean to avoid stale-entry merge on re-runs / agent resume.
+    # cp -Rp src/. dst/ merges files into an existing dst, so entries
+    # deleted from the host between launches would persist in the
+    # snapshot and be visible to the new container.
+    rm -rf "$snapshot_path" 2>/dev/null || true
+
     if ! mkdir -p "$snapshot_path" 2>/dev/null; then
         log_warn "Snapshot dir mkdir failed for ${snapshot_path}, falling back to live mount"
         echo "$host_path"
@@ -1468,13 +1475,31 @@ _snapshot_dir_filtered() {
     cp -Rp "${host_path}/." "${snapshot_path}/" 2>/dev/null || true
     find "$snapshot_path" \( -type s -o -type p -o -type c -o -type b \) -delete 2>/dev/null || true
 
-    # Only treat the snapshot as a fallback miss if the source had
-    # content but the snapshot is empty (cp produced nothing).
-    if [[ -z "$(ls -A "$snapshot_path" 2>/dev/null)" ]] && [[ -n "$(ls -A "$host_path" 2>/dev/null)" ]]; then
-        log_warn "Snapshot of ${host_path} is empty, falling back to live mount"
+    # Harden the snapshot subtree AFTER cp -Rp (which would otherwise
+    # propagate the source dir's mode onto snapshot_path and undo any
+    # earlier chmod). Inner file modes are preserved by `-p` (so
+    # id_rsa stays 0600), but the ancestor SNAPSHOT_DIR is created by
+    # the caller with the default umask (typically 0755). chmod 0700
+    # here keeps the credential contents enumerable only by the
+    # current user even if the parent is loose.
+    chmod 0700 "$snapshot_path" 2>/dev/null || true
+
+    # Validate completeness via regular-file count parity. This subsumes
+    # an "is the snapshot empty?" check and catches:
+    #   - cp entirely failed (src=N, dst=0)
+    #   - cp partial failure / mid-copy ENOSPC (src=N, dst=K, K<N)
+    #   - source contained only sockets (src=0, dst=0 — match, no fallback;
+    #     an empty snapshot is still a valid socket-free :U source)
+    # Mirrors atomic-copy.sh's count-based guard (issues #328, #335).
+    local _src_count _dst_count
+    _src_count=$(find "$host_path" -type f 2>/dev/null | wc -l | tr -d ' \n')
+    _dst_count=$(find "$snapshot_path" -type f 2>/dev/null | wc -l | tr -d ' \n')
+    if [[ "$_src_count" != "$_dst_count" ]]; then
+        log_warn "Snapshot file count mismatch for ${host_path} (src=${_src_count} dst=${_dst_count}), falling back to live mount"
         echo "$host_path"
         return 1
     fi
+
     echo "$snapshot_path"
 }
 
@@ -1525,12 +1550,22 @@ generate_filesystem_includes() {
                 # those entry types, aborting container-prepare. Other
                 # directories keep the live mount + fuse-overlayfs CoW path.
                 local mount_source="$expanded_path"
+                local _dir_snapshot_fallback=0
                 if [[ -f "$expanded_path" ]]; then
                     mount_source=$(_snapshot_file "$expanded_path" "$relative_path")
                     log_debug "Snapshot (file): ${expanded_path} -> ${mount_source}"
                 elif [[ -d "$expanded_path" ]] && is_macos && _path_has_unstattable_entries "$expanded_path"; then
-                    mount_source=$(_snapshot_dir_filtered "$expanded_path" "$relative_path")
-                    log_debug "Snapshot (dir, socket-free): ${expanded_path} -> ${mount_source}"
+                    if mount_source=$(_snapshot_dir_filtered "$expanded_path" "$relative_path"); then
+                        log_debug "Snapshot (dir, socket-free): ${expanded_path} -> ${mount_source}"
+                    else
+                        # Fallback returned live host_path; mounting it with :U
+                        # would re-trip the chown-walk on virtio-fs ENOENT (the
+                        # bug this code prevents). Drop :U for THIS entry; the
+                        # in-container atomic-copy classifier (#336) handles
+                        # the remaining cp-stage socket failures.
+                        _dir_snapshot_fallback=1
+                        log_warn "Directory snapshot for ${expanded_path} fell back to live mount; :U dropped on this entry to avoid #338 chown-walk"
+                    fi
                 fi
 
                 # Mount to staging directory (read-only).
@@ -1546,9 +1581,10 @@ generate_filesystem_includes() {
                 # Issue #338: :U makes podman chown-walk the source via lstat(),
                 # which trips on virtio-fs ENOENT for AF_UNIX sockets. The
                 # directory-snapshot above produces a socket-free source for
-                # affected dirs so :U traversal succeeds.
+                # affected dirs so :U traversal succeeds. If the snapshot
+                # itself fell back to the live path, drop :U for that entry.
                 local _staging_mount_opts="ro"
-                if is_macos; then
+                if is_macos && [[ "$_dir_snapshot_fallback" -eq 0 ]]; then
                     _staging_mount_opts="ro,U"
                 fi
                 VOLUME_MOUNTS+=("-v" "${mount_source}:${staging_path}:${_staging_mount_opts}")
--- a/scripts/launch-agent.sh
+++ b/scripts/launch-agent.sh
@@ -1408,6 +1408,76 @@ _snapshot_file() {
     fi
 }
 
+#-------------------------------------------------------------------------------
+# _path_has_unstattable_entries <host_path>
+#
+# Returns 0 if the subtree under <host_path> contains any AF_UNIX socket,
+# FIFO, or device entry — the entry classes that virtio-fs on macOS
+# returns from readdir() but errors on stat()/lstat() (issue #338).
+# Used by generate_filesystem_includes() to decide whether a directory
+# source needs pre-filter snapshotting before applying the :U mount
+# flag (which makes podman chown-walk the source via lstat).
+#
+# Uses `find ... -print -quit` so it returns after the first match —
+# fast even on large trees like ~/.claude/projects.
+#-------------------------------------------------------------------------------
+_path_has_unstattable_entries() {
+    local p="$1"
+    [[ -d "$p" ]] || return 1
+    [[ -n "$(find "$p" \( -type s -o -type p -o -type c -o -type b \) -print -quit 2>/dev/null)" ]]
+}
+
+#-------------------------------------------------------------------------------
+# _snapshot_dir_filtered <host_path> <relative_name>
+#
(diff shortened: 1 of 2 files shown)


aviadshiber:
## Ensemble Code Review — Majority Vote (3 independent reviewers)

Three independent reviewers were dispatched in parallel against `55e5a8a`. Only findings flagged by **at least 2 of 3** reviewers are reported below; singletons are dropped as noise. Verdicts: **3/3 REQUEST_CHANGES**.

### Overview
Fixes #338. On macOS, the `:U` mount flag (added in #333 for the #328 UID-remap fix) causes podman's container-prepare step to recursively `lstat()` the bind-mount source, and virtio-fs returns ENOENT on AF_UNIX socket entries listed by `readdir()` (e.g. `~/.claude/.git/fsmonitor--daemon.ipc`), aborting container-prepare with exit 127. PR #336's in-container atomic-copy classifier can't help because it runs after the container starts. This PR pre-scans HOME directory mount sources host-side and, when sockets/FIFOs/devices are present, snapshots a filtered copy to `~/.kapsis/snapshots/$AGENT_ID/<rel>/` and mounts that instead. Gated by `is_macos`, only triggers when offenders are detected, otherwise the existing fast `:ro,U` + fuse-overlayfs CoW path is preserved.

### Consensus findings (≥2 reviewers agreed)

| # | Severity | Category | Finding | Votes |
|---|----------|----------|---------|-------|
| 1 | **MAJOR** | TESTS | **Inlined helpers in `tests/test-snapshot-staging.sh:105-140` will silently drift from production.** `_path_has_unstattable_entries` and `_snapshot_dir_filtered` are verbatim copies of the launch-agent.sh functions with no sync guard. A future edit to one will pass green tests while shipping broken code. Mitigation: extract to a sourced library (e.g. `scripts/lib/snapshot-helpers.sh`), or follow the `eval`-via-sed pattern already used in `test-workspace-mount-validation.sh:33` to load the live function. | 3/3 |
| 2 | **MAJOR** | TESTS | **All 9 new tests use `mkfifo` as a stand-in for AF_UNIX sockets; the literal #338 reproducer is never exercised.** The acknowledging comment at `tests/test-snapshot-staging.sh:438-441` is reasonable, but `python3 -c "import socket,sys; socket.socket(socket.AF_UNIX).bind(sys.argv[1])" "$path"` is a one-liner that would directly cover the failing entry class. Test plan items 4 & 5 (manual macOS repro) remain unchecked in the PR description. | 3/3 |
| 3 | **MAJOR** | SECURITY | **Snapshot outer-directory mode is not preserved.** `mkdir -p` at `scripts/launch-agent.sh:1459` (and the `SNAPSHOT_DIR` init) is umask-filtered (typically 0755). For `~/.ssh` (host mode 0700), the snapshot dir becomes group/other-traversable. File modes (0600) are correctly preserved by `cp -Rp`, but the parent-dir traversal weakening is a confidentiality regression vs. the host. The test at line 545-547 explicitly concedes this. Fix: `chmod` the per-source snapshot dir to match `stat`-derived source mode after `cp -Rp`. | 3/3 (R1: MAJOR, R2: MINOR, R3: NIT — voted MAJOR because R1 substantively analyzed the `~/.ssh` impact) |
| 4 | MINOR | CORRECTNESS | **`cp -Rp` exit status swallowed via `\|\| true`; partial copies silently succeed.** `_snapshot_dir_filtered` at `scripts/launch-agent.sh:1468` only detects "src non-empty but dst empty" as failure — a partial copy (e.g. ENOSPC, permission error mid-walk) leaves an incomplete snapshot mounted into the container with no warning. At minimum, capture cp's exit and `log_warn` on non-zero. | 2/3 |
| 5 | MINOR | CORRECTNESS / PORTABILITY | **Symlink handling is under-specified.** `find` is called without `-L`, so symlinks pointing to socket-bearing subtrees aren't detected by the pre-scan, and `-type s/p/c/b` tests the link rather than the target. A symlink-to-socket survives `find -delete` in the snapshot as a dangling link. Practically benign for `:U` chown-walk (lstat operates on the link, not the target), but the assumption deserves an inline comment, and consider `-xtype` or documenting the chosen behavior. | 3/3 (R1, R2, R3 each flagged a symlink-related variant) |

### Positive findings (≥2 reviewers confirmed)

- **POSITIVE / CORRECTNESS:** Targeted, macOS-only branch — `is_macos && _path_has_unstattable_entries` guards prevent any Linux regression or clean-tree macOS regression (`scripts/launch-agent.sh:1531`). (3/3)
- **POSITIVE / CONVENTIONS:** `log_warn` correctly writes to stderr (verified in `scripts/lib/logging.sh`), so the dual `log_warn` + `echo "$host_path"` fallback pattern does **not** contaminate `$(...)` command-substitution capture. No bug there. (3/3)
- **POSITIVE / INTERACTION:** Preserves prior fixes — #333's `:U` flag still applied unconditionally on macOS, #328's UID remap behavior untouched, #336's in-container atomic-copy classifier explicitly retained as defense in depth (acknowledged at lines 1446-1447). Inline comments tying the change to #328/#333/#336 are excellent documentation hygiene. (3/3)
- **POSITIVE / PORTABILITY:** `find -print -quit`, `find -delete`, `find -type s/p/c/b`, `cp -Rp`, and `mkfifo` all confirmed BSD + GNU portable. Conventions (`[[ ]]`, quoting, `set -euo pipefail`, `is_macos` from `compat.sh`, no raw echo for operational output) observed throughout. (2/3 explicit, 1/3 implicit)

### Singletons (excluded from consensus, listed for awareness only)

These were flagged by exactly one reviewer and **are not part of the majority recommendation**, but the author may still find them useful:
- Snapshot directory leak — `~/.kapsis/snapshots/$AGENT_ID` has no cleanup path; pre-existing from #164 but materially worsened by full-tree snapshots.
- `find` stderr suppressed in `_path_has_unstattable_entries` — permission errors mid-walk silently fall through to the bare `:U` path.
- Non-HOME absolute directories with sockets are not handled (the else-branch at ~1563 only snapshots files).
- `find -type c/-type b` scope creep beyond AF_UNIX sockets is a minor over-fix; harmless on `~/.claude` but undocumented.

### Recommendation

**REQUEST_CHANGES** (3/3 verdicts). The fix logic is sound for the targeted bug and the diff is well-scoped, but two MAJOR test gaps (drift risk + missing AF_UNIX coverage) and one MAJOR security gap (snapshot mode not preserved for `~/.ssh`-class sources) warrant changes before merge. Tightening the test plan unchecked items (manual macOS repro + #328 regression check) would also strengthen confidence.

---
*Ensemble review: 3 parallel independent Claude reviewers, strict majority vote (≥2/3 agreement required to surface a finding).*

---

Issue #340, opened 2026-05-14
Conversations bind-mount creates /home/developer/.claude as root mode 1755 on macOS, breaking atomic_copy_dir replacement step


aviadshiber:
## Summary

The conversations bind-mount introduced in kapsis 2.28.0 (PR #322 — `feat(conversations): persist agent JSONL via shared host volume`) causes podman to auto-create `/home/developer/.claude` as **root-owned with mode 1755** (sticky bit) inside the container. This makes `.claude` unwritable to the `developer` user, breaking every downstream attempt to populate it — most visibly `atomic_copy_dir`'s rm-rf-and-mv replacement step in `setup_staged_config_overlays`.

The bug is the **mount-path choice** (`/home/developer/.claude/conversations`), not anything in atomic-copy.sh. The earlier #328 / #335 fixes work around symptoms but can't fix the auto-mkdir behavior of podman.

## Reproduction

Podman + macOS + `--userns=keep-id`. Run any container that bind-mounts SOMETHING under `/home/developer/.claude/X` without pre-existing `.claude`:

```bash
mkdir -p /tmp/test-conv
podman run --rm --userns=keep-id \
    -v "$HOME/.claude:/kapsis-staging/.claude:ro" \
    -v "/tmp/test-conv:/home/developer/.claude/conversations" \
    --entrypoint /bin/bash localhost/kapsis-slack-bot:latest -c '
echo "owner=$(stat -c %U:%G /home/developer/.claude) mode=$(stat -c %a /home/developer/.claude)"
echo "developer writable: $([[ -w /home/developer/.claude ]] && echo YES || echo NO)"
source /opt/kapsis/lib/logging.sh
source /opt/kapsis/lib/atomic-copy.sh
atomic_copy_dir /kapsis-staging/.claude /home/developer/.claude 2>&1 | head -5
echo "atomic_copy_dir rc=$?"
echo "file count in dst: $(find /home/developer/.claude -type f 2>/dev/null | wc -l)"
echo "file count in tmp leftover: $(find /home/developer/.atomic-copy-dir-* -type f 2>/dev/null | wc -l)"
'
```

**Output:**
```
owner=root:root mode=1755
developer writable: NO
rm: cannot remove '/home/developer/.claude/conversations': Permission denied
mv: cannot move '/home/developer/.atomic-copy-dir-aEnXK3' to '/home/developer/.claude/.atomic-copy-dir-aEnXK3': Permission denied
[...]

PR #342, opened 2026-05-14
fix(mount): skip write probe in overlay mode — workspace is read-only by design (#341)

### LINKED ISSUE #341: probe_mount_readiness write probe rejects overlay-mode workspace by design (false-positive mount failure)
## Summary

`probe_mount_readiness` in `scripts/entrypoint.sh` performs a write probe (`mktemp $workspace/.kapsis-probe-XXXXXX && : > probe_file`) unconditionally before the agent exec. In **overlay mode the workspace is read-only by design** (see existing short-circuit at `scripts/lib/inject-status-hooks.sh:325` — "In overlay mode the workspace is read-only by design. Skip workspace-side writes entirely"), so the write probe always fails with `EROFS` and the container exits 4 with `KAPSIS_MOUNT_FAILURE[probe_mount_readiness]: /workspace is read-only or write failed at startup (rc=1, virtio-fs drop)` — a **false-positive mount failure**.

This blocks every overlay-mode agent on macOS hosts using podman 5.x with `--userns=keep-id`, where the `:O,upperdir=,workdir=` overlay mount surfaces as `(ro,relatime,...,upperdir=...,workdir=...)` despite the explicit upper/work dirs.

## Reproduction (macOS, podman 5.8.2, kapsis 2.28.5)

The bug manifests for any overlay-mode launch — directly visible from a fresh `podman machine restart`:

```bash
# 1) Confirm podman mounts /workspace read-only with :O,upperdir,workdir + keep-id:
mkdir -p ~/.ai-sandboxes/probe-{upper,work}
podman run --rm --userns=keep-id \
  -v "$HOME/git/<some-repo>:/workspace:O,upperdir=$HOME/.ai-sandboxes/probe-upper,workdir=$HOME/.ai-sandboxes/probe-work" \
  --entrypoint /bin/bash localhost/kapsis-sandbox:latest -c \
  'mount | grep "/workspace "; touch /workspace/.t 2>&1'
```

Output:
```
overlay (ro,relatime,context=...,lowerdir=...,upperdir=...,workdir=...,redirect_dir=nofollow,uuid=on,userxattr)
touch: cannot touch '/workspace/.t': Read-only file system
```

2) Any kapsis overlay-mode launch then exits with code 4 once it reaches `probe_mount_readiness`:

```
KAPSIS_MOUNT_FAILURE[probe_mount_readiness]: /workspace is read-only or write failed at startup (rc=1, virtio-fs drop)
[kapsis] Failed (exit code: 4)
```

Real agent trace (kapsis 2.28.4, agent `t2284i1557` and `t2284k1828`): `.claude` staging succeeds (`LSP servers injected into settings.local.json`), then `probe_mount_readiness` fires and exits 4 — even though virtio-fs is healthy (stat probe passes, only the write probe fails).

## Root cause

`probe_mount_readiness` was added in PR #280 / #276 to detect macOS virtio-fs drops where `/workspace` appears mounted but writes silently fail. The write probe is the right detection **for worktree mode** (`-v $WORKTREE:/workspace` — read-write bind). In overlay mode (`-v $REPO:/workspace:O,upperdir=,workdir=`), the merged view is intentionally not writable from the agent's perspective: the agent's changes go to the upperdir (host-side), staged configs (`.claude`, `.ssh`, etc.) go to `$HOME` via CoW overlays, and task status goes to `/kapsis-status`. **The agent is never supposed to write to `/workspace` directly in overlay mode** — this is already documented and short-circuited in `inject-status-hooks.sh:321-328`.

So in overlay mode the write probe has no signal — it returns the same failure whether virtio-fs is healthy or degraded. Conversely, the **stat probe** at the top of `probe_mount_readiness` still has signal (virtio-fs drops surface as `ENOENT` or hang), and `liveness-monitor.sh::_mount_check_probe` runs the same mode-aware probe mid-run.

For comparison, `liveness-monitor.sh` already handles this correctly (lines 530-561): it uses `stat` + `ls -A` for both modes and only runs a worktree-specific git sentinel check (`test -d /workspace/.git-safe`) when `KAPSIS_SANDBOX_MODE == worktree`. `probe_mount_readiness` should follow the same pattern.

## Suggested fix

Skip the write probe in overlay mode. Keep the stat probe — it still detects virtio-fs drops. Mirror the existing overlay short-circuit pattern from `inject-status-hooks.sh:325`.

```diff
 probe_mount_readiness() {
     local workspace="${1:-/workspace}"
     ...
     # stat probe (unchanged)
     ...
     if [[ $rc -ne 0 ]]; then
         echo "KAPSIS_MOUNT_FAILURE[probe_mount_readiness]: $workspace inaccessible at startup (stat rc=$rc, virtio-fs drop)" >&2
         return 1
     fi

+    # In overlay mode the workspace is read-only by design (the agent's writes
+    # go to host-side upperdir, $HOME via staged-config CoW, and /kapsis-status
+    # for task progress — never to /workspace). A write probe here would always
+    # fail with EROFS regardless of mount health, masking the stat probe's
+    # signal and producing false-positive mount failures. The mid-run
+    # liveness-monitor uses stat + ls -A (no write probe) for the same reason —
+    # see scripts/lib/liveness-monitor.sh::_mount_check_probe. Mirrors the
+    # existing overlay short-circuit in scripts/lib/inject-status-hooks.sh:325.
+    if [[ "${KAPSIS_SANDBOX_MODE:-}" == "overlay" ]]; then
+        log_debug "Skipping workspace write probe (overlay mode — read-only by design)"
+        return 0
+    fi
+
     # Write probe (worktree mode only) — unchanged
     local probe_file
     probe_file="$(mktemp "$workspace/.kapsis-probe-XXXXXX" 2>/dev/null)" ...
```

## Why not "make overlay /workspace writable instead"

`:O,upperdir=,workdir=` mounting as `ro` on macOS+keep-id+podman 5.x is a separate matter (likely a podman quirk around upperdir ownership across the `keep-id` boundary — explicit `:rw` is rejected with "can't use 'O' with other options"). But the existing kapsis design ALREADY treats overlay `/workspace` as read-only from the agent's perspective — see `inject-status-hooks.sh`, the gist re-homing in `launch-agent.sh:1828`, and the `KAPSIS_SANDBOX_MODE == overlay` short-circuits at `entrypoint.sh:1245` and `:1370`. Changing that design is a much bigger change; the surgical fix is to make `probe_mount_readiness` mode-aware like its mid-run counterpart already is.

## Tests

Add 2 tests, update 1:

- `test_probe_mount_readiness_overlay_skips_write_probe`: `KAPSIS_SANDBOX_MODE=overlay`, workspace dir is `chmod 0555`, expect rc=0 (the fix).
- `test_probe_mount_readiness_overlay_still_detects_missing_workspace`: `KAPSIS_SANDBOX_MODE=overlay`, workspace dir does not exist, expect rc=1 + sentinel (stat probe still fires).
- `test_probe_mount_readiness_readonly` (existing): make worktree-mode semantics explicit by setting `KAPSIS_SANDBOX_MODE=worktree` so behavior is unambiguous when the test file is read in isolation.

## Workaround until fixed

Inline-patch `scripts/entrypoint.sh::probe_mount_readiness()` in the installed kapsis to skip the write probe in overlay mode, **and rebuild the kapsis-sandbox image** (the function runs inside the container). For users who can't rebuild: use worktree mode (`--branch ...`) instead of overlay.

## Consumer status

slack-bot's overlay-mode Q&A tasks fail every launch on macOS 2.28.x. Two prior issues in the same overlay-mode startup chain (#338 fixed by #339, #340 still open) mask this one — runs failed earlier before reaching `probe_mount_readiness`. With #339 merged and #340 inline-patched, the failure surfaces cleanly as this issue.

aviadshiber:
## Summary

`probe_mount_readiness` rejects every overlay-mode workspace on macOS with a false-positive `mount_failure` (exit 4). The probe runs an unconditional write check (`mktemp $workspace/.kapsis-probe-XXXXXX && : > probe_file`), but in overlay mode `/workspace` is read-only by design — the agent's writes go to the host upperdir (via `:O,upperdir=,workdir=`), to `$HOME` for staged-config CoW overlays, and to `/kapsis-status` for task progress. The write probe always fails with EROFS, masks the stat probe above, and produces a `KAPSIS_MOUNT_FAILURE[probe_mount_readiness]: /workspace is read-only or write failed at startup` sentinel even when virtio-fs is perfectly healthy.

Fixes #341.

## Root cause

`probe_mount_readiness` was added in #280 / #276 to detect macOS virtio-fs drops where `/workspace` appears mounted but writes silently fail — a real failure mode for **worktree mode** (`-v $WORKTREE:/workspace` — read-write bind). In **overlay mode** (`-v $REPO:/workspace:O,upperdir=,workdir=`), `/workspace` is intentionally read-only from the agent's perspective, as already encoded in:
- `scripts/lib/inject-status-hooks.sh:321-328` ("In overlay mode the workspace is read-only by design. Skip workspace-side writes entirely")
- `scripts/launch-agent.sh:1828` ("Overlay mode: /workspace is read-only, so re-home the runtime gist file")
- `scripts/entrypoint.sh:1245`, `:1370` (overlay short-circuits for gist writes)

The write probe therefore has no signal in overlay mode — it returns the same failure whether virtio-fs is healthy or degraded.

On the host side, podman 5.x mounts `-v X:/workspace:O,upperdir=Y,workdir=Z` as `(ro,relatime,...,upperdir=...,workdir=...)` under `--userns=keep-id` on macOS — directly observable with `mount | grep workspace` inside any such container.

The mid-run liveness-monitor already gets this right: `scripts/lib/liveness-monitor.sh::_mount_check_probe` (lines 530-561) uses `stat` + `ls -A` for both modes and only adds a worktree-specific git sentinel check (`test -d /workspace/.git-safe`). `probe_mount_readiness` is the outlier.

## The change

Add a mode-aware short-circuit before the write probe in `probe_mount_readiness`. Stat probe still runs for both modes — virtio-fs drops surface as `ENOENT` or hang on `stat` regardless of mode.

```diff
     if [[ $rc -ne 0 ]]; then
         echo "KAPSIS_MOUNT_FAILURE[probe_mount_readiness]: $workspace inaccessible at startup (stat rc=$rc, virtio-fs drop)" >&2
         return 1
     fi

+    # Issue #341: in overlay mode the workspace is read-only by design — the
+    # agent's writes go to the host-side upperdir (via :O,upperdir=,workdir=),
+    # to $HOME for staged-config CoW overlays (.claude, .ssh, …), and to
+    # /kapsis-status for task progress. The agent never writes to /workspace
+    # directly. A write probe here would always fail with EROFS regardless of
+    # mount health, masking the stat probe above and producing false-positive
+    # mount failures (exit 4) for every overlay-mode launch on macOS where
+    # podman 5.x mounts ":O,upperdir=,workdir=" as ro under --userns=keep-id.
+    if [[ "${KAPSIS_SANDBOX_MODE:-}" == "overlay" ]]; then
+        log_debug "Pre-agent mount readiness probe: stat passed; skipping write probe in overlay mode (read-only by design)"
+        return 0
+    fi
+
-    # Write probe: the real failure mode is that bash cannot open files for
+    # Write probe (worktree mode): the real failure mode is that bash cannot
+    # open files for output redirect. Verify that the agent will be able to
+    # create files under the workspace before we exec into it.
```

## Tests

`tests/test-workspace-mount-validation.sh` — **26/26 green** (24 existing + 2 new):

| Test | Setup | Expected |
|------|-------|----------|
| `test_probe_mount_readiness_overlay_skips_write_probe` (new) | overlay mode + `chmod 0555 workspace` | rc=0 (write probe skipped, stat OK) |
| `test_probe_mount_readiness_overlay_still_detects_missing_workspace` (new) | overlay mode + non-existent workspace | rc=1 + `KAPSIS_MOUNT_FAILURE[probe_mount_readiness]:` + `inaccessible at startup` (stat-probe sentinel, not write-probe sentinel) |
| `test_probe_mount_readiness_readonly` (tightened) | `KAPSIS_SANDBOX_MODE=worktree` made explicit + `chmod 0555 workspace` | rc=1 + sentinel (regression guard — same semantics as before, just no longer relies on the env-var default) |

The remaining 23 tests are untouched and still pass.

## Why not "make overlay /workspace writable instead"

`:O,upperdir=,workdir=` mounting as `ro` on macOS+keep-id+podman 5.x is a separate concern (`:rw,O,upperdir=,workdir=` is rejected with "can't use 'O' with other options"), and the existing kapsis design already treats overlay `/workspace` as read-only from the agent's perspective — see references in the root-cause section. Reversing that design is a much bigger change with downstream impact on staged-config CoW, status writes, and the upperdir-collection workflow. The surgical fix is to make `probe_mount_readiness` mode-aware like `_mount_check_probe` already is.

## Risk

- **False negatives in overlay mode?** No new ones — the stat probe still runs and catches the same class of mid-virtio-fs drops it did before (mount disappears entirely → `stat` fails). The write probe had no signal in overlay mode (always EROFS), so removing it removes only noise. Mid-run liveness checks are unchanged.
- **Worktree mode regression?** None — the existing `test_probe_mount_readiness_readonly` still passes (the only behavioral change in that test is making the mode assumption explicit; the chmod-0555 / rc=1 / sentinel assertions are unchanged).
- **Tests rely on `KAPSIS_SANDBOX_MODE` env-var leakage?** No — both new tests save and restore the prior value (or unset it) so test ordering doesn't matter.

## Consumer status

Unblocks slack-bot's overlay-mode Q&A flow on macOS 2.28.x. Two prior issues in the same chain (#338 / #339, #340) were masking this one. With #339 merged and #340 outstanding (separate PR), this is the last blocker for overlay-mode end-to-end.

Closes #341

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -1684,9 +1684,11 @@ validate_workspace_mount() {
 # the agent command. On macOS virtio-fs can degrade during that window too.
 #
 # This probe runs immediately before exec and uses `timeout` so a hung virtio-
-# fs transport cannot freeze the container. A write probe is required (not
-# just stat) because the failure mode in #276 is that bash's `>` redirect open
-# fails even when stat succeeds on the parent dir.
+# fs transport cannot freeze the container. In worktree mode a write probe is
+# required (not just stat) because the failure mode in #276 is that bash's `>`
+# redirect open fails even when stat succeeds on the parent dir. In overlay
+# mode the workspace is read-only by design (Issue #341) so only the stat
+# probe runs — see the overlay short-circuit in the function body below.
 #
 # On failure, emits the same KAPSIS_MOUNT_FAILURE: sentinel to stderr that the
 # liveness monitor uses (scripts/lib/liveness-monitor.sh:613), so the existing
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -1743,9 +1743,27 @@ probe_mount_readiness() {
         return 1
     fi
 
-    # Write probe: the real failure mode is that bash cannot open files for
-    # output redirect. Verify that the agent will be able to create files
-    # under the workspace before we exec into it.
+    # Issue #341: in overlay mode the workspace is read-only by design — the
+    # agent's writes go to the host-side upperdir (via :O,upperdir=,workdir=),
+    # to $HOME for staged-config CoW overlays (.claude, .ssh, …), and to
+    # /kapsis-status for task progress. The agent never writes to /workspace
+    # directly. A write probe here would always fail with EROFS regardless of
+    # mount health, masking the stat probe above and producing false-positive
+    # mount failures (exit 4) for every overlay-mode launch on macOS where
+    # podman 5.x mounts ":O,upperdir=,workdir=" as ro under --userns=keep-id.
+    # The mid-run liveness-monitor follows the same mode-aware pattern — see
+    # scripts/lib/liveness-monitor.sh::_mount_check_probe (uses stat + ls -A
+    # for both modes, only adds a git sentinel check in worktree mode).
+    # Mirrors the existing overlay short-circuit in
+    # scripts/lib/inject-status-hooks.sh:325.
+    if [[ "${KAPSIS_SANDBOX_MODE:-}" == "overlay" ]]; then
+        log_debug "Pre-agent mount readiness probe: stat passed; skipping write probe in overlay mode (read-only by design)"
+        return 0
+    fi
+
+    # Write probe (worktree mode): the real failure mode is that bash cannot
+    # open files for output redirect. Verify that the agent will be able to
+    # create files under the workspace before we exec into it.
     # Use mktemp for an unpredictable filename; fall back to PID-suffix on
     # systems where mktemp is unavailable inside the container (extremely rare).
     local probe_file
--- a/tests/test-workspace-mount-validation.sh
+++ b/tests/test-workspace-mount-validation.sh
@@ -446,8 +446,9 @@ test_probe_mount_readiness_overlay_skips_write_probe() {
     local saved_mode="${KAPSIS_SANDBOX_MODE:-}"
     export KAPSIS_SANDBOX_MODE="overlay"
 
+    local stderr_output
     local exit_code=0
-    probe_mount_readiness "$test_dir/workspace" 2>/dev/null || exit_code=$?
+    stderr_output=$(probe_mount_readiness "$test_dir/workspace" 2>&1 >/dev/null) || exit_code=$?
 
     chmod 0755 "$test_dir/workspace"
     if [[ -n "$saved_mode" ]]; then
@@ -458,6 +459,12 @@ test_probe_mount_readiness_overlay_skips_write_probe() {
 
     assert_equals "0" "$exit_code" \
         "Overlay-mode read-only workspace should pass probe (stat OK, write probe skipped)"
+    # Issue #341 user-facing fix: the original bug was the spurious sentinel
+    # appearing on every overlay-mode launch. Lock that in — a future change
+    # that re-emits the sentinel while still returning 0 would re-introduce
+    # the false-positive mount_failure on the host side.
+    assert_not_contains "$stderr_output" "KAPSIS_MOUNT_FAILURE[probe_mount_readiness]:" \
+        "Overlay-mode probe must NOT emit the false-positive sentinel"
 }
 
 test_probe_mount_readiness_overlay_still_detects_missing_workspace() {
--- a/tests/test-workspace-mount-validation.sh
+++ b/tests/test-workspace-mount-validation.sh
@@ -392,7 +392,7 @@ test_probe_mount_readiness_missing() {
 }
 
 test_probe_mount_readiness_readonly() {
-    log_test "probe_mount_readiness returns 1 + sentinel when workspace is read-only"
+    log_test "probe_mount_readiness returns 1 + sentinel when workspace is read-only (worktree mode)"
     # Skip the write-probe case when running as root — root bypasses mode bits.
     if [[ $(id -u) -eq 0 ]]; then
         log_skip "Skipping read-only probe test — running as root, mode bits bypassed"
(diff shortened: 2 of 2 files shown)


aviadshiber (commented):
## Ensemble review

**Verdict:** LGTM. Surgical mode-aware short-circuit that correctly removes a false-positive `KAPSIS_MOUNT_FAILURE` source on every overlay-mode launch under podman 5.x `:O,upperdir=,workdir=` + `--userns=keep-id` on macOS. Tests cover both the new branch and the existing worktree-mode regression guard. Three non-blocking suggestions; the design choices are justified.

Produced by 4 parallel specialised reviewers (correctness, pattern-consistency, tests, risk / defense-in-depth) against PR head `9a94488`.

### Strengths

- **Placement & semantics are right.** The short-circuit sits after the stat probe and before the write probe (`scripts/entrypoint.sh`). Stat-probe failures still emit the sentinel and `return 1`; the new branch only fires on stat success. Caller at `scripts/entrypoint.sh:1943` (`if ! probe_mount_readiness; then ... exit 4`) treats `return 0` as success — correct signal.
- **Pattern matches the recent cluster.** `[[ "${KAPSIS_SANDBOX_MODE:-}" == "overlay" ]]` aligns with `scripts/lib/inject-status-hooks.sh:325`, `scripts/entrypoint.sh:1245`, `scripts/entrypoint.sh:1370`. The empty-string default is the safer choice for a worktree-focused probe — unset env → write probe still runs, preserving pre-PR behavior. The asymmetry with `scripts/lib/liveness-monitor.sh:550` (`${KAPSIS_SANDBOX_MODE:-overlay}` + positive worktree check) is acceptable because each function's correct default differs.
- **Tests are robust.** The save/restore idiom in all three new/tightened tests handles the originally-unset case explicitly (`if [[ -n "$saved_mode" ]]; then export; else unset; fi`), which is actually more careful than the sibling `_skips_probe_container` idiom. Tightened `_readonly` test preserves chmod-0555 / rc=1 / sentinel assertions verbatim — worktree regression guard intact.
- **Defense-in-depth preserved.** Host pre-launch `probe_virtio_fs_health` (`scripts/lib/podman-health.sh:85`), mid-run `_mount_check_probe` (`scripts/lib/liveness-monitor.sh:526-563`), `/kapsis-status` named volume on macOS, and `vfkit-watchdog.sh` all still surface real mount failures. The startup write probe was the least valuable of these for overlay mode (the agent never writes `/workspace` directly anyway — writes go to host upperdir / `$HOME` / `/kapsis-status`).
- **The deferred alternative is the right call.** `:rw,O,upperdir=,workdir=` is rejected by podman, and overlay-as-read-only is load-bearing across staged-config CoW, status writes, and the upperdir-collection workflow.
- **No security concern.** `KAPSIS_SANDBOX_MODE` is injected by the trusted host (`scripts/launch-agent.sh:1934`); the container cannot forge it.

### Suggestions (non-blocking)

| # | What |
|---|------|
| **S1** | **PR description's `scripts/launch-agent.sh:1828` reference is stale.** Line 1828 is the credential `inject_to_file` branch, not an overlay short-circuit. The actual overlay short-circuit in `launch-agent.sh` is at **lines 1941-1945** (`# Overlay mode: /workspace is read-only, so re-home the runtime gist file…`). Code is fine; just worth updating the PR body for future readers. |
| **S2** | **Function-header docblock at `scripts/entrypoint.sh:1685-1697` is now stale.** It still asserts: *"A write probe is required (not just stat) because the failure mode in #276 is that bash's `>` redirect open fails even when stat succeeds on the parent dir."* After this PR that statement is only true in worktree mode. Recommend a one-line addition to the header (e.g., *"Write probe runs in worktree mode only; overlay mode skips it because `/workspace` is read-only by design — see Issue #341 and the overlay branch below."*) so the next reader doesn't re-introduce an unconditional write. |
| **S3** | **The overlay-skip test discards stderr and never asserts the spurious sentinel is gone.** `test_probe_mount_readiness_overlay_skips_write_probe` uses `2>/dev/null`, so it locks in `rc=0` but not the user-visible fix (no `KAPSIS_MOUNT_FAILURE[probe_mount_readiness]:` sentinel on stderr). The original bug was the *sentinel* showing up on every overlay launch — a future regression could legitimately re-emit it while still returning 0. Suggest capturing stderr like the sibling `_readonly` test (`stderr_output=$(probe_mount_readiness … 2>&1 >/dev/null)`) and adding `assert_not_contains "$stderr_output" "KAPSIS_MOUNT_FAILURE[probe_mount_readiness]:"`. Locks in the actual user-facing fix. |

### Nits (optional)

- No test for `KAPSIS_SANDBOX_MODE` set to an unrecognised value (e.g., `"foo"`). Code treats unknown as worktree-style; a one-line test or a contract comment in `_overlay_skips_write_probe` would document the default.
- Stat-probe-fail coverage in overlay mode only exercises the nonexistent-path branch. Adding a "path exists but is not a directory" case would harden against a future stat-probe refactor.

### Notes

- Other cited cross-references verified accurate: `scripts/lib/inject-status-hooks.sh:321-328`, `scripts/entrypoint.sh:1245`, `scripts/entrypoint.sh:1370`, `scripts/lib/liveness-monitor.sh:530-561`.
- No other unguarded write probes against `/workspace` found across `scripts/` that this PR should also have covered. (`scripts/lib/podman-health.sh:176` writes to a host-side tempdir, not the overlay; `scripts/lib/progress-monitor.sh:188` is a pre-existing latent issue in an unsupported-agent fallback path, not regressed by this PR.)
- Recommend deferring the test-coverage Nits to a follow-up; S1, S2, S3 are all low-effort and worth doing in this PR.

---

Issue #221, opened 2026-03-24
Container starts without /workspace mount (error_during_execution: Path /workspace does not exist)


aviadshiber:
## Bug Description

After upgrading to Kapsis 2.16.3, some agent containers start without the `/workspace` directory being mounted. Claude CLI fails immediately with:

```
Path "/workspace" does not exist
```

The response file contains raw JSON with `"type":"result","subtype":"error_during_execution"` and all token counts at zero — Claude never even started.

## Evidence from agent 471da6

Container exited code 1 after ~57 minutes (likely spent retrying/waiting before giving up):

```
[2026-03-24 14:08:01] Starting container...
[2026-03-24 15:05:28] Timer [container]: 3446s elapsed  
[2026-03-24 15:05:28] Container exited with code: 1
```

Response file contents (truncated):
```json
{
  "type": "result",
  "subtype": "error_during_execution",
  "duration_ms": 0,
  "is_error": true,
  "num_turns": 0,
  "errors": [
    "Path \"/workspace\" does not exist",
    "Error: connect ECONNREFUSED 0.0.0.0:4317"
  ]
}
```

The ECONNREFUSED on port 4317 (OpenTelemetry gRPC) is secondary — the primary issue is `/workspace` not existing.

## Kapsis log

Worktree was created successfully on the host:
```
[14:07:55] Worktree ready: /Users/aviad.s/.kapsis/worktrees/products-471da6
[14:07:56] Sanitized git ready: /Users/aviad.s/.kapsis/sanitized-git/471da6
```

But the container couldn't see `/workspace`. This suggests the volume mount for the worktree failed silently.

## Config

Using `slack-bot-agent.yaml` with:
```yaml
agent:
  type: claude-cli
  command: >-
    ...
  workdir: /workspace
```

## Impact

Agent fails silently, user sees raw JSON error dumped to Slack. No useful work is done.

## Possibly related

This might be related to the sanitized-git objects symlink issue (#219) — both involve path mapping between host and container.

PR #345, opened 2026-05-14
fix(mount): skip write probe in overlay mode — read-only by design (#341)

### LINKED ISSUE #341: probe_mount_readiness write probe rejects overlay-mode workspace by design (false-positive mount failure)

aviadshiber:
## Summary

- Skip the write probe in `probe_mount_readiness` when `KAPSIS_SANDBOX_MODE=overlay` — workspace is intentionally read-only by design, so the probe always failed with EROFS producing false-positive exit-4 mount failures on every overlay-mode launch on macOS (fixes #341)
- Update `probe_mount_readiness` header comment to clarify the write probe is worktree-mode only after #341
- Add two new tests: overlay mode skips write probe (rc=0, no spurious sentinel), overlay mode still detects missing workspace via stat probe (rc=1 + sentinel)
- Tighten existing `test_probe_mount_readiness_readonly` to set `KAPSIS_SANDBOX_MODE=worktree` explicitly

## Note

The changes on this branch are equivalent to those already merged into `main` via #342. File diffs vs `main` are empty — this PR exists for CI tracking purposes and can be closed/merged as a no-op.

## Test plan

- [ ] `./tests/run-all-tests.sh --quick` — no regressions
- [ ] `tests/test-workspace-mount-validation.sh` — 26/26 green (24 existing + 2 new)

https://claude.ai/code/session_01AtipW18AHZ9VkvxLHSegDH

---
_Generated by [Claude Code](https://claude.ai/code/session_01AtipW18AHZ9VkvxLHSegDH)_

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -1684,9 +1684,11 @@ validate_workspace_mount() {
 # the agent command. On macOS virtio-fs can degrade during that window too.
 #
 # This probe runs immediately before exec and uses `timeout` so a hung virtio-
-# fs transport cannot freeze the container. A write probe is required (not
-# just stat) because the failure mode in #276 is that bash's `>` redirect open
-# fails even when stat succeeds on the parent dir.
+# fs transport cannot freeze the container. In worktree mode a write probe is
+# required (not just stat) because the failure mode in #276 is that bash's `>`
+# redirect open fails even when stat succeeds on the parent dir. In overlay
+# mode the workspace is read-only by design (Issue #341) so only the stat
+# probe runs — see the overlay short-circuit in the function body below.
 #
 # On failure, emits the same KAPSIS_MOUNT_FAILURE: sentinel to stderr that the
 # liveness monitor uses (scripts/lib/liveness-monitor.sh:613), so the existing
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -1743,9 +1743,27 @@ probe_mount_readiness() {
         return 1
     fi
 
-    # Write probe: the real failure mode is that bash cannot open files for
-    # output redirect. Verify that the agent will be able to create files
-    # under the workspace before we exec into it.
+    # Issue #341: in overlay mode the workspace is read-only by design — the
+    # agent's writes go to the host-side upperdir (via :O,upperdir=,workdir=),
+    # to $HOME for staged-config CoW overlays (.claude, .ssh, …), and to
+    # /kapsis-status for task progress. The agent never writes to /workspace
+    # directly. A write probe here would always fail with EROFS regardless of
+    # mount health, masking the stat probe above and producing false-positive
+    # mount failures (exit 4) for every overlay-mode launch on macOS where
+    # podman 5.x mounts ":O,upperdir=,workdir=" as ro under --userns=keep-id.
+    # The mid-run liveness-monitor follows the same mode-aware pattern — see
+    # scripts/lib/liveness-monitor.sh::_mount_check_probe (uses stat + ls -A
+    # for both modes, only adds a git sentinel check in worktree mode).
+    # Mirrors the existing overlay short-circuit in
+    # scripts/lib/inject-status-hooks.sh:325.
+    if [[ "${KAPSIS_SANDBOX_MODE:-}" == "overlay" ]]; then
+        log_debug "Pre-agent mount readiness probe: stat passed; skipping write probe in overlay mode (read-only by design)"
+        return 0
+    fi
+
+    # Write probe (worktree mode): the real failure mode is that bash cannot
+    # open files for output redirect. Verify that the agent will be able to
+    # create files under the workspace before we exec into it.
     # Use mktemp for an unpredictable filename; fall back to PID-suffix on
     # systems where mktemp is unavailable inside the container (extremely rare).
     local probe_file
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -1686,9 +1686,13 @@ validate_workspace_mount() {
 # This probe runs immediately before exec and uses `timeout` so a hung virtio-
 # fs transport cannot freeze the container. In worktree mode a write probe is
 # required (not just stat) because the failure mode in #276 is that bash's `>`
-# redirect open fails even when stat succeeds on the parent dir. In overlay
-# mode the workspace is read-only by design (Issue #341) so only the stat
-# probe runs — see the overlay short-circuit in the function body below.
+# redirect open fails even when stat succeeds on the parent dir; /workspace is
+# read-write so the probe has signal. In overlay mode the workspace is
+# intentionally read-only (Issue #341): agent writes target $HOME and
+# /kapsis-status instead, so a write probe on /workspace always returns EROFS
+# regardless of mount health — only the stat probe runs. If KAPSIS_SANDBOX_MODE
+# is unset or empty the write probe runs (worktree behavior), which is the safe
+# default for callers that do not export the variable.
 #
 # On failure, emits the same KAPSIS_MOUNT_FAILURE: sentinel to stderr that the
 # liveness monitor uses (scripts/lib/liveness-monitor.sh:613), so the existing
@@ -1757,8 +1761,16 @@ probe_mount_readiness() {
     # scripts/lib/liveness-monitor.sh::_mount_check_probe (uses stat + ls -A
     # for both modes, only adds a git sentinel check in worktree mode).
     # Mirrors the existing overlay short-circuit in
-    # scripts/lib/inject-status-hooks.sh:325.
-    if [[ "${KAPSIS_SANDBOX_MODE:-}" == "overlay" ]]; then
+    # scripts/lib/inject-status-hooks.sh::inject_gist_instructions.
+    #
+    # Normalise the mode string: lowercase and strip whitespace so "Overlay",
+    # " overlay", and future overlay-family names match. Unset/empty is treated
+    # as worktree (write probe runs) — safe default for callers that do not
+    # export KAPSIS_SANDBOX_MODE.
+    local _probe_mode="${KAPSIS_SANDBOX_MODE:-}"
+    _probe_mode="${_probe_mode,,}"                    # bash lowercase
+    _probe_mode="${_probe_mode//[[:space:]]/}"        # strip whitespace
+    if [[ "$_probe_mode" == "overlay" ]]; then
         log_debug "Pre-agent mount readiness probe: stat passed; skipping write probe in overlay mode (read-only by design)"
         return 0
     fi
--- a/tests/test-workspace-mount-validation.sh
+++ b/tests/test-workspace-mount-validation.sh
@@ -60,6 +60,23 @@ cleanup_test_dirs() {
     _TEST_DIRS=()
 }
 
+# _with_sandbox_mode <mode> <cmd...>: run <cmd> with KAPSIS_SANDBOX_MODE=<mode>,
+# restoring the original value on return. Eliminates repetitive save/export/restore
+# boilerplate in probe_mount_readiness tests.
+_with_sandbox_mode() {
+    local _mode="$1"; shift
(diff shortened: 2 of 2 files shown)


aviadshiber (commented):

**Note on PR state:** the diff vs `main` is empty — these commits already landed via #342. Review was conducted against the commits themselves (`9a94488`, `e64cc46`). Findings below apply to the merged code in `main`, surfaced here for any follow-up work.

**Vote tally:** 27 raw findings → 5 survived majority (≥2/3 reviewers).
**Verdict:** APPROVE (2/3); 1 reviewer voted COMMENT.

---

### Findings (≥2 reviewers agreed)

1. **[LOW] `scripts/entrypoint.sh:1761` — Inconsistent default for unset `KAPSIS_SANDBOX_MODE`** (3/3)
   The new short-circuit uses `${KAPSIS_SANDBOX_MODE:-}` (empty → falls through to write probe, i.e. worktree behavior). But `scripts/lib/liveness-monitor.sh:550`, `scripts/entrypoint.sh:1588,1805,1916`, and `validate_workspace_mount`'s `${1:-overlay}` all default to **`overlay`**. A caller that leaves `KAPSIS_SANDBOX_MODE` unset (e.g. K8s backend or a future entry point) will re-hit the exact #341 false positive that this PR fixes. Reviewer C considered this HIGH; A and B rated LOW because `launch-agent.sh` always exports the variable in current callers. Either way: aligning the default to `overlay` (matching peers) or asserting the variable is set at function entry would harden the fix.

2. **[LOW] `scripts/entrypoint.sh:1761` — Strict equality on the mode string** (3/3)
   The check is exact `[[ "${KAPSIS_SANDBOX_MODE:-}" == "overlay" ]]`. Values like `"Overlay"`, `" overlay"`, or future mode names (e.g. `"overlay-fuse"`, `"overlay-ro"`) silently fall through to the write probe and re-emit the false-positive sentinel. Consider normalizing once (`mode=${KAPSIS_SANDBOX_MODE,,}` + trim) at function entry, or comparing against an explicit allowlist.

3. **[LOW] `tests/test-workspace-mount-validation.sh:370-392` — Earlier tests don't pin the mode** (2/3)
   `test_probe_mount_readiness_healthy` and `test_probe_mount_readiness_missing` were not updated to explicitly export `KAPSIS_SANDBOX_MODE` even though `test_probe_mount_readiness_readonly` was tightened to do so in this PR. Their behavior now depends on prior-test cleanup leaving the variable unset — fragile, and would silently change semantics if the default ever flips to `overlay`. Suggest mirroring the `readonly` test's pattern for consistency.

4. **[NIT] `scripts/entrypoint.sh:1687-1691` — Header comment gaps after the docs update** (3/3, different angles)
   - A: doesn't document what the default is when the mode is unset
   - B: loses the prior emphatic "(not just stat)" rationale for why the worktree write probe exists at all
   - C: doesn't say what protects overlay mode from the EROFS-failure-mode the worktree path guards against (answer: nothing — by design, agent writes target `$HOME`/`/kapsis-status`, not `/workspace`)
   A one-line addition covering "overlay writes go elsewhere, so write-probe coverage is unnecessary there" would close all three.

5. **[NIT] `tests/test-workspace-mount-validation.sh` — Duplicated save/restore boilerplate** (2/3)
   The save → set → run → restore pattern for `KAPSIS_SANDBOX_MODE` is repeated 3–4 times across the new tests. A small `_with_sandbox_mode <mode> <fn>` helper (or a `trap`-based guard) would reduce drift risk in future tests.

---

### Findings raised by only one reviewer (not in majority — recorded for transparency)

- [MEDIUM] Overlay path drops the `ls -A` empty-dir check used by `_mount_check_probe` (claimed parity); could miss the #221 empty-listing virtio-fs drop pre-exec.
- [MEDIUM] `test_probe_mount_readiness_overlay_skips_write_probe` doesn't prove causality (a companion `worktree`-mode run on the same fixture expecting rc=1 would).
- Test path `/nonexistent/workspace-overlay-$$` could collide with stale state; using `$(make_test_dir)/missing-$$` would be safer.
- Root-skip means new overlay tests don't exercise the fix on root-only CI lanes.
- `KAPSIS_AGENT_ID` not explicitly set in the `_overlay_still_detects_missing_workspace` test (could short-circuit at line 1707 before the stat probe).
- Cross-reference to `scripts/lib/inject-status-hooks.sh:325` is by line number, which rots; prefer the function name (`inject_gist_instructions`).
- `assert_not_contains` matches `[probe_mount_readiness]` literally; a future tag rename would break the negative assertion silently.
- The function name `probe_mount_readiness` implies broader coverage than the implementation provides (e.g. doesn't probe `/kapsis-status`).
- `log_debug` message wording / success-log parity nits.
- The 13-line in-body comment partly duplicates the function-header comment.
- Gate placement between stat probe and write probe could be re-broken by a future contributor adding logic in the gap.

---

_This review was generated by an ensemble of 3 independent reviewers; findings above are those raised by a majority. None block merge — the diff vs main is empty._

---

PR #349, opened 2026-05-15
fix(autoheal): sweep wedged zombie containers so virtio-fs auto-heal is autonomous (#348)

### LINKED ISSUE #348: Zombie wedged containers block maybe_autoheal_podman_vm — cascade turns transient virtio-fs degradation into permanent launch-blocked state
## Symptom

Every new agent launch fails with:
```
KAPSIS_MOUNT_FAILURE[probe_virtio_fs_health]: virtio-fs degraded at launch — aborting before container start
```

`maybe_autoheal_podman_vm` refuses to restart the VM because `count_running_kapsis_containers` returns >0. The "running" container is in fact a **wedged zombie** (PID-1 in D-state on the hung virtio-fs mount) that cannot be killed via `podman stop`, `podman rm -f`, or even `podman kill -s KILL`. Only `podman machine stop && podman machine start` reaps it.

## Live reproduction on macOS Apple Hypervisor (kapsis 2.28.4, podman 5.x)

```bash
$ podman ps --filter label=kapsis.managed=true
480ba1031197  kapsis-sandbox:latest  ...  Up About an hour  kapsis-e3f8f4

$ timeout 12 podman stop -t 10 kapsis-e3f8f4 ; echo rc=$?
rc=124   # hangs past the 10s timeout

$ timeout 12 podman rm -f kapsis-e3f8f4 ; echo rc=$?
rc=124   # also hangs

$ podman inspect kapsis-e3f8f4 --format '{{.State.Status}} pid={{.State.Pid}}'
running pid=553226

$ timeout 5 podman exec kapsis-e3f8f4 /bin/true ; echo rc=$?
rc=124   # container shell unresponsive
```

The container is **alive-but-wedged**: `inspect` says running, `exec` times out, `stop` and `rm -f` hang. The Linux PID-1 inside the VM is in uninterruptible D-state on a syscall against the degraded virtio-fs mount — even SIGKILL can't reap a D-state process.

## Root cause cascade

1. Agent runs successfully for hours
2. macOS Apple Hypervisor virtio-fs mount degrades under heavy I/O (well-known platform issue; this report is **not** about fixing that)
3. Container's PID-1 ends up blocked in D-state on a virtio-fs syscall
4. The orchestrating caller (in our case the slack-bot's `stop_container` wrapper around `podman stop` / `podman rm -f`) observes both commands hang, swallows the timeout, and moves on — leaving the container "Up" in podman's view
5. The wedged container still carries `kapsis.managed=true` (set at \`scripts/launch-agent.sh:2333\`)
6. Next launch:
   - `probe_virtio_fs_health` correctly fires (rc=124, virtio-fs degraded)
   - `maybe_autoheal_podman_vm` (\`scripts/lib/podman-health.sh:323-332\`) checks `count_running_kapsis_containers`
   - `count_running_kapsis_containers` (\`scripts/lib/podman-health.sh:231-245\`) does `podman ps --filter 'label=kapsis.managed=true'` with **no responsiveness probe** → returns 1 (counting the zombie)
   - Auto-heal refuses to restart the VM "to protect running agents" → exits with `KAPSIS_MOUNT_FAILURE`
7. `_kill_vfkit_zombie` (\`scripts/lib/compat.sh:310-403\`) exists but is **only reachable from inside `_podman_machine_restart`** — which is precisely the path the count check just refused to take. **Catch-22.**

A transient macOS hypervisor degradation is thus converted into a **permanently launch-blocked state** until the operator manually runs `podman machine stop && podman machine start`.

## Affected files

- \`scripts/lib/podman-health.sh:231-245\` — `count_running_kapsis_containers` (no responsiveness probe)
- \`scripts/lib/podman-health.sh:294-358\` — `maybe_autoheal_podman_vm` (refuses on any non-zero count, no escape hatch)
- \`scripts/lib/compat.sh:310-403\` — `_kill_vfkit_zombie` (unreachable when zombie is blocking count)
- \`scripts/lib/kapsis-cleanup.sh:597-622\` — cleanup only acts on **stopped** containers; no zombie reaper for `Up`-but-wedged ones
- \`scripts/launch-agent.sh:2805\` — call site

## Proposed fix (Kapsis side)

Two complementary changes, either of which is sufficient on its own but both together close the loop:

**A. Responsiveness probe in `count_running_kapsis_containers`** (\`podman-health.sh:231-245\`):

```bash
count_running_kapsis_containers() {
    local ids
    ids=$(podman ps --filter "label=kapsis.managed=true" --format '{{.ID}}')
    local responsive=0
    for id in $ids; do
        if timeout 2 podman exec "$id" /bin/true >/dev/null 2>&1; then
            responsive=$((responsive + 1))
        else
            log_warn "Container $id is unresponsive (zombie) — excluding from running count"
        fi
    done
    echo "$responsive"
}
```

Wedged containers are excluded from the count. They hold no salvageable work anyway (the agent inside them is dead).

**B. Pre-restart zombie reaper** in `maybe_autoheal_podman_vm`:

Before the count check refuses, sweep wedged containers:

```bash
maybe_autoheal_podman_vm() {
    if ! probe_virtio_fs_health; then
        # Sweep wedged kapsis containers BEFORE checking count
        for id in $(podman ps --filter "label=kapsis.managed=true" --format '{{.ID}}'); do
            if ! timeout 2 podman exec "$id" /bin/true >/dev/null 2>&1; then
                log_warn "Force-removing wedged container $id (D-state PID-1, unresponsive)"
                timeout 5 podman rm -f "$id" 2>/dev/null || true
                # If it survives, VM restart will reap it
            fi
        done

        local running
        running=$(count_running_kapsis_containers)
        if [[ $running -eq 0 ]]; then
            _podman_machine_restart    # which calls _kill_vfkit_zombie
            return 0
        fi
        # ... existing refuse path
    fi
}
```

Combined with (A), this means wedged containers get one chance at podman-level removal, then if they survive they're excluded from the count, then the VM restart kills them at the hypervisor layer (`_kill_vfkit_zombie`).

## Why this matters

This makes the auto-heal **autonomous**. Right now the system requires a human (me, today) to:
1. Notice the bot hasn't responded for hours
2. Recognize the symptom (\`KAPSIS_MOUNT_FAILURE\` repeating)
3. Identify which container is the zombie
4. Manually `podman rm -f` (which will hang)
5. Manually `podman machine stop && start`

With the fix, the next launch attempt observes virtio-fs degraded, identifies and reaps the zombie, restarts the VM, and proceeds — no human in the loop.

## Reference: caller-side context

For completeness — the slack-bot wrapper that produces the zombie is `stop_container` in `~/.claude/skills/slack-cli/scripts/handlers/utils.py:240-267`. The bot's caller swallows the `TimeoutExpired` at DEBUG level. We'll fix the bot side too, but the more durable fix lives here in Kapsis because any caller of `podman stop` against a wedged container hits the same problem.

## Related

- Project memory `project_kapsis_virtio_fs_issues.md` — three issues: virtio-fs mount drop, corrupted store surviving cleanup, bot not auto-recovering. This issue concerns problem #3 (auto-recovery).
- \`probe_virtio_fs_health\` (the mitigation that detects the trigger) works correctly. It's the auto-heal *response* that's broken.

aviadshiber:
Closes #348.

## Problem

On macOS Apple Hypervisor, virtio-fs occasionally degrades under heavy I/O and pins the container's PID-1 in uninterruptible D-state on a virtio-fs syscall. `podman ps` shows the container "Up", but `podman stop`, `podman rm -f`, and `podman exec` all hang — and even SIGKILL cannot reap a D-state process. Only `podman machine stop && podman machine start` (which `_kill_vfkit_zombie` does by killing the hypervisor at the host layer) clears the zombie.

The pre-launch auto-heal already restarts the Podman VM when it detects virtio-fs degradation, but it short-circuited: `count_running_kapsis_containers` counted the zombie as "running", and `maybe_autoheal_podman_vm` then refused to restart "to protect running agents". A transient hypervisor hiccup turned into a **permanently launch-blocked** state until an operator manually ran `podman machine stop && start`. Catch-22: `_kill_vfkit_zombie` already exists, but it sits on the path the count check just refused to take.

## Fix

Two complementary changes in `scripts/lib/podman-health.sh`:

**(A) Responsiveness probe in `count_running_kapsis_containers`** — each kapsis-labelled container is probed via `timeout N podman exec <id> /bin/true`. Wedged containers (whose agent is already dead) are excluded from the count. A new shared helper `_probe_container_responsive` powers the probe and **fails closed** when no `timeout` / `gtimeout` binary is available, so we never hang the caller while probing a wedged container.

**(B) Pre-restart zombie reaper in `maybe_autoheal_podman_vm`** — a new `_sweep_wedged_kapsis_containers` runs before the count check. It attempts a bounded `podman rm -f` on each wedged container — best-effort, may hang. If `rm -f` fails, (A) still excludes those containers from the count, the restart proceeds, and `_kill_vfkit_zombie` reaps them at the hypervisor layer.

New constant `KAPSIS_DEFAULT_VFS_EXEC_PROBE_TIMEOUT=2` in `scripts/lib/constants.sh`; overridable via `KAPSIS_VFS_EXEC_PROBE_TIMEOUT`.

## Test plan

- [x] `bash tests/test-podman-health.sh` — all 23 tests pass (17 pre-existing + 6 new for #348)
- [x] `shellcheck --severity=warning scripts/lib/podman-health.sh scripts/lib/constants.sh tests/test-podman-health.sh` — clean
- [x] `bash -n` syntax check — clean

### New tests

| Test | What it proves |
|------|----------------|
| `test_count_excludes_wedged_containers` | Wedged IDs are dropped from the count |
| `test_count_excludes_all_when_no_timeout_binary` | Fail-closed: count returns 0 and `exec` is never invoked when no `timeout` binary |
| `test_autoheal_recovers_when_only_wedged_containers_present` | **Headline #348 regression** — VM restart proceeds when the only "running" container is wedged |
| `test_autoheal_sweep_attempts_rm_for_wedged` | Sweep invokes `podman rm -f` on wedged containers |
| `test_autoheal_sweep_tolerates_rm_failure` | Heal proceeds even when `rm -f` exits non-zero (the realistic case for truly wedged containers) |
| `test_autoheal_still_refuses_when_responsive_containers_running` | Regression: live agents still block restart — wedged-only is the new escape hatch, not a blanket bypass |

### Manual macOS verification

1. Reproduce a wedged kapsis container (`podman stop` hangs, `podman exec` times out, `inspect` says running).
2. `KAPSIS_DEBUG=1 ./scripts/launch-agent.sh ...`.
3. Expected log lines:
   - `Wedged container … detected — attempting podman rm -f (best-effort, may hang)`
   - `Excluding wedged container … from running count`
   - `Restarting Podman VM …`
   - `Virtio-fs recovered after VM restart`
4. `podman ps -a` shows the zombie gone.

## Files

- `scripts/lib/podman-health.sh` — new `_probe_container_responsive`, new `_sweep_wedged_kapsis_containers`, updated `count_running_kapsis_containers`, updated `maybe_autoheal_podman_vm`
- `scripts/lib/constants.sh` — new `KAPSIS_DEFAULT_VFS_EXEC_PROBE_TIMEOUT`
- `tests/test-podman-health.sh` — 6 new tests + 2 fakes updated to handle `exec`

https://claude.ai/code/session_017iw3NujfGnrfumNnWZ1BUD

---
_Generated by [Claude Code](https://claude.ai/code/session_017iw3NujfGnrfumNnWZ1BUD)_


aviadshiber (commented):
## Ensemble review (majority vote across 5 independent reviewers)

**Verdict tally:** 3 APPROVE · 1 COMMENT · 1 REQUEST_CHANGES → **APPROVE with comments** (posting as COMMENT because the bot account authored the PR and cannot self-approve).

The fix is well-scoped and correctly addresses #348: wedged zombies no longer block the auto-heal path. Label-based filtering matches `launch-agent.sh`, fail-closed behaviour when no `timeout` binary is on PATH is the right choice given the macOS-only triggering condition, and the 6 new tests genuinely verify the new behaviour (each would fail under regression). Conventions are clean (`[[ ]]`, `((count++)) || true`, `log_*` helpers, quoting).

Below are findings ordered by how many of the 5 reviewers raised them. Only the top two cleared the majority threshold; the rest are documented for the author's judgement.

---

### Majority findings (≥3/5)

#### 1. Double `podman ps` + double per-container exec probe between sweep and count — 4/5 reviewers
- File: `scripts/lib/podman-health.sh` (`maybe_autoheal_podman_vm` → `_sweep_wedged_kapsis_containers` then `count_running_kapsis_containers`)
- `_sweep_wedged_kapsis_containers` runs `podman ps` and probes every container, then `count_running_kapsis_containers` does the same work again. On a degraded VM where every probe hits the bounded timeout, total cost is ~`2 · N · probe_timeout` (+ up to 5s per `rm -f`). The two `ps` calls also see independent snapshots — a container started between them could be `rm -f`'d as wedged and then missing from the count.
- Suggested fix: have `_sweep_wedged_kapsis_containers` emit the set of confirmed-responsive (or confirmed-wedged) IDs on stdout, and reuse that list in the count instead of re-probing. Cuts probe latency roughly in half and eliminates the snapshot drift.

#### 2. Hardcoded `5`-second `rm -f` timeout breaks the constants pattern — 3/5 reviewers
- File: `scripts/lib/podman-health.sh:360` (the literal `5` in `"$timeout_cmd" 5 "$podman_bin" rm -f "$id"`)
- This PR actively expands the `KAPSIS_DEFAULT_VFS_*` family in `constants.sh` (adding `KAPSIS_DEFAULT_VFS_EXEC_PROBE_TIMEOUT`), so leaving the sweep's `rm -f` timeout as a magic literal is inconsistent with the precedent the PR itself sets. CLAUDE.md "Things to Avoid" #1 calls this pattern out.
- Suggested fix:
  ```bash
  # constants.sh
  readonly KAPSIS_DEFAULT_VFS_RM_FORCE_TIMEOUT=5

  # podman-health.sh
  local rm_timeout="${KAPSIS_VFS_RM_FORCE_TIMEOUT:-${KAPSIS_DEFAULT_VFS_RM_FORCE_TIMEOUT:-5}}"
  "$timeout_cmd" "$rm_timeout" "$podman_bin" rm -f "$id" &>/dev/null || true
  ```

---

### Minority findings (2/5 — worth considering, not blocking)

- **Misleading `log_warn` for recently-exited containers** (R2, R4): if a container exits between `podman ps` and the probe's `podman exec`, the latter returns non-zero and the warning calls it "wedged". Distinguish `timeout` exit `124` (truly wedged) from other non-zero (benign race) and log the latter at `debug`.
- **`/bin/true` assumption** (R3, R4): holds for Debian-based kapsis images but would falsely classify a distroless or `scratch`-based `kapsis.managed=true` image as wedged. Low risk today; consider documenting the implicit image contract.
- **Timeout-cmd cache nuances** (R1, R3): `_KAPSIS_TIMEOUT_CMD` is `export`ed (leaks into child `podman` processes) and the `[[ -n ... ]]` guard re-scans when the cached value is the empty string (no timeout binary). Drop the `export` and use `[[ "${_KAPSIS_TIMEOUT_CMD+set}" == "set" ]]` for a sentinel-aware check.
- **Test brittleness around the no-timeout-binary path** (R1, R4): the `_vfs_timeout_cmd` override only works because nothing source-time invokes it; an explicit `unset _KAPSIS_TIMEOUT_CMD` in the `bash -c` body would harden the test against future regressions.

### Singleton findings (1/5 — listed for completeness)

- 2s default probe timeout may false-positive on a cold VM and trigger destructive `rm -f` against a healthy slow container (R4)
- Test fakes return `exit 1` instantly rather than modelling a real D-state `timeout`/exit-124 hang (R2)
- `_kill_vfkit_zombie` fallback stub doesn't anchor/validate `$1` (R2)
- Sweep does only one probe before `rm -f` — no re-probe for transient slowness (R4)

---

None of these are merge-blockers, and the majority of reviewers approve. The two majority findings (shared ps/probe state, and promoting `5` to a named constant) are worth a follow-up commit before merge if convenient.

---

PR #352, opened 2026-05-17
fix(hooks): write to settings.json not settings.local.json at user scope (#351)

### LINKED ISSUE #351: auto-gist + status hooks never fire: injected into ~/.claude/settings.local.json which Claude Code does not load at user scope
## Summary

Kapsis injects its PostToolUse / Stop / gist hooks into `~/.claude/settings.local.json` inside the container, but **Claude Code does not load `settings.local.json` at the user scope** — it's a project-only file. As a result, none of Kapsis's injected hooks ever fire in headless (`claude -p`) sessions, and the auto-gist feature is silently broken: `/workspace/.kapsis/gist.txt` is never written, and every `kapsis-<project>-<id>.json` status file has `"gist": null` (verified across 192 historical files on my machine — 0 with a non-null gist).

This affects every slack-bot agent and likely every interactive run too.

## Environment

- Kapsis 2.28.7 (Homebrew, macOS)
- Apple Hypervisor (vfkit) Podman machine
- Agent type: `claude-cli`, headless mode (`claude --dangerously-skip-permissions -p ...`)
- Config has `inject_gist: true`, `gist_llm: true`

## Root Cause (verified inside the container)

`scripts/lib/inject-status-hooks.sh` writes to `${HOME}/.claude/settings.local.json`:

```bash
# inject-status-hooks.sh:55
local settings_local="${settings_dir}/settings.local.json"
```

Result inside the container (`kapsis-9d0960`):

```
$ ls -la /home/developer/.claude/settings*.json
-rw-------. 1 developer developer 1788 May 16 19:08 /home/developer/.claude/settings.local.json

$ cat /home/developer/.claude/settings.local.json
{
  "hooks": {
    "PostToolUse": [
      { "matcher": "*", "hooks": [{ "type": "command", "command": "/opt/kapsis/hooks/kapsis-gist-hook.sh", "timeout": 10 }] },
      { "matcher": "*", "hooks": [{ "type": "command", "command": "/opt/kapsis/hooks/kapsis-status-hook.sh", "timeout": 5 }] }
    ],
    "Stop": [
      { "hooks": [{ "type": "command", "command": "/opt/kapsis/hooks/kapsis-stop-hook.sh", "timeout": 5 }] }
    ]
  },
  ...
}
```

But Claude Code's own debug log (`~/.claude/debug/latest`) shows it never picks the file up:

```
2026-05-16T19:08:50.235Z [DEBUG] Broken symlink or missing file encountered for settings.json at path: /workspace/.claude/settings.local.json
2026-05-16T19:08:50.277Z [DEBUG] Registered 0 hooks from 0 plugins
2026-05-16T19:08:50.351Z [DEBUG] Watching for changes in setting files
    /home/developer/.claude/settings.json,
    /workspace/.claude/settings.json,
    /workspace/.claude/settings.local.json
...
2026-05-16T19:09:04.260Z [DEBUG] Hooks: Found 0 total hooks in registry
2026-05-16T19:09:15.830Z [DEBUG] Hooks: Found 0 total hooks in registry
(... repeats forever — Kapsis hooks never registered)
```

Notice the watched-paths list:

| Scope             | Loaded paths                                                          |
|-------------------|-----------------------------------------------------------------------|
| User (`$HOME`)    | `settings.json` only — **no `.local.json` variant exists at this scope** |
| Project (workspace) | `settings.json` AND `settings.local.json` |

This matches the official Claude Code [settings docs](https://docs.anthropic.com/en/docs/claude-code/settings): `settings.local.json` is a project-only file (it's the gitignored sibling of `settings.json`). At the user scope there is only `~/.claude/settings.json`.

## Reproduction

```bash
# Any kapsis launch with inject_gist:true, then:
podman exec <container-name> bash -c \
  'grep -E "Registered .* hooks|total hooks in registry" /home/developer/.claude/debug/latest | head -3'

# Expected: "Registered N hooks from M plugins" with N > 0
# Actual:   "Registered 0 hooks from 0 plugins"
#           "Hooks: Found 0 total hooks in registry"
```

Confirm none of the kapsis hooks fire by checking the status JSON over time:

```bash
# This never updates beyond the initial heartbeat fields:
jq '{phase, progress, message, gist}' ~/.kapsis/status/kapsis-*-<agent_id>.json
# Always shows gist: null no matter how many Edit/Write/Bash calls the agent makes
```

## Why this slipped through

1. `kapsis-status` shows phase/progress/heartbeat which are written by the host-side liveness monitor, not by the in-container status hook. So the bot/CLI looks "alive" even though the in-container hook chain is dead.
2. Most tests for `inject-status-hooks.sh` verify the file was *written* — not that Claude *loaded* it.
3. The host CLI never reads `gist.txt` directly — only the status JSON, which is null whenever the hook chain is broken.

## Proposed Fix

Change `scripts/lib/inject-status-hooks.sh` to write to `~/.claude/settings.json` (the canonical user-scope file Claude actually loads) and merge with any existing content, instead of `settings.local.json`:

```diff
- local settings_local="${settings_dir}/settings.local.json"
+ local settings_local="${settings_dir}/settings.json"
```

The existing `jq` merge already preserves prior content, so this is safe: if the user has their own `~/.claude/settings.json`, the kapsis hooks get appended alongside their hooks. The `chmod 600` line is fine for `settings.json` too.

The same change is needed in any other code path that points at `settings.local.json` for the user-scope file (e.g. `inject-plugin-hooks.sh`, `filter-agent-config.sh` comment).

### Alternative considered

Writing into the project-local `/workspace/.claude/settings.local.json` instead. **Rejected** because in worktree mode `/workspace` is the user's actual git tree — we'd pollute their repo and could break across re-runs. User-scope `settings.json` is the right home for container-injected hooks.

## Acceptance Criteria

- After this fix, `grep "Registered .* hooks" ~/.claude/debug/latest` inside the container shows `Registered ≥ 3 hooks` for any kapsis launch with `inject_gist: true`.
- `/workspace/.kapsis/gist.txt` exists and updates after the first Edit/Write/Bash tool call.
- `kapsis-status` shows the gist within ~5–10 seconds of the agent's first tool call.
- New unit test in `tests/unit/test_inject_status_hooks.bats` asserts the destination filename is `settings.json` (not `settings.local.json`).

## Blast radius

The fix only touches the user-scope settings file inside the container. The container is ephemeral (rebuilt per agent), so users' host configs are unaffected.

aviadshiber:
Closes #351

## Problem

Claude Code only loads `settings.local.json` at **project scope** (`/workspace/.claude/settings.local.json`). At **user scope** (`~/.claude/`), it only reads `settings.json`. All three Kapsis injection scripts were writing to `~/.claude/settings.local.json`, so every injected hook — PostToolUse status, Stop, gist, LSP servers, plugin hooks — silently never fired.

From Claude Code's own debug log inside a kapsis container:
```
[DEBUG] Watching for changes in setting files
    /home/developer/.claude/settings.json,
    /workspace/.claude/settings.json,
    /workspace/.claude/settings.local.json
[DEBUG] Hooks: Found 0 total hooks in registry
(... repeats forever)
```

The `settings.local.json` in `~/.claude/` was never in the watched list. Result: `gist: null` across 192 historical status files; auto-gist completely broken; status phase/message never updated by in-container hooks.

### Source files

| File | Change |
|------|--------|
| `scripts/lib/inject-status-hooks.sh` | Write to `settings.json` (was `settings.local.json`) |
| `scripts/lib/inject-plugin-hooks.sh` | Write to `settings.json`; remove redundant `user_settings` alias |
| `scripts/lib/inject-lsp-config.sh` | Write to `settings.json` |
| `scripts/lib/agent-types.sh` | Return `settings.json` from `get_agent_hook_config_path()` |
| `scripts/hooks/agent-adapters/claude-adapter.sh` | Return `settings.json` from `get_claude_config_path()` |

### Tests (10 files, 121+ tests — all green)

- All path assertions updated from `settings.local.json` → `settings.json`
- `seed_kapsis_settings_local()` updated to `jq`-merge into existing content
- Regression guard: `settings.local.json` must NOT be created at user scope

### Integration tests — mock API server (new)

**`tests/lib/mock-api-server.py`** — minimal Python Anthropic API mock (stdlib only):
- Returns a Bash `tool_use` on the first request → triggers PostToolUse hooks
- Returns `end_turn` on the follow-up (after `tool_result`) → triggers Stop hooks
- Prints `READY:{port}` on startup; port is self-selected to avoid TOCTOU races
- No real API credentials required

**`tests/test-integration-claude-hooks.sh`** (3 tests against mock API):
1. **PostToolUse gist hook fires** — Claude Code reads hooks from `settings.json` and dispatches the gist hook when it executes a tool
2. **Stop hook fires** — Claude Code dispatches the Stop hook when the session ends (`end_turn`)
3. **PostToolUse receives tool event JSON** — verifies the JSON payload (`tool_name`, `tool_input`) passed to hooks matches the format Kapsis hooks expect

Skips gracefully when `claude` CLI or `python3` are absent.

### E2E tests — live API (new, gated on secret)

**`tests/test-e2e-claude-hooks.sh`** (3 tests using `CLAUDE_CODE_OAUTH_TOKEN`):
- Same 3 scenarios as integration tests but against the real Anthropic API
- Skips when `CLAUDE_CODE_OAUTH_TOKEN` is not set

### CI restructure

```
unit-tests ─┐
             ├──→ e2e-tests (optional, gated on CLAUDE_CODE_OAUTH_TOKEN secret)
integration-tests ─┘
```

- `unit-tests` and `integration-tests` run **in parallel** (no dependency between them)
- `e2e-tests` runs **only after both pass** (`needs: [unit-tests, integration-tests]`)
- `e2e-tests` is **optional** in the `ci-success` gate (won't block merge if secret isn't configured)
- Deleted `e2e-real.yml` (folded into `ci.yml`'s `e2e-tests` job)

## Acceptance criteria (from #351)

- [x] Hooks land in `settings.json`, never in `settings.local.json` (verified by tests)
- [x] `gist.txt` is written when the PostToolUse hook fires (integration test #1)
- [x] Stop hook fires when the session ends (integration test #2)
- [x] Hook payload has correct JSON structure (`tool_name`, `tool_input`) (integration test #3)
- [x] All 121+ affected tests pass

https://claude.ai/code/session_01GN4hKrnktWtNusMBqeSu2U


aviadshiber (commented):

**Verdict: APPROVE** (2 APPROVE / 1 REQUEST_CHANGES) — fix is correct, scoped, and well-tested. Three findings corroborated by ≥2 reviewers; non-blocking.

The rename from `settings.local.json` → `settings.json` correctly fixes #351. Claude Code only watches `settings.json` at user scope (confirmed in the debug log), so the previous writes were a silent no-op. All five source call sites are updated, the jq merges preserve user content (`permissions`, `model`, `env`, `mcpServers`, `enabledPlugins`, `lspServers`, `attribution`), the entrypoint ordering (filter → status → LSP → plugin) reads before mutating, and 113 tests pass. The container security property is unchanged: `~/.claude/settings.json` is CoW-staged from the host but writes go to the upper layer — the host file is never modified and the file is not exported back.

### Corroborated findings (≥2 of 3 reviewers)

**1. `user_settings` is now a dead alias of `settings_local`** (3/3, minor) — `scripts/lib/inject-plugin-hooks.sh:107-109`

Both variables now expand to `${settings_dir}/settings.json`. The symlink-check loop drops `user_settings` because it's redundant *today*, but `user_settings` is still *read* at line 154–156. If a future refactor points `user_settings` back at a different path, the symlink check on a file the script reads will be silently missing.

Suggestion: delete `user_settings` and read `enabledPlugins` from `settings_local` directly, so the invariant survives future edits.

**2. Stale `settings.local.json` references in non-test code and docs** (3/3, severity ranges minor → major)

The fix touches code but leaves ~17 references in comments, log strings, configs, and user-facing docs:

- `docs/PLUGINS.md`, `docs/STATUS-TRACKING.md`, `docs/CONFIG-REFERENCE.md`, `docs/GIT-WORKFLOW.md` — these are the highest priority; the very docs operators read are the path that misled the original implementation.
- `scripts/launch-agent.sh:772, 2008`
- `scripts/entrypoint.sh:682, 1264, 1300, 1832, 1849`
- `scripts/lib/inject-plugin-hooks.sh:98, 269, 274`
- `configs/claude.yaml:118`
- `CLAUDE.md:245`

Suggestion: `grep -rl 'settings\.local\.json' scripts/ configs/ docs/ CLAUDE.md` and sweep. User-facing docs first.

**3. `chmod 600` happens after the jq merge, not after the `echo '{}'` bootstrap** (2/3, minor) — `scripts/lib/inject-status-hooks.sh:62`, `scripts/lib/inject-lsp-config.sh:82`, `scripts/lib/inject-plugin-hooks.sh:139`

When the file is created with `echo '{}' > "$settings_local"`, it picks up the entrypoint umask (typically 644) and is only narrowed to 600 after a successful merge. If jq fails on the first run, the file persists at 644 for subsequent readers in the container. Container is single-user so impact is low, but it's a cheap defense-in-depth fix.

Suggestion: `chmod 600 "$settings_local"` immediately after the `echo '{}' >` bootstrap.

### Minority concerns (1 of 3, not blocking)

The test-coverage reviewer flagged three additional items worth surfacing for the author's judgment:

- **No regression test pins the filename invariant.** All updated tests would pass equally if the path were renamed to a value Claude Code doesn't read. A one-line `assert get_claude_config_path` ends in `/settings.json` (not `settings.local.json`) would prevent re-introduction of the bug.
- **`test_malformed_settings_local_handled` weakens to "no crash".** The original asserted a WARN naming the affected plugin. Now a corrupt `settings.json` is silently "no plugins enabled" at INFO inside this function (the WARN actually fires from `inject_claude_hooks` earlier, but no test asserts that chain).
- **No integration test exercises the full inject chain inside a container.** The original bug — Claude Code silently not loading the file — would not be caught by any current test.

### What's good

- Scope is tight: 9 files, 131 / 119 LOC.
- Test updates are mechanical and correct; non-trivial rewrites (`test_settings_local_absent_at_start`, `test_malformed_settings_local_handled`, `test_symlink_settings_local_rejected`) are well-justified in the PR body.
- `seed_kapsis_settings_local` correctly switches from overwrite to jq-merge so it doesn't clobber `enabledPlugins` written by `set_enabled`.
- Symlink fail-closed behavior, `enabledPlugins` dedup, and idempotency assertions all preserved.
- Pre-existing brainstorm doc in the PR body shows architect/contrarian/security checks were run before implementation.

---
_Ensemble review: 3 independent reviewers (correctness, security, test coverage). Findings listed above appear in ≥2 of 3 reports; minority items called out separately._

---

PR #353, opened 2026-05-17
ci(integration-tests): add timeout-minutes to prevent hung jobs

aviadshiber:
$(cat <<'EOF'
## Summary

- Adds `timeout-minutes: 15` at the job level for `integration-tests`
- Adds `timeout-minutes: 10` at the step level for the "Run integration tests" step

## Problem

The `integration-tests` CI job (introduced in #352) has no timeout. If Claude Code hangs connecting to the mock API server (e.g., attempts to reach `api.anthropic.com` before honouring `ANTHROPIC_BASE_URL`), the job blocks CI for the default 6-hour GitHub Actions limit rather than failing fast with a clear timeout error.

## Fix

Two timeouts in `ci.yml`:
- Job-level: 15 minutes (hard cap — GitHub cancels the whole job)
- Step-level: 10 minutes (cancels just the test step, leaving room for any post-step cleanup)

## Test plan
- [ ] Verify CI passes on this PR (all required checks green)
- [ ] Confirm `integration-tests` job shows `timeout-minutes: 15` in the workflow run

https://claude.ai/code/session_01D6WHX7XTiBdBrnwBZ9uMx7
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_01D6WHX7XTiBdBrnwBZ9uMx7)_


aviadshiber:

**Verdict: COMMENT — approve direction, ship after considering follow-ups.** Votes: 2 × COMMENT, 1 × APPROVE, 0 × REQUEST_CHANGES.

### Consensus findings

1. **`e2e-tests` has the same hang exposure** (all 3 reviewers). It also runs Claude Code against the live API but has no `timeout-minutes`, so it would still fall back to GitHub's 6-hour default. Out of scope for this PR, but worth a follow-up issue so the fix isn't half-applied.
2. **No on-failure diagnostic step** (#1, #2). When the 10-min step cap fires, GitHub kills the process and the logs lack mock-server request log, `ss -tnp`, and `ps auxf`. Next timeout will be nearly as opaque as the original hang. Suggestion: add an `if: failure()` step that dumps mock-API server output and network state.
3. **Symptom vs. root cause** (#2 strong). The underlying bug — Claude Code reaching `api.anthropic.com` instead of honouring `ANTHROPIC_BASE_URL` — isn't addressed. Defense-in-depth options: block egress to `api.anthropic.com` in the runner (hosts override or iptables), and assert that the mock server received ≥1 request after the test (fail-fast instead of timeout).
4. **Convention is partially consistent** (all 3). Matches the existing step-level pattern in `container-tests` (30m build / 45m tests). `unit-tests`, `lint`, `config-validation`, `verify-pinning` lack timeouts but are fast/deterministic, so deferring is reasonable.

### Values sanity check
- 15min job / 10min step: ~5 min headroom for checkout + `npm install -g @anthropic-ai/claude-code@latest` + git config. Step fires first with a clear error before the job cap. All three reviewers found these values reasonable.

### Recommendation
Merge as-is — it's a low-risk safety net that unblocks the queue today. Open follow-up issues for: (a) timeout on `e2e-tests`, (b) `if: failure()` diagnostic step, (c) hardening the mock-API harness against egress leaks.

---

aviadshiber:
@claude open issues for followups

PR #369, opened 2026-05-18
fix(dashboard,test): increase macOS fs.watch budget to 4000 ms

aviadshiber:
$(cat <<'EOF'
## Root cause

PR #368 fixed the `status-additional.test.ts` deletion test by replacing a fixed `Bun.sleep(400)` with a 30×50 ms = 1500 ms polling loop. However, the test is **still failing** on `macos-latest` CI after that fix.

`status.test.ts` (the sibling test) already documents the real constraint:

> *"fs.watch on macOS can take several seconds to deliver the first event under test load"*

It uses **80×50 ms = 4000 ms** for its write-event test and that test passes reliably. The deletion test used only 1500 ms — still not enough on a loaded macOS CI runner.

## Fix

Increase the polling loop from 30 to 80 iterations (1500 ms → 4000 ms), matching the proven budget in `status.test.ts`.

```ts
// Before (still flaky on macos-latest)
for (let i = 0; i < 30 && !drops.includes("kapsis-demo-drop1.json"); i++) {

// After
for (let i = 0; i < 80 && !drops.includes("kapsis-demo-drop1.json"); i++) {
```

The loop exits early as soon as the event fires — no extra latency on fast runners (Linux inotify typically fires in <100 ms).

## Evidence

- `test (macos-latest)` still failing on PR #363 after PR #368 merged: job 76479589928
- `status.test.ts` write-event test uses 80×50 ms and passes consistently on both platforms

https://claude.ai/code/session_013x8NxAoUauaRZBypSCJkMK
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_013x8NxAoUauaRZBypSCJkMK)_


aviadshiber:
Superseded by #363's followup commit. `main` already has the same test with a more generous 5000 ms / 100-iteration budget (vs. this PR's proposed 4000 ms / 80 iterations), so merging this would be a small regression. The fix this PR was aimed at is live on main as of commit `b5010d0` (now in the squash-merge of #363).

PR #375, opened 2026-05-18
fix(launch): serialize sandbox setup to work around macOS AVF virtio-fs cache race

aviadshiber:
## Root cause (debug-debate diagnosis, 2026-05-18)

On macOS Podman applehv hosts, the Apple Virtualization framework's virtio-fs server has a documented cache-coherency bug:

- Apple Feedback **FB16008360**
- podman/podman#23061 (macOS vm virtiofs concurrency issue)
- podman/podman#24725 (mkdirat returns EACCES on virtio-fs)
- docker/for-mac#6690 (Docker already shipped a cache-invalidation-on-hardlink workaround that Podman 5.8.2 lacks)

When two agent launches overlap their heavy host-side metadata I/O window (`git worktree add` on a multi-GB repo, sanitized-git prep, overlay upper/work dir creation), the shared AVF virtio-fs cache wedges and ALL in-flight containers see lookup failures **in the same instant** — producing same-second exit-code-4 `mount_failure` clusters across unrelated agents.

### Smoking gun

Inside the macOS Podman VM, `sudo dmesg | grep overlayfs` shows EACCES (errno 13) lines for every dead agent ID in the corpus:

```
overlayfs: failed to create directory /Users/aviad.s/.ai-sandboxes/products-c4b778/work/work (errno: 13); mounting read-only
overlayfs: failed to create directory /Users/aviad.s/.ai-sandboxes/products-b55e6d/work/work (errno: 13); mounting read-only
overlayfs: failed to create directory /Users/aviad.s/.ai-sandboxes/products-d29782/work/work (errno: 13); mounting read-only
overlayfs: failed to create directory /Users/aviad.s/.ai-sandboxes/helm-charts-6cd095/work/work (errno: 13); mounting read-only
overlayfs: failed to create directory /Users/aviad.s/.ai-sandboxes/products-5eb6fc/work/work (errno: 13); mounting read-only
overlayfs: failed to get metacopy (-2)
```

8 failures observed in a ~3 h window on one developer's host. Two same-second clusters confirmed:
- `b55e6d` (07:32:56Z) + `e7a561` (07:33:27Z) both fail at exactly **07:39:55Z**
- `36f288` (07:55:21Z) + `7095ff` (07:55:18Z) both fail at exactly **08:03:14Z**

Each pair is a launch within 30 s of its partner — both members inside the ~109 s metadata-I/O window simultaneously when the cache wedged.

## Fix

New `scripts/lib/launch-lock.sh` provides a host-scoped advisory mutex around the launch I/O burst. Implementation choices:

- **`mkdir`-based lock** (atomic on POSIX, portable to macOS without extra deps; follows the existing `_gc_lock_acquire` pattern in `launch-agent.sh`).
- **Wraps `setup_sandbox`** so two agents can't enter the heavy worktree + sanitized-git phase simultaneously.
- **Configurable post-cooldown** (default 5 s on Darwin, 0 elsewhere) inside `launch_lock_release` lets the VM-side overlay mount of the current agent's container settle before the next launch's `podman run` starts a fresh overlay against the same cache.
- **Default-on for Darwin only**: on Linux hosts virtio-fs runs natively in the kernel and the AVF cache race does not apply, so the lock would add pointless contention.
- **Stale-lock recovery**: if the PID file in the lock dir names a dead process, the lock is force-released and re-acquired — so a crashed launch doesn't deadlock the next agent.
- **Bounded acquire-timeout** (default 600 s): a wedged holder eventually fails loudly rather than hanging forever.
- **EXIT-trap release**: safety net for the case where the script dies between `launch_lock_acquire` and the normal release point.

### Env-var knobs

| Variable | Default | Notes |
|---|---|---|
| `KAPSIS_LAUNCH_LOCK_ENABLED` | `true` on Darwin, `false` elsewhere | Opt-out for users without the bug |
| `KAPSIS_LAUNCH_LOCK_TIMEOUT` | `600` (s) | Max wait before giving up |
| `KAPSIS_LAUNCH_LOCK_POST_COOLDOWN` | `5` on Darwin, `0` elsewhere | Tuning for VM-side mount stabilization |
| `KAPSIS_LAUNCH_LOCK_DIR` | `$HOME/.kapsis/locks/launch.lock.d` | Custom lock path |

## Tested

- `bash -n` on both files (no syntax errors)
- Acquire / release happy path
- Stale-lock recovery (dead PID file in lock dir → reclaimed cleanly)
- Contention (parent waits for child to release, then acquires)
- Disabled mode (`KAPSIS_LAUNCH_LOCK_ENABLED=false` is a clean no-op)

Smoke-test script in commit message.

## Structural follow-up (separate PR)

This PR mitigates the **trigger** (concurrent launch overlap). The **bug class** (overlay metadata on virtio-fs is inherently racy) is a separate, larger fix: move overlay `upper`/`work` dirs off the virtio-fs share onto VM-native ext4 (Podman named volume). The `/workspace` bind can stay virtio-fs; only the overlay-internal metadata needs to move.

Tracking the bug-class fix as a follow-up issue.

### DIFF
--- a/scripts/launch-agent.sh
+++ b/scripts/launch-agent.sh
@@ -128,6 +128,12 @@ source "$SCRIPT_DIR/lib/vfkit-watchdog.sh"
 # Source host-side status volume sync (Issue #276) — no-op when KAPSIS_STATUS_VOLUME unset
 source "$SCRIPT_DIR/lib/status-sync.sh"
 
+# Source launch-phase serialization mutex (macOS AVF virtio-fs cache-coherency
+# bug workaround). Default-on for Darwin, no-op elsewhere; see launch-lock.sh
+# for design rationale and env-var knobs (KAPSIS_LAUNCH_LOCK_ENABLED,
+# KAPSIS_LAUNCH_LOCK_TIMEOUT, KAPSIS_LAUNCH_LOCK_POST_COOLDOWN).
+source "$SCRIPT_DIR/lib/launch-lock.sh"
+
 # Network isolation mode: none (isolated), filtered (DNS allowlist - default), open (unrestricted)
 NETWORK_MODE="${KAPSIS_NETWORK_MODE:-$KAPSIS_DEFAULT_NETWORK_MODE}"
 CLI_NETWORK_MODE=""  # Track if CLI explicitly set network mode
@@ -2616,6 +2622,11 @@ main() {
             stop_status_sync "${AGENT_ID:-}" "$KAPSIS_STATUS_VOLUME" \
                 "${KAPSIS_STATUS_DIR:-$HOME/.kapsis/status}" 2>/dev/null || true
         fi
+        # Release the launch-phase lock if we still hold it on abnormal exit
+        # (the normal release happens right after setup_sandbox in main flow,
+        # but a crash between acquire and that release would leak the lock
+        # to the next agent's acquire-timeout). Idempotent.
+        launch_lock_release 2>/dev/null || true
         # Delegate backend-specific cleanup (secrets env file, inline spec, dns pin, etc.)
         backend_cleanup 2>/dev/null || true
         # Ensure status transitions to 'complete' on abnormal exit (Fix #168)
@@ -2708,14 +2719,25 @@ main() {
         status_phase "initializing" 15 "Pre-flight check passed" || true
     fi
 
-    # Setup sandbox (worktree/overlay) — only for backends that support it
+    # Setup sandbox (worktree/overlay) — only for backends that support it.
+    #
+    # Wrapped with the host-scoped launch lock (macOS-only by default) so two
+    # agents don't enter the heavy metadata-I/O window simultaneously and
+    # trip the AVF virtio-fs cache-coherency bug — see launch-lock.sh. The
+    # post-cooldown happens inside launch_lock_release so the VM-side overlay
+    # mount of THIS agent's container settles before the next launch starts.
     if backend_supports "worktree" || backend_supports "overlay"; then
         log_timer_start "sandbox_setup"
+        if ! launch_lock_acquire; then
+            log_warn "Launch lock unavailable — proceeding without serialization (next agent may race)"
+        fi
         if ! setup_sandbox; then
+            launch_lock_release  # release before failing exit
             status_complete 1 "Sandbox setup failed (mode: ${SANDBOX_MODE:-unknown})"
             _STATUS_COMPLETE_SHOWN=true
             exit 1
         fi
+        launch_lock_release
         log_timer_end "sandbox_setup"
     else
         log_info "Backend '$BACKEND' handles sandbox setup in-cluster"
--- a/scripts/lib/launch-lock.sh
+++ b/scripts/lib/launch-lock.sh
@@ -1,0 +1,156 @@
+#!/usr/bin/env bash
+# scripts/lib/launch-lock.sh
+#
+# Launch-phase serialization mutex.
+#
+# Why this exists
+# ---------------
+# On macOS Podman applehv hosts, the Apple Virtualization framework's
+# virtio-fs server has a documented cache-coherency bug (Apple Feedback
+# FB16008360, podman/podman#23061, podman/podman#24725) that produces:
+#
+#   overlayfs: failed to create directory .../<id>/work/work (errno: 13);
+#               mounting read-only
+#   overlayfs: failed to get metacopy (-2)
+#
+# inside the VM and surfaces as Kapsis exit code 4 / error_type=mount_failure.
+# When two agent launches overlap their heavy host-side metadata I/O window
+# (git worktree add on a multi-GB repo, sanitized-git prep, overlay upper/work
+# dir creation), the shared AVF virtio-fs cache wedges and ALL in-flight
+# containers see lookup failures in the same instant — producing "same-second
+# failure cluster" exit-code-4s across unrelated agents.
+#
+# This helper provides a host-scoped advisory lock that callers wrap around
+# their launch I/O burst. While one agent holds the lock, other launches
+# block until it is released (plus an optional post-release cooldown so the
+# VM-side mount has time to stabilize before the next launch's `podman run`
+# starts a fresh overlay).
+#
+# The lock is INTENTIONALLY scoped per host (not per project / not per repo)
+# because the AVF virtio-fs server is a single shared resource. Two agents
+# launching against different repos can wedge the same cache.
+#
+# Design choices
+# --------------
+# - Uses `mkdir` (atomic on POSIX) for the lock — `flock` is not portable to
+#   macOS without an extra Homebrew install, and we already use the mkdir
+#   pattern in kapsis (see _gc_lock_acquire in launch-agent.sh).
+# - Stale-lock recovery: if the PID file inside the lock dir names a dead
+#   process, the lock is force-released and re-acquired. Protects against
+#   crashed launches deadlocking the next agent.
+# - Default-on for Darwin only. On Linux hosts virtio-fs runs natively in
+#   the kernel; the AVF cache race does not apply, so the lock would add
+#   pointless contention. Users can override with KAPSIS_LAUNCH_LOCK_ENABLED.
+# - Acquire-timeout default 600 s — bounded so a wedged holder eventually
+#   fails the new launch loudly rather than hanging forever.
+# - Post-cooldown default 5 s on macOS — measured empirically as enough for
+#   the VM's overlay mount to settle after `podman run`. Tunable via
+#   KAPSIS_LAUNCH_LOCK_POST_COOLDOWN.
+
+# shellcheck disable=SC2034  # used by sourcing scripts
+KAPSIS_LAUNCH_LOCK_DIR="${KAPSIS_LAUNCH_LOCK_DIR:-$HOME/.kapsis/locks/launch.lock.d}"
+KAPSIS_LAUNCH_LOCK_TIMEOUT="${KAPSIS_LAUNCH_LOCK_TIMEOUT:-600}"
+
+# Default-on for Darwin, default-off elsewhere.
+if [[ -z "${KAPSIS_LAUNCH_LOCK_ENABLED:-}" ]]; then
+    if [[ "$(uname -s)" == "Darwin" ]]; then
+        KAPSIS_LAUNCH_LOCK_ENABLED=true
+    else
+        KAPSIS_LAUNCH_LOCK_ENABLED=false
+    fi
+fi
+
+# 5 s post-cooldown on macOS, 0 elsewhere.
+if [[ -z "${KAPSIS_LAUNCH_LOCK_POST_COOLDOWN:-}" ]]; then
+    if [[ "$(uname -s)" == "Darwin" ]]; then
+        KAPSIS_LAUNCH_LOCK_POST_COOLDOWN=5
+    else
+        KAPSIS_LAUNCH_LOCK_POST_COOLDOWN=0
+    fi
+fi
+
+# Internal state — non-empty when this process holds the lock.
+_LAUNCH_LOCK_HELD=""
+
+# launch_lock_acquire — block until the host-scoped launch lock is held.
+#
+# Returns 0 on success, 1 on timeout or if disabled. Idempotent: a process
+# that already holds the lock just returns 0 immediately.
+#
+# Callers should pair this with launch_lock_release in an EXIT trap so the
+# lock is released even on abnormal termination.
+launch_lock_acquire() {
+    if [[ "$KAPSIS_LAUNCH_LOCK_ENABLED" != "true" ]]; then
+        return 0
+    fi
+    if [[ -n "$_LAUNCH_LOCK_HELD" ]]; then
+        return 0  # already held
+    fi
+
+    local lock_dir="$KAPSIS_LAUNCH_LOCK_DIR"
+    local timeout_s="$KAPSIS_LAUNCH_LOCK_TIMEOUT"
+    local pid_file="$lock_dir/pid"
+
+    mkdir -p "$(dirname "$lock_dir")" 2>/dev/null || {
+        log_warn "launch_lock_acquire: cannot create $(dirname "$lock_dir") — proceeding without lock"
+        return 0
+    }
+
+    local waited=0
+    local logged_wait=false
+    while ! mkdir "$lock_dir" 2>/dev/null; do
+        # Stale-lock recovery: if PID file names a dead process, reclaim.
+        if [[ -f "$pid_file" ]]; then
+            local holder_pid
+            holder_pid=$(cat "$pid_file" 2>/dev/null || echo "")
+            if [[ -n "$holder_pid" ]] && ! kill -0 "$holder_pid" 2>/dev/null; then
+                log_warn "launch_lock_acquire: reclaiming stale lock from dead PID $holder_pid"
+                rm -rf "$lock_dir"
+                continue
+            fi
+            if [[ "$logged_wait" != "true" ]] && (( waited == 0 )); then
+                log_info "launch_lock_acquire: waiting for launch held by PID ${holder_pid:-unknown} (max ${timeout_s}s)"
+                logged_wait=true
+            fi
+        fi
+
+        if (( waited >= timeout_s )); then
+            log_warn "launch_lock_acquire: timeout after ${timeout_s}s — proceeding without lock (risk of virtio-fs cluster failure)"
+            return 1
+        fi
+        sleep 1
+        waited=$((waited + 1))
+    done
+
+    echo "$$" > "$pid_file"
+    _LAUNCH_LOCK_HELD="$lock_dir"
+    if (( waited > 0 )); then
+        log_info "launch_lock_acquire: acquired after ${waited}s wait"
+    else
+        log_debug "launch_lock_acquire: acquired immediately"
+    fi
+    return 0
+}
+
+# launch_lock_release — release the host-scoped launch lock (with cooldown).
+#
+# Sleeps for KAPSIS_LAUNCH_LOCK_POST_COOLDOWN seconds BEFORE removing the lock
+# dir, giving the VM-side overlay mount time to stabilize so the next agent's
+# `podman run` doesn't race the current one.
+#
+# Safe to call when no lock is held (no-op).
+launch_lock_release() {
+    if [[ -z "$_LAUNCH_LOCK_HELD" ]]; then
+        return 0
+    fi
+
+    local cooldown="$KAPSIS_LAUNCH_LOCK_POST_COOLDOWN"
+    if (( cooldown > 0 )); then
+        log_debug "launch_lock_release: cooldown ${cooldown}s before release"
+        sleep "$cooldown"
+    fi
+
+    rm -rf "$_LAUNCH_LOCK_HELD" 2>/dev/null || true
+    log_debug "launch_lock_release: released $_LAUNCH_LOCK_HELD"
+    _LAUNCH_LOCK_HELD=""
+}


scripts/lib/launch-lock.sh
```diff
@@ -0,0 +1,156 @@
...
+
+        if (( waited >= timeout_s )); then
+            log_warn "launch_lock_acquire: timeout after ${timeout_s}s — proceeding without lock (risk of virtio-fs cluster failure)"
+            return 1
+        fi
+        sleep 1
+        waited=$((waited + 1))
+    done
+
+    echo "$$" > "$pid_file"
```
aviadshiber, line 125: **🛑 B1 — TOCTOU race in stale recovery + non-atomic acquire**

Two waiters that both observe the same stale `holder_pid` can both `rm -rf "$lock_dir"`, then both retry `mkdir`. Worse: the loser's `rm -rf` from a previous iteration can land *after* the winner's `mkdir` on line 101 has succeeded but *before* the winner's `echo "$$" > "$pid_file"` on line 125. The winner's directory is deleted out from under it, the loser then succeeds at `mkdir` on its retry, and **both waiters set `_LAUNCH_LOCK_HELD` and proceed into `setup_sandbox`** — exactly the race the lock is supposed to prevent.

Tied to this: a waiter that arrives between the winner's `mkdir` (101) and `echo $$` (125) reads an absent `pid_file` and silently skips its stale-recovery branch (line 103 `[[ -f "$pid_file" ]]` → false), waiting a full timeout for an acquire that already completed.

**Fix**: make acquire + PID-write one atomic operation via stage + rename, where only one rename can win:
```bash
local stage; stage="${lock_dir}.stage.$$"
rm -rf "$stage" 2>/dev/null
mkdir "$stage" 2>/dev/null || { sleep 1; ((waited++)) || true; continue; }
echo "$$" > "$stage/pid"
if mv "$stage" "$lock_dir" 2>/dev/null; then
    _LAUNCH_LOCK_HELD="$lock_dir"
    return 0
fi
rm -rf "$stage"  # someone else won the rename
```
`mv` of a directory to a non-existent target is atomic on the same filesystem; only one mover wins, and the PID file is present from the moment the lock dir exists.

---

scripts/lib/launch-lock.sh
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    local waited=0
+    local logged_wait=false
+    while ! mkdir "$lock_dir" 2>/dev/null; do
+        # Stale-lock recovery: if PID file names a dead process, reclaim.
+        if [[ -f "$pid_file" ]]; then
+            local holder_pid
+            holder_pid=$(cat "$pid_file" 2>/dev/null || echo "")
+            if [[ -n "$holder_pid" ]] && ! kill -0 "$holder_pid" 2>/dev/null; then
```
aviadshiber, line 106: **⚠️ I — `kill -0` can't distinguish "dead" from "cross-user EPERM"**

`kill -0 "$holder_pid"` fails with EPERM (exit 1) when the PID belongs to another user — currently indistinguishable from "no such process". With the default `KAPSIS_LAUNCH_LOCK_DIR=$HOME/.kapsis/...` this is largely mitigated, but a shared CI runner with `KAPSIS_LAUNCH_LOCK_DIR=/var/run/...` would incorrectly reclaim a lock held by a live process owned by another user.

**Fix**: distinguish the two via the exit code (ESRCH=1 only):
```bash
if [[ -n "$holder_pid" ]]; then
    kill -0 "$holder_pid" 2>/dev/null
    case $? in
        0) ;;   # alive, owned by us — wait
        1) log_warn "..."; rm -rf "$lock_dir"; continue ;;  # ESRCH only → reclaim
        *) ;;   # EPERM or other — alive but inaccessible, wait
    esac
fi
```

---

scripts/lib/launch-lock.sh
```diff
@@ -0,0 +1,156 @@
...
+            fi
+            if [[ "$logged_wait" != "true" ]] && (( waited == 0 )); then
+                log_info "launch_lock_acquire: waiting for launch held by PID ${holder_pid:-unknown} (max ${timeout_s}s)"
+                logged_wait=true
+            fi
+        fi
+
+        if (( waited >= timeout_s )); then
+            log_warn "launch_lock_acquire: timeout after ${timeout_s}s — proceeding without lock (risk of virtio-fs cluster failure)"
+            return 1
```
aviadshiber, line 119: **⚠️ I — Two issues at the timeout branch**

1. **Silent degradation** — after waiting up to 600s, returning 1 here causes `launch-agent.sh:2731-2733` to `log_warn` and proceed *without* the lock, which deterministically triggers the AVF cluster failure the lock exists to prevent. Orchestrators (the slack-bot referenced in `CLAUDE.md`) would rather get a clean `error_type=launch_lock_timeout` they can retry with backoff. Suggest: return 1 here AND change the caller at `launch-agent.sh:2731` to `status_complete 1 "launch_lock_timeout"; exit 1`. If a conservative rollout is preferred, gate behind `KAPSIS_LAUNCH_LOCK_TIMEOUT_FATAL=true` (default true on Darwin).

2. **Message gives no remediation** — first-time users see this 600s warning and have no idea what to do. Suggest:
```bash
log_warn "Launch lock timed out after ${timeout_s}s; held by PID ${holder_pid:-unknown}. Inspect: ps -p ${holder_pid:-?}; force-release: rm -rf '$lock_dir' (or kapsis-cleanup --release-locks); override: KAPSIS_LAUNCH_LOCK_ENABLED=false. See docs/CONFIG-REFERENCE.md."
```
(The matching `--release-locks` flag would be added to `kapsis-cleanup.sh`.)

---

scripts/lib/launch-lock.sh
```diff
@@ -0,0 +1,156 @@
...
+# Safe to call when no lock is held (no-op).
+launch_lock_release() {
+    if [[ -z "$_LAUNCH_LOCK_HELD" ]]; then
+        return 0
+    fi
+
+    local cooldown="$KAPSIS_LAUNCH_LOCK_POST_COOLDOWN"
+    if (( cooldown > 0 )); then
+        log_debug "launch_lock_release: cooldown ${cooldown}s before release"
+        sleep "$cooldown"
```
aviadshiber, line 150: **⚠️ I — EXIT-trap re-entrancy during cooldown sleep**

If the user sends SIGINT (Ctrl-C) while this `sleep "$cooldown"` is running, bash interrupts the sleep, the EXIT trap fires (`launch-agent.sh:2629` calls `launch_lock_release` again), and the early-return guard on line 143 still sees `_LAUNCH_LOCK_HELD` non-empty (lines 153/155 haven't run yet). The re-entrant call falls through to this same cooldown block and re-sleeps the full 5s — a user-visible 10s "frozen on Ctrl-C" delay.

**Fix**: add a re-entry guard at the top of `launch_lock_release` and (related to I3 in the summary) make the cooldown opt-in so the EXIT-trap path can skip it:
```bash
launch_lock_release() {
    local skip_cooldown="${1:-}"
    [[ -n "${_LAUNCH_LOCK_RELEASING:-}" ]] && return 0
    [[ -z "$_LAUNCH_LOCK_HELD" ]] && return 0
    _LAUNCH_LOCK_RELEASING=1
    if [[ "$skip_cooldown" != "skip_cooldown" ]] && (( KAPSIS_LAUNCH_LOCK_POST_COOLDOWN > 0 )); then
        sleep "$KAPSIS_LAUNCH_LOCK_POST_COOLDOWN"
    fi
    rm -rf "$_LAUNCH_LOCK_HELD" 2>/dev/null || { log_warn "leaked: $_LAUNCH_LOCK_HELD"; _LAUNCH_LOCK_RELEASING=""; return 1; }
    _LAUNCH_LOCK_HELD=""
    _LAUNCH_LOCK_RELEASING=""
}
```
Then `launch-agent.sh:2629` calls `launch_lock_release skip_cooldown`.

---

scripts/lib/launch-lock.sh
```diff
@@ -0,0 +1,156 @@
...
+    fi
+
+    rm -rf "$_LAUNCH_LOCK_HELD" 2>/dev/null || true
```
aviadshiber, line 153: **⚠️ I — Ghost-lock divergence when `rm -rf` fails**

`|| true` silently swallows `rm -rf` failures (EBUSY on NFS, EROFS on a degraded mount, etc.). When that happens, line 155 still unsets `_LAUNCH_LOCK_HELD` but the on-disk directory still exists. The next `launch_lock_acquire` call in the same process:
- Bypasses the idempotency guard at line 86 (no in-memory lock).
- Tries `mkdir "$lock_dir"` — fails (still exists).
- Stale-recovery checks `kill -0 $$` — succeeds (we ARE alive).
- → Loops the full 600s timeout and proceeds without lock.

**Fix**: don't unset `_LAUNCH_LOCK_HELD` if `rm -rf` failed:
```bash
if ! rm -rf "$_LAUNCH_LOCK_HELD" 2>/dev/null; then
    log_warn "launch_lock_release: rm -rf failed; lock dir leaked: $_LAUNCH_LOCK_HELD"
    return 1
fi
log_debug "launch_lock_release: released $_LAUNCH_LOCK_HELD"
_LAUNCH_LOCK_HELD=""
```

---

scripts/lib/launch-lock.sh
```diff
@@ -0,0 +1,156 @@
...
+fi
+
+# 5 s post-cooldown on macOS, 0 elsewhere.
+if [[ -z "${KAPSIS_LAUNCH_LOCK_POST_COOLDOWN:-}" ]]; then
+    if [[ "$(uname -s)" == "Darwin" ]]; then
+        KAPSIS_LAUNCH_LOCK_POST_COOLDOWN=5
+    else
+        KAPSIS_LAUNCH_LOCK_POST_COOLDOWN=0
+    fi
+fi
```
aviadshiber, line 70: **⚠️ I + 💡 S — Darwin gate is the wrong shape + raw `uname -s` vs `is_macos`**

1. **Gate shape**: the real bug condition is "podman runs in a VM with a shared virtio-fs", not "macOS host". This gate over-fires on macOS + K8s backend (see inline on `launch-agent.sh:2729`, where the lock takes effect for K8s launches with no benefit) and under-fires on Linux developers running `podman machine` (Fedora Silverblue, etc.). Narrowing to `is_macos && [[ "${BACKEND:-podman}" == "podman" ]]` fixes the over-fire; a `_running_in_podman_machine` helper (cached `podman machine list` check) covers both.

2. **`is_macos` helper available**: `scripts/lib/compat.sh` defines `is_macos`/`is_linux` and is sourced at `launch-agent.sh:118` — before this file at `:135`. Other macOS-conditional libs use it (`vfkit-watchdog.sh:30`, `podman-health.sh:28-29`). Replace both `if [[ "$(uname -s)" == "Darwin" ]]` checks with `if is_macos`. To keep the lib usable standalone:
```bash
declare -f is_macos >/dev/null || is_macos() { [[ "$(uname -s)" == "Darwin" ]]; }
```

---

scripts/launch-agent.sh
```diff
@@ -2616,6 +2622,11 @@ main() {
             stop_status_sync "${AGENT_ID:-}" "$KAPSIS_STATUS_VOLUME" \
                 "${KAPSIS_STATUS_DIR:-$HOME/.kapsis/status}" 2>/dev/null || true
         fi
+        # Release the launch-phase lock if we still hold it on abnormal exit
+        # (the normal release happens right after setup_sandbox in main flow,
+        # but a crash between acquire and that release would leak the lock
+        # to the next agent's acquire-timeout). Idempotent.
+        launch_lock_release 2>/dev/null || true
```
aviadshiber, line 2629: **⚠️ I — EXIT-trap release order delays final status write by 5s**

`launch_lock_release` here runs BEFORE `backend_cleanup` (line 2631), `status_complete` (line 2634), and `display_complete` (line 2643). On macOS the release blocks for `KAPSIS_LAUNCH_LOCK_POST_COOLDOWN=5s` — so every clean exit adds 5s before the final `status.json` write that `kapsis-status.sh --watch` and the dashboard SSE stream are polling for.

The cooldown's purpose ("settle host-side metadata I/O before the *next* launch starts") is preserved regardless of *when* in the trap it runs — by the time the trap fires, the agent's container has long since started. Two fixes:

1. **Reorder**: move this `launch_lock_release` call to AFTER `status_complete`/`display_complete` (but before `display_cleanup`).
2. **OR skip cooldown in the trap path** by using the `skip_cooldown` argument suggested in the inline comment on `launch-lock.sh:147` — cooldown is only useful for the in-flow release at line 2740, where the next launch's `mkdir` is genuinely imminent.

(2) composes better with the existing trap ordering and is what I'd suggest.

---

scripts/launch-agent.sh
```diff
@@ -2708,14 +2719,25 @@ main() {
...
 
-    # Setup sandbox (worktree/overlay) — only for backends that support it
+    # Setup sandbox (worktree/overlay) — only for backends that support it.
+    #
+    # Wrapped with the host-scoped launch lock (macOS-only by default) so two
+    # agents don't enter the heavy metadata-I/O window simultaneously and
+    # trip the AVF virtio-fs cache-coherency bug — see launch-lock.sh. The
+    # post-cooldown happens inside launch_lock_release so the VM-side overlay
+    # mount of THIS agent's container settles before the next launch starts.
     if backend_supports "worktree" || backend_supports "overlay"; then
```
aviadshiber, line 2729: **⚠️ I — K8s backend accidentally takes the lock**

`scripts/backends/k8s.sh:200` declares `worktree) return 0;  # K8s pods clone and handle git`, so this guard is `true` for the K8s backend. On a macOS host with `BACKEND=k8s`:
- `KAPSIS_LAUNCH_LOCK_ENABLED=true` by default (Darwin).
- The guard is true (K8s claims worktree).
- The lock fires — but K8s does sandbox setup *in-cluster*, so there is no host-side metadata I/O burst on the macOS Podman VM to serialize.
- Worse: a single hung K8s launch (or a wedged in-cluster job) blocks all parallel K8s launches on the host for up to 600s.

**Fix**: gate on backend explicitly, not on a capability proxy that K8s technically satisfies. One option:
```bash
if [[ "${BACKEND:-podman}" == "podman" ]] && (backend_supports "worktree" || backend_supports "overlay"); then
    # ... lock wrap ...
fi
```
Or — possibly cleaner — push the check into `launch_lock_acquire` itself (early-return when `$BACKEND` is anything other than `podman`).

---

scripts/launch-agent.sh
```diff
@@ -2708,14 +2719,25 @@ main() {
...
         log_timer_start "sandbox_setup"
+        if ! launch_lock_acquire; then
+            log_warn "Launch lock unavailable — proceeding without serialization (next agent may race)"
+        fi
```
aviadshiber, line 2733: **⚠️ I — Three issues clustered in this block**

1. **`sandbox_setup` timer conflates wait + setup time** (line 2730). `log_timer_start "sandbox_setup"` runs before the acquire at 2731, so the recorded `sandbox_setup` duration includes up to 600s of lock-wait. Metrics on contended hosts will be misleading.

2. **Status invisibility during wait** (2731). `status.json` was set to `phase=initializing, progress=15` at line 2719 and stays there for the entire lock wait. To the dashboard and `kapsis-status.sh --watch`, a waiter is indistinguishable from a hung initializer.

3. **Timeout proceeds without lock** (2731-2733). When `launch_lock_acquire` returns 1 (600s timeout), this `log_warn`+continue triggers the AVF cluster failure the lock prevents — worst of both worlds. (Detailed fix in the inline comment on `launch-lock.sh:117`.)

**Combined fix** for (1) + (2) + (3):
```bash
log_timer_start "launch_lock_wait"
status_phase "waiting_for_launch_lock" 14 "Waiting for concurrent launch" || true
if ! launch_lock_acquire; then
    status_complete 1 "launch_lock_timeout"
    _STATUS_COMPLETE_SHOWN=true
    exit 1
fi
log_timer_end "launch_lock_wait"
log_timer_start "sandbox_setup"
```
Inside `launch_lock_acquire`, when entering the wait loop, refine the status message to `"Waiting for concurrent launch (held by PID ${holder_pid:-?})"`.

Per CLAUDE.md Dashboard Sync Rule: adding `"waiting_for_launch_lock"` to the phase union also requires updating `dashboard/shared/src/index.ts` `AgentStatus` and adding a `StatusPill` render in the UI.

---

aviadshiber (commented):
## Ensemble Review

> **Disclosure**: produced by 5 specialised reviewers running in parallel (correctness/race-safety, integration, shell quality/style, cross-platform/operational, test coverage). Findings have been deduplicated and ranked. The PR author retains final judgment on every item.

### Intended verdict: `REQUEST_CHANGES` — posted as `COMMENT`

You asked for `REQUEST_CHANGES`, but GitHub blocks an author from request-changing their own PR. Posting as `COMMENT` instead; treat the findings below as if the verdict were `REQUEST_CHANGES`.

Primary reason for the strong verdict: **B1** below — a TOCTOU race in the stale-lock recovery path lets two waiters double-acquire the host lock, defeating the PR's stated guarantee in the same stale-lock scenario the PR explicitly claims to handle ("if PID file names a dead process, the lock is force-released and re-acquired"). Concentrated cluster of important findings around lock semantics, integration ordering, and observability on top of that.

The diagnosis is excellent and the overall direction is sound — addressing B1 + I1–I7 + I10 lands this safely.

### What the PR does well

- Outstanding root-cause work — the dmesg evidence (errno 13 + same-second cluster timestamps) is a tight story; the linked Apple/Podman issues are the right ones.
- Implementation mirrors the existing `_gc_lock_acquire` pattern at `scripts/launch-agent.sh:1127-1148`.
- ShellCheck-clean at `severity=warning` (verified locally); `bash -n` clean.
- EXIT-trap safety-net release at `launch-agent.sh:2629` is a thoughtful belt-and-braces touch.
- Sensible defaults: Darwin-only with `KAPSIS_LAUNCH_LOCK_ENABLED` opt-out.

### Findings summary

| Severity | Count | Notes |
|---|---|---|
| 🛑 Blocking | 1 | Inline on `launch-lock.sh:101-125` |
| ⚠️ Important | 10 | Inline where line-specific; rest below |
| 💡 Suggestion | 10 | Below |
| 🪶 Nit | 6 | Below |

---

### 🛑 Blocking

**B1. TOCTOU race in stale recovery + non-atomic acquire** — `launch-lock.sh:101-125`. Inline. Two waiters observing the same dead PID can both `rm -rf` + `mkdir`-acquire, AND the PID write at line 125 happens after the `mkdir` wins so a third waiter sees an empty `pid_file` and silently skips its stale-recovery branch. Fix is the stage-and-rename approach in the inline.

---

### ⚠️ Important

- **I1. K8s backend takes the lock on macOS** — `scripts/backends/k8s.sh:200` declares `worktree) return 0`, so the guard at `launch-agent.sh:2729` fires for K8s launches even though K8s does sandbox in-cluster. Gate on `BACKEND=podman` explicitly. Inline on `launch-agent.sh:2729`.
- **I2. Lock-acquire timeout silently degrades** — `launch-agent.sh:2731-2733` proceeds without the lock after 600s, deterministically triggering the AVF cluster failure the PR exists to prevent. Should `status_complete 1 "launch_lock_timeout"; exit 1`. Inline on `launch-lock.sh:117-119` and `launch-agent.sh:2730-2733`.
- **I3. EXIT-trap release order delays final status write by 5s** — `launch-agent.sh:2629` runs the 5s cooldown before `backend_cleanup`/`status_complete`/`display_complete`. Every clean exit blocks 5s before the final `status.json` write the dashboard and `kapsis-status.sh --watch` are polling for. Inline.
- **I4. Status invisibility during 600s wait** — `status.json` stays at `phase=initializing, progress=15` for the entire wait. Add a `status_phase "waiting_for_launch_lock"`. Inline on `launch-agent.sh:2730-2733`.
- **I5. Darwin gate is the wrong shape** — the real condition is "podman in a VM with a virtio-fs share", not "macOS host". Over-fires on macOS+K8s (see I1), under-fires on Linux `podman machine`. Inline on `launch-lock.sh:55-69`.
- **I6. EXIT-trap re-entrancy during cooldown sleep** — SIGINT during the 5s sleep at `launch-lock.sh:150` re-enters `launch_lock_release` through the EXIT trap and re-sleeps the full cooldown. Add a `_LAUNCH_LOCK_RELEASING` re-entry guard. Inline.
- **I7. Ghost-lock divergence on `rm -rf` failure** — `launch-lock.sh:153` `|| true` silently swallows failures (EBUSY on NFS, etc.); in-memory state and on-disk state then diverge, and the next acquire deadlocks. Inline.
- **I8. `kill -0` can't distinguish dead from cross-user EPERM** — `launch-lock.sh:106`. Mitigated by default `KAPSIS_LAUNCH_LOCK_DIR`, but worth fixing defensively. Inline.
- **I9. Documentation gap — 4 new env-vars are undocumented** — `KAPSIS_LAUNCH_LOCK_{ENABLED,TIMEOUT,POST_COOLDOWN,DIR}` are absent from `CLAUDE.md`, `docs/CONFIG-REFERENCE.md`, `docs/ARCHITECTURE.md`, `agent-sandbox.yaml.template`. The established pattern (e.g. `KAPSIS_LIVENESS_*`, `KAPSIS_VFKIT_WATCHDOG_*`) is a dedicated section in `CLAUDE.md` plus an exposed YAML block in the sandbox template. Add a `launch_lock:` block in `agent-sandbox.yaml.template` and parse it.
- **I10. Zero automated tests for a safety-critical library** — the PR's "Tested" section lists 5 manual smoke checks; none reproducible in CI. Add `tests/test-launch-lock.sh` modeled on `tests/test-logging.sh` and wire into the `libs` category at `tests/run-all-tests.sh:60` and `QUICK_TESTS`. Minimum coverage:
  - `test_acquire_release_happy_path`
  - `test_idempotent_acquire` / `test_idempotent_release_noop`
  - `test_stale_reclaim_dead_pid` (use `(exit 0) &; wait $!`-style reaped PID for portability)
  - `test_contention_blocks` (background subshell holds, parent waits, child proceeds after release)
  - `test_acquire_timeout` (with `KAPSIS_LAUNCH_LOCK_TIMEOUT=2`)
  - `test_disabled_is_noop` (with `KAPSIS_LAUNCH_LOCK_ENABLED=false`)
  - `test_post_cooldown_observed` (with `KAPSIS_LAUNCH_LOCK_POST_COOLDOWN=2`)
  - **`test_concurrent_stale_reclaim_race`** — direct regression test for B1
  - `test_setup_sandbox_failure_releases_lock` — protects the inline release at `launch-agent.sh:2735`

---

### 💡 Suggestion

- **S1.** `probe_virtio_fs_health` (in `scripts/lib/podman-health.sh`) isn't currently wired into the launch flow. With this lock, a queue of N waiters on a degraded VM will serialize through "acquire → set up → fail → release" instead of one early probe failing the leader and short-circuiting the queue. Call `maybe_autoheal_podman_vm` once before `launch_lock_acquire`. (Reasonable follow-up PR; not a regression.)
- **S2.** 5s cooldown is paid on the `setup_sandbox` failure path at `launch-agent.sh:2734-2735`. Cooldown's "next launch needs settled mount" semantics don't apply when the current launch is exiting in failure. Use the `skip_cooldown` argument suggested in the inline on `launch-lock.sh:147`.
- **S3.** Validate `KAPSIS_LAUNCH_LOCK_TIMEOUT > KAPSIS_LAUNCH_LOCK_POST_COOLDOWN` at source time — otherwise user override deterministically deadlocks waiters in cooldown.
- **S4.** Stale-lock recovery currently has a "lock dir exists but no PID file" deadlock path (waits the full 600s). After ~30s without a PID file, treat as stale and reclaim.
- **S5.** No cleanup story for the new path. Add `--release-locks` to `scripts/kapsis-cleanup.sh` (`rm -rf "$HOME/.kapsis/locks/launch.lock.d"`). NFS-mounted `$HOME` is genuinely broken (`kill -0` always reports remote PIDs as alive) — document as known-unsupported.
- **S6.** Per CLAUDE.md Dashboard Sync Rule, if I4's `waiting_for_launch_lock` phase is adopted, also update `dashboard/shared/src/index.ts` `AgentStatus.phase` union and add a `StatusPill` render. Otherwise open a `dashboard-sync` follow-up issue.
- **S7.** Add `tests/test-launch-serialization.sh` (integration category) — two parallel `launch-agent.sh --dry-run` runs, assert that the second's `sandbox_setup` log timer starts after the first's ends. Exercises the wiring the unit test can't.
- **S8.** Add a test that `setup_sandbox` failure still releases the lock — protects the inline release at `launch-agent.sh:2735` from silent regression.
- **S9.** User-facing log messages need remediation hints. Current `"timeout after 600s — proceeding without lock (risk of virtio-fs cluster failure)"` doesn't tell users what to do. Suggested rewrite in the inline on `launch-lock.sh:117`.
- **S10.** Use `is_macos` from `scripts/lib/compat.sh` (already sourced) instead of raw `uname -s` at `launch-lock.sh:56, 65`. Consistency with `vfkit-watchdog.sh:30`, `podman-health.sh:28-29`. Inline.

---

### 🪶 Nit

- **N1.** Stray blank line at `scripts/launch-agent.sh:82` is unrelated to the launch-lock change and pollutes the diff.
- **N2.** Comment at `launch-agent.sh:2727-2728` describes cooldown semantics incorrectly — cooldown runs at sandbox-setup exit, before `podman run`. Reword to "settles host-side worktree/overlay metadata I/O before the next launch starts its own heavy host-side burst."
- **N3.** Numeric env-vars (`KAPSIS_LAUNCH_LOCK_TIMEOUT`, `_POST_COOLDOWN`) aren't validated — `KAPSIS_LAUNCH_LOCK_POST_COOLDOWN=abc` will trip `set -e` inside `(( cooldown > 0 ))`. Sanitize at source time. Same applies to `mkdir -p` at `launch-lock.sh:94` swallowing ENOSPC/EROFS/EACCES.
- **N4.** `cat "$pid_file" 2>/dev/null || echo ""` at `launch-lock.sh:105` is UUOC; `{ read -r holder_pid < "$pid_file"; } 2>/dev/null || holder_pid=""` is more idiomatic. Cosmetic.
- **N5.** Prefer `printf '%s\n' "$$" > "$pid_file"` over `echo` at `launch-lock.sh:125` to match the logging.sh convention. Cosmetic.
- **N6.** Document the 5 module-state vars (`KAPSIS_LAUNCH_LOCK_DIR`, `_TIMEOUT`, `_POST_COOLDOWN`, `_ENABLED`, `_LAUNCH_LOCK_HELD`) in a header comment block so future tests can re-source cleanly without drift.

---

### Verified non-issues (briefing concerns that turned out fine)

- `set -euo pipefail` omission in the header is consistent with `logging.sh`, `compat.sh`, `vfkit-watchdog.sh`, `podman-health.sh`, `status.sh` — all sourced libs omit it. ✓
- Arithmetic at lines 111, 117, 122, 148 is set-e-safe (test-position `(( ))` and assignment-form `$((...))`). ✓
- Missing `init_logging` call is fine — `logging.sh` auto-inits on first `log_*` call via `BASH_SOURCE[2]`. ✓
- `# shellcheck disable=SC2034` at line 50 is correct and matches repo convention. ✓

---

Issue #335, opened 2026-05-13
overlay-mode .claude staging still fails on 2.28.2: cp emits ENOTSUPP + 'preserving permissions', fallback hits RO dst


aviadshiber:
## Summary

Follow-up to #328 (closed by PR #334 in 2.28.2). After installing 2.28.2 and rebuilding the agent image stack (`kapsis-sandbox` → `kapsis-claude-cli` → `kapsis-slack-bot`), overlay-mode staging of `.claude` still fails — but via different cp stderr patterns that PR #334's classifier does not match, plus a likely-pre-existing read-only-destination problem in the end-of-function fallback.

The PR #334 fix IS engaged inside the container (verified `_atomic_cp_with_enoent_tolerance` + `_ATOMIC_CP_STAT_ENOENT_REGEX` etc. are present in `/opt/kapsis/lib/atomic-copy.sh`), but real-world stderr on this kapsis-built image emits different message text than the regex anticipated.

Reproduction: direct `kapsis ~/git/products --config ... --spec ... --agent-id ...` with no `--branch` (overlay mode). Agent `t22821551`, exit 1, 38s elapsed. macOS host, kapsis 2.28.2, freshly rebuilt images.

## Trace

```
[WARN] Overlay failed for .claude (rc=1):
  fuse-overlayfs: cannot retrieve path for /kapsis-upper/.claude:
  No such file or directory
[WARN] atomic_copy_dir: cp failed for .claude:
  cp: cannot stat '/kapsis-staging/.claude/./.git/fsmonitor--daemon.ipc':
  Operation not supported
cp: preserving permissions for '/home/developer/.atomic-copy-dir-1Ac6gY/.':
  Permission denied
[WARN] atomic_copy_dir: using fallback copy for .claude
[WARN] atomic_copy_dir: fallback cp failed for .claude:
  cp: cannot create directory '/home/developer/.claude/./.git':
  Permission denied
cp: cannot create directory '/home/developer/.claude/./issues': Permission denied
cp: cannot create directory '/home/developer/.claude/./plugins': Permission denied
... (~35 more "cannot create directory" lines for each top-level dir)
/opt/kapsis/lib/inject-lsp-config.sh: line 82:
  /home/developer/.claude/settings.local.json: Permission denied
```

Full log: `~/.kapsis/logs/kapsis-t22821551.log`.

### A) cp emits "Operation not supported" (ENOTSUPP), not "No such file or directory" (ENOENT)

```
[...]

PR #380, opened 2026-05-18
fix(image): ship inject-plugin-hooks.sh in container + build-time guard against future drift

aviadshiber:
## Problem

`scripts/lib/inject-plugin-hooks.sh` is in trunk (added in `8d5eea8` feat(plugins): inject Claude Code plugin hooks into settings.local.json), and `scripts/entrypoint.sh` calls `install_plugin_hooks` → `$KAPSIS_HOME/lib/inject-plugin-hooks.sh` (line 1306), but the matching `COPY` line in `Containerfile` was never added.

At runtime inside the container:

1. `KAPSIS_INSTALL_PLUGINS=true` is set correctly (verified by `yq` + `launch-agent.sh` plumbing).
2. `entrypoint.sh` reaches the `claude|claude-cli|claude-code` case and calls `install_plugin_hooks`.
3. `install_plugin_hooks` does `[ ! -f "$inject_script" ]` → file missing → `log_debug "Plugin hook injection script not found — skipping"` (debug-level, invisible by default) → returns `0`.
4. No "Installing Claude Code plugin hooks..." line in the agent's kapsis-*.log; no plugin hooks ever fire; even `/lint-java` and other plugin features silently no-op.

This affects **every container built since 8d5eea8 landed**, regardless of how many plugins are whitelisted via `agent.plugin_whitelist` in the agent yaml.

## How it slipped past the tests (the deeper problem)

The repo had `tests/test-plugin-hook-injection.sh`, `tests/test-e2e-claude-hooks.sh`, etc., but every one of them sources the lib files **directly from the source tree on the host** (`export KAPSIS_HOME="$KAPSIS_ROOT"`, `source '$LIB_DIR/inject-plugin-hooks.sh'`). None of them build or run the container image. Commit `5cd2435 refactor(tests): rename e2e→integration for simulated tests` acknowledged this naming gap but the actual coverage gap (no container-image assertion) remained — so the bug was completely outside the test envelope.

This PR therefore fixes the bug AND addresses the test gap so the same class of regression is caught at PR time.

## Fix

**(1) `Containerfile`: ship the missing file** — single `COPY` line, placed next to its sibling `rewrite-plugin-paths.sh`.

**(2) `Containerfile`: build-time regression guard.** A `RUN` step that enumerates every `$KAPSIS_HOME/lib/<file>` reference in `entrypoint.sh` and asserts each file exists in `/opt/kapsis/lib/` after all `COPY` instructions complete. Fails the build with a remediation hint on miss. Fail-closed if zero references are found (would otherwise pass vacuously). This is the spirit of the Release Artifact Rule we added to CLAUDE.md/AGENTS.md after PR #373.

**(3) `tests/test-container-plugin-hooks.sh`: real container-image e2e test.** Modeled exactly on `tests/test-container-libs.sh` (added after the analogous status.sh-not-in-image bug, see `kapsis-DEV-209078`). Five tests:

- `test_inject_plugin_hooks_lib_exists` — file present in the built image
- `test_inject_plugin_hooks_lib_executable` — chmod 755
- `test_inject_plugin_hooks_function_callable` — sourceable + defines `inject_plugin_hooks`
- `test_install_plugin_hooks_finds_script_when_enabled` — entrypoint helper resolves the path and reaches the "Installing Claude Code plugin hooks..." log line (not silent log_debug skip)
- `test_plugin_hook_injected_into_settings_e2e` — full end-to-end: stage a fixture plugin in a host tmpdir, bind-mount it as `/home/developer/.claude`, run `inject_plugin_hooks` inside `$KAPSIS_TEST_IMAGE` (the same image launch-agent.sh runs for real agents), then grep the marker command out of the resulting `settings.json`

All five run inside the real image via `run_in_container` / direct `podman run`, self-skip via `skip_if_no_overlay_rw` when container support isn't available. Registered in the `security` category (next to its prior-art analog `test-container-libs.sh`) and `QUICK_TESTS`.

## Verification

| Check | Result |
|---|---|
| Guard dry-run against pre-fix state | Lists `inject-plugin-hooks.sh` as single missing file — catches exactly the regression we want |
| Guard dry-run against post-fix state | "All entrypoint.sh /opt/kapsis/lib/ references present (12 files)." |
| `bash -n` on new test file | OK |
| `shellcheck` on new test file | Clean |
| `awk`-eval extraction for `install_plugin_hooks` (used by test 4) | Sanity-tested locally — extracts the function from `entrypoint.sh` and `eval`s it correctly |
| Local image rebuild + 5-test run | In progress on a separate machine; will post results as a follow-up comment |

## Test plan

- [x] Containerfile diff reviewed.
- [x] Guard logic dry-runs PASS post-fix, FAIL with simulated regression.
- [x] New test syntax-checks clean.
- [ ] Local image rebuild completes and all 5 e2e tests pass (in progress).
- [ ] Once merged: cut new release, downstream consumers pick up via `brew upgrade kapsis` + image rebuild.


aviadshiber:
Follow-up tracked in #381 — convert remaining host-only 'e2e/integration' tests to real container-based tests so the test envelope catches the next bug of this shape, not just the COPY-line variant.

aviadshiber:
## Validation — build-time guard fires in real build

Ran `rebuild-bot-image.sh` on a Mac host. Layer 1 (`kapsis-sandbox`) and Layer 2 (`kapsis-claude-cli`) built successfully against this branch's Containerfile. The build-time guard added in this PR executed inside both image builds, BOTH iterations of it (commit 1 + the hardened-against-zero-match followup), and reported:

```
[lib-completeness-check] All entrypoint.sh /opt/kapsis/lib/ references present (12 files).
```

at STEP 73/81 (commit 1) and STEP 73/82 (commit 2, after the hardening) respectively.

This is direct end-to-end validation that:

1. The new `COPY scripts/lib/inject-plugin-hooks.sh` line lands the file in the right place in the image filesystem
2. The build-time guard correctly enumerates the 12 `$KAPSIS_HOME/lib/<file>` references in entrypoint.sh and verifies each is present in `/opt/kapsis/lib/` AFTER all COPYs complete
3. The hardened version doesn't regress on the happy path (still prints the same success line with 12 files)

In other words: the bug this PR fixes (silent log_debug skip because `inject-plugin-hooks.sh` wasn't in the image) **cannot recur on this branch** — the guard would have failed the build with a clear error message before any container could run.

Layer 3 (`kapsis-slack-bot`) failed unrelated to this PR — host memory pressure (Mac swap at 92.6%) killed the podman machine mid `pip install` of an unrelated dep, which is what aborted the rebuild script. The kapsis-side layers are fully validated.

## What's still pending

The new `tests/test-container-plugin-hooks.sh` (5 tests) needs to be run against the rebuilt image as a follow-up to confirm runtime behavior (lib sourceable, function callable, entrypoint helper resolves the path, fixture plugin actually lands in settings.json). Currently blocked on the same podman-machine restart that the slack-bot layer needs. Will post results once podman is back up.

Until then, the build-log evidence above + the dry-run audits in the PR description + the syntax/awk/shellcheck checks should be sufficient to land. The follow-up test results will confirm the runtime semantics but won't change the verdict on whether the COPY+guard combo works as designed.

aviadshiber:
## Post-restart validation — `tests/test-container-plugin-hooks.sh` 5/5 PASS

Ran the new container-based e2e test against the rebuilt images after a clean Mac restart (host swap back to 0/0, podman machine restarted, layers 1+2 already had the COPY fix from `dd9347f`, slack-bot layer rebuilt).

```
[PASS] test_inject_plugin_hooks_lib_exists
[PASS] test_inject_plugin_hooks_lib_executable
[PASS] test_inject_plugin_hooks_function_callable
[PASS] test_install_plugin_hooks_finds_script_when_enabled
[PASS] test_plugin_hook_injected_into_settings_e2e

Total:  5
Passed: 5
Failed: 0
```

End-to-end path confirmed: `inject-plugin-hooks.sh` is shipped in the image, `install_plugin_hooks()` resolves it (no silent skip), and the fixture plugin's hook command lands in `settings.json` after injection. The bug behind this PR is reproduced and fixed.

### Side find — fixture bug in the e2e test (commit `ba06b2c`)

The first run actually came back 4/5 with the e2e test failing on `host-enabled & whitelisted: 0`. Root cause was the fixture itself, not the production path: it wrote `installed_plugins.json` in v1 (object) shape, but the injector's schema-defense check at `inject-plugin-hooks.sh:182-186` requires v2 (array of versions) — which matches the real on-disk layout (verified against `~/.claude/plugins/installed_plugins.json`).

Wrapping the plugin spec in `[...]` makes the fixture match the production shape and the suite goes green. Pushed as a 4th commit on the branch so PR #380 carries a self-validating test.

### Why this still matters even though tests 1-4 passed

Tests 1-4 prove the COPY fix at the file level. Without the fixture fix, test 5 would have continued to silently exit 0 (`No plugin hooks to inject (host-enabled & whitelisted: 0)`) every time — so the very test designed to catch end-to-end regressions of *this* PR's bug would have masked them. Worth folding in before merge.

aviadshiber:
## Phase A end-to-end verified — hooks DO load under `claude -p`

Sidestepped the bot and ran `launch-agent.sh ~/git/kapsis --config ~/.kapsis/phase-a-test.yaml` with `install_plugins: true`, `inject_gist: true`, `KAPSIS_AUDIT_ENABLED: true` and a one-line Write task. Smoking gun in the preserved worktree at `~/.kapsis/worktrees/kapsis-561ab3/.kapsis/gist.txt`:

```
Writing: phase-a-marker.txt
```

That string is produced only by `kapsis-gist-hook.sh::derive_summary()` firing on PostToolUse. The hook header explicitly says "Fires on every PostToolUse event" and requires `KAPSIS_INJECT_GIST=true` (both true in this run). The hook output file cannot exist unless the hook actually fired inside the container.

### Phase A decision matrix outcome

| Q | Evidence | Result |
|---|---|---|
| Q1: Plugin-hook injector ran? | `[entrypoint] Installing Claude Code plugin hooks...` log line (run `da286e`, then again `daf1d7`) — the **production path** for the COPY fix this PR adds | ✅ YES |
| Q2: Kapsis hooks fire under `-p`? | `.kapsis/gist.txt = "Writing: phase-a-marker.txt"` | ✅ YES |
| Q3: Plugin hooks fire? | Same code path as Q2; not isolated with a marker plugin yet but plumbing demonstrably works | ✅ Strongly indicated |
| Q4: `settings.local.json` loaded? | Implied by Q2 — the gist hook lives there | ✅ YES |
| Q5: Task completed? | `phase-a-marker.txt` written with exact content | ✅ YES |

**Row 1 of the runbook decision matrix → Phase B/C of the plugin-hook plan is unblocked. B0 stream-json fallback is NOT needed.**

### Why this strengthens the PR's case

Without this PR's COPY fix, the `install_plugin_hooks()` call in `entrypoint.sh` silently `log_debug`-skipped for every container build for ~6 days, breaking the entire downstream hook system the bot depends on — not just plugin hooks. Phase A confirms that once the COPY is in place, the full pipeline works end-to-end under `claude -p`.

### Side findings (file separately, do not block this PR)

1. **`debug-${ID}.log` is consistently 0 bytes** even on clean exit-0 runs with `--debug -p`. Old "OOM-killed before stderr flushed" theory is wrong — it's a claude-CLI observability gap.
2. **Issue #256 (commit_failure exit 6) trips on kapsis self-runs** because `.pre-commit-config.yaml` requires the `pre-commit` tool which isn't in the container. Bot avoids this with `KAPSIS_COMMIT_NO_VERIFY=true` (would be useful to document in CLAUDE.md).
3. **`gist_updated_at` stays `null` in `status.json`** even when `.kapsis/gist.txt` is written — file→JSON propagation isn't happening for minimal configs that don't enable the LLM gist agent. Cosmetic only.

aviadshiber:
## PR #380 follow-up — atomic_copy_dir bind-mount-busy + fuse-overlayfs upstream fix

Building on the Phase A verification comment above. Phase E (the end-to-end matrix verification) uncovered that **PR #380's COPY fix is necessary but not sufficient** — a second, independent bug breaks the host→container propagation of `~/.claude/settings.json::enabledPlugins`. With `enabledPlugins` arriving empty inside the container, `inject-plugin-hooks.sh`'s filter rejects every plugin regardless of the whitelist, and plugin hooks (deeperdive-java-linter etc.) never fire in the bot.

This follow-up adds three commits to the branch that fix the second bug and add regression tests.

### The bug, in three rows

| Layer | `settings.json` keys | `enabledPlugins` count |
|---|---|---|
| Host `~/.claude/settings.json` | 13 | **37** |
| `/kapsis-staging/.claude/settings.json` (RO mount) | 13 | **37 (intact)** |
| Container `$HOME/.claude/settings.json` (BEFORE these commits) | 2 (just `hooks` + `attribution`) | **0** ❌ |

Root cause is two layers cascading:

1. **fuse-overlayfs fails for `.claude`** because `/kapsis-upper` doesn't exist in the image (Containerfile never created it). The per-entry `mkdir -p $upper` at `entrypoint.sh:655` silently swallowed the EACCES. Every staged config falls back to `atomic_copy_dir` on every cold start.
2. **`atomic_copy_dir`'s replacement strategy** uses `rm -rf "$dst" && mv "$tmp_dir" "$dst"` with no error check. When `$dst` contains a podman bind-mount descendant (`.claude/conversations` is bind-mounted via `-v ~/.kapsis/conversations/<id>:/home/developer/.claude/conversations`), `rm -rf` silently returns 0 but cannot unlink the busy mount. GNU `mv tmp_dir dst-as-existing-dir` then moves tmp **INSIDE** `$dst` (`$HOME/.claude/.atomic-copy-dir-XXX/`). Function returns 0; caller thinks staging succeeded.

### The three new commits

- `0fe56c7 fix(image): pre-create /kapsis-upper and /kapsis-work as developer-owned` — Containerfile adds a new `RUN` step before `USER ${USERNAME}`; entrypoint.sh:655 surfaces previously-swallowed mkdir failures via `log_warn`.
- `2e54277 fix(atomic-copy): detect rm-rf-leaves-dst busy-mount, fall back to merge cp` — post-`rm -rf "$dst"` existence check; if `$dst` survives → merge-style `cp -rp` instead of mv-into-existing-dst. Also `set -o pipefail` at file scope (NOT `-e`/`-u`; existing code relies on `|| true` patterns).
- `51b9f12 test: cover bind-mount-busy descendant and .claude enabledPlugins staging` — new unit test for the busy-mount scenario using `mount --bind` inside a `--cap-add=SYS_ADMIN` container, plus new `tests/test-container-claude-staging.sh` with 2 image-level tests (`/kapsis-upper`/`/kapsis-work` exist + developer-owned, end-to-end staging preserves `enabledPlugins`). SKIPs cleanly on rootless where the bug can't be reproduced.

### Local validation

| Test suite | Result |
|---|---|
| `shellcheck` on all 4 modified `.sh` files | clean |
| `tests/test-atomic-copy.sh` (unit) | 63/63 PASS — no regression |
| `tests/test-atomic-copy-integration.sh` | 3/3 PASS (1 SKIP rootless) |
| `tests/test-container-claude-staging.sh` (NEW) | 2/2 PASS (1 SKIP rootless) |
| `tests/test-container-plugin-hooks.sh` (this PR's own e2e) | 5/5 PASS |

### End-to-end Phase E verification — ALL 4 PASS CRITERIA MET

Ran the same probe agent that originally found the bug (`~/.kapsis/phase-a-test.yaml` config + a 4-step Bash diagnostic task), against the rebuilt image stack:

| # | Assertion | Result |
|---|---|---|
| 1 | `enabledPlugins` length in container `settings.json` matches host (37) | ✅ **`37`** (was `0` before fix) |
| 2 | PostToolUse hooks list contains kapsis hooks AND plugin hooks (esp. `java_linter_reminder.py`) | ✅ all present: `kapsis-gist-hook.sh`, `kapsis-status-hook.sh`, `.../deeperdive-java-linter/.../java_linter_reminder.py`, plus deeperdive-ab-automation, deeperdive-review-ensemble, hookify, claude-island-state, agent-vitals — the bot's full plugin whitelist |
| 3 | No `.atomic-copy-dir-*` directory leftover in `~/.claude/` | ✅ **NO_LEFTOVER** — the mv-into-existing-dst nesting bug is fixed |
| 4 | `.kapsis/gist.txt` contains derived activity (Phase A baseline preserved) | ✅ **`Running: jq`** — kapsis-gist-hook still fires correctly |

### Most powerful evidence

My atomic-copy fix's log signature appeared on **two consecutive** cold-start runs:

```
[WARN] [entrypoint] atomic_copy_dir: rm-rf left .claude in place (rm rc=1; 
likely a busy bind-mount descendant such as .claude/conversations). 
Using merge-style cp instead of mv-into-existing-dst (would nest tmp inside dst).
```

That's the bug being **detected** in the actual production code path and the **merge-copy fallback** taking over — proving both that the bug exists and that the fix activates correctly.

### Scope caveats

* On macOS hosts where `/dev/fuse` isn't available to the podman VM, `fuse-overlayfs` falls back to `atomic_copy_dir` even after commit `0fe56c7`. That's not a regression — atomic-copy's new merge fallback handles it cleanly (proven in production above). The Containerfile fix still matters because it gets us back on the overlay fast path wherever `/dev/fuse` IS available (CI runners, Linux hosts).
* The new busy-mount tests require `CAP_SYS_ADMIN` to run; they SKIP cleanly on rootless local dev. CI rootful lanes will exercise them.

aviadshiber:
## Review-ensemble findings addressed in `5476a60`

Ran `/review-ensemble` (5 parallel reviewers: correctness, security, quality, test, performance) against the full PR diff. Verdict: **APPROVE (majority, 3-2)**. Quality, security, performance all APPROVE; correctness and test flagged real issues worth landing before merge.

Applied in one follow-up commit:

| Severity | Concern | Fix |
|---|---|---|
| HIGH (correctness) | `find $dst -mindepth 1 -maxdepth 1 \| while ... rm -rf` lacks `\|\| true` — under `set -o pipefail`, bare-call sites under `set -e` would abort when `find` fails (e.g. dst removed cross-FS mv path) | Trailing `\|\| true` on the pipeline |
| MEDIUM (correctness) | rm-succeeded-but-mv-failed branch leaves `$dst` non-existent, so the subsequent `find` fails | `mkdir -p "$dst"` in the mv-failure branch |
| MEDIUM (correctness) | Merge `cp -rp` was bare; other cp sites use `_atomic_cp_with_enoent_tolerance` to handle macOS+virtio-fs socket/FIFO stat errors (issue #335/#328) | Wrap merge cp the same way |
| MEDIUM (security, CWE-59) | No symlink defense on the merge cp. Attacker-planted symlink in the persistent `conversations/` bind-mount would be followed by GNU cp default semantics | Add `--remove-destination` so dst entries are unlinked before cp opens them |
| HIGH (test) | `test_claude_staging_..._busy_conversations_mount` used `echo "SKIP" && return 0`; framework's `run_test` counts that as PASSED (no per-test SKIPPED counter exists). The rootless test was passing vacuously | Mirror sibling pattern: `log_skip "…"` so the operator at least sees `[SKIP]` in output. Framework's missing skip counter is out of scope. |

### Findings NOT applied (with rationale)

- **HIGH (test) — fragile awk function extraction** (`awk '/^foo\(\) \{$/,/^\}$/'` stops at the first bare `}`): verified empirically that neither extracted function (`setup_staged_config_overlays`, `install_plugin_hooks`) currently has a bare `}` at column 0 inside its body, so the pattern works today. Concern is forward-looking; defer to a separate hardening PR with proper sourceable-lib refactor.
- Performance findings (extra image layer, double `_atomic_count_files` walk on src, 2× I/O in merge path, podman-test container startup overhead): all real, all "medium" but none on a hot path that bites under production load. Captured in follow-up issues.
- Quality findings (comment wording, underscore prefix consistency): cosmetic; defer.
- Test reviewer's medium findings (jq guard, probe memoization, path-taken indicator, --timeout flag, no negative-assertion): worth a separate test-hardening pass; not blocking this PR.

### Validation of the follow-up commit

- shellcheck clean on both files
- `tests/test-atomic-copy.sh`: **63/63** (existing units, no regression)
- `tests/test-atomic-copy-integration.sh`: **3/3** (new wrapper preserves virtio-fs stderr tolerance)
- `tests/test-container-claude-staging.sh`: **2/2**, test-2 now correctly emits `[SKIP]` on rootless instead of masquerading as `[PASS]`
- End-to-end Phase E re-run: `enabledPlugins=37`, `NO_LEFTOVER`, `gist.txt=Running: jq` — all 3 PASS criteria still hold after the hardening

aviadshiber:
@claude everything that is deferred should be filed as issue

PR #393, opened 2026-05-28
feat(cleanup): TTL-based snapshot cleanup + disk pressure warning (#389)

### LINKED ISSUE #389: snapshots/ grows unbounded — 109 GB after 87 days, filled disk to 98%
## Symptom

`~/.kapsis/snapshots/` accumulates one directory per agent run and is never cleaned. On my machine this grew to **109 GB across 852 directories spanning 2026-03-04 → 2026-05-25** (87 days). Combined with `worktrees/` (19 GB) and `recovery-216578-hotfix/` (2.1 GB), total `~/.kapsis/` reached **127 GB**.

This consumed enough disk space (`/dev/disk3s5` at 98% / 10 GiB free) to start breaking unrelated processes — including `mktemp` calls inside kapsis itself.

## Impact (production failure)

The Slack bot listener (which dispatches each Slack thread to a kapsis container) silently stopped processing all messages. Symptom: bot socket connected, events received, **but every agent dispatch failed** with:

\`\`\`
mktemp: mkstemp failed on /var/folders/9x/.../T/tmp.TPadp0dvLI: No space left on device
mktemp: mkstemp failed on /var/folders/9x/.../T//kapsis-resolv-eLhzeK: No space left on device
[kapsis] Failed (exit code: 1, elapsed: 0m 29s)
09:06:57 [ERROR] handlers.claude_handler: Agent bf3de9: infrastructure failure (exit code 125)
\`\`\`

To the user this looked like *"the bot doesn't react"* — there is no surfaced error and no telemetry pointing at disk pressure.

## Expected

One of:
1. Snapshots auto-cleaned when their owning agent completes (preferred — short-lived state).
2. A configurable TTL (e.g. default 7 days) with a housekeeping pass — matches the pattern in the slack-bot housekeeper (`handlers.housekeeping` runs every 900s and prunes specs/inbox/responses/agent_refs/images/audio).
3. At minimum: `kapsis-cleanup --snapshots` flag and an entry in `references/cleanup.md` so users can opt in. Today the cleanup doc lists worktrees, sandboxes, status, sanitized-git, containers, volumes, logs, ssh-cache — **snapshots and conversations are absent**.

## Repro

\`\`\`bash
# Run any kapsis-driven workload for a few weeks (Slack bot dispatch is the easiest)
ls ~/.kapsis/snapshots/ | wc -l
du -sh ~/.kapsis/snapshots/
stat -f "%Sm %N" -t "%Y-%m-%d" ~/.kapsis/snapshots/* | sort | head
# → entries dating back to original install
\`\`\`

Each snapshot directory is 2-3 GB on this codebase (Java monorepo plus dependencies), so accumulation is rapid.

## Suggested fix

- Add a `cleanup_snapshots_older_than` config (default 7d) and run on `kapsis-cleanup` and on agent completion.
- Add `--snapshots` / `--snapshots-older-than` flags to `kapsis-cleanup`.
- Document `~/.kapsis/snapshots/` in `references/cleanup.md` table.
- Emit a warning (or metric) when `~/.kapsis/` exceeds a configurable size threshold so this fails loudly next time instead of silently breaking dispatch via `mktemp`.

## Workaround used today

\`\`\`bash
find ~/.kapsis/snapshots -maxdepth 1 -mindepth 1 -type d -mtime +0 -print0 | xargs -0 rm -rf
\`\`\`

Reclaimed 114 GB; bot recovered immediately after restart.

## Related

See also #382 (separate snapshot/storage issue) — different failure mode but adjacent surface area.

aviadshiber:
Closes #389.

## Summary

`~/.kapsis/snapshots/` previously grew unbounded — orphan snapshot dirs (left behind by hung agents, hard kills, and crashes) accumulated indefinitely because `backend_cleanup()` only fires on a clean trap exit. In the reported incident the directory reached **109 GB / 852 dirs / 87 days**, filled the host disk to 98%, and silently broke Slack-bot dispatch via `mktemp: No space left on device`.

This PR makes `kapsis-cleanup` reclaim orphan snapshots on a TTL by default and adds a disk-pressure warning so the next leak fails loudly via routine cleanup instead of silently via a full disk.

## Changes

- **`clean_snapshots()`** in `scripts/kapsis-cleanup.sh` modeled on `clean_conversations()` with a 14-day default TTL.
  - Runs in the default cleanup block (`kapsis-cleanup`, `kapsis-cleanup --all`).
  - New flags: `--snapshots`, `--snapshots-older-than <days>`.
  - Uses `get_dir_mtime` (not `get_file_mtime`, which rejects non-files).
  - Respects `--dry-run`, `--project`, `--agent` filters, refuses symlinked top-level dirs.
- **`check_disk_pressure()`** runs after every cleanup:
  - Warns when `~/.kapsis/` exceeds threshold (default 50 GB, override via `KAPSIS_DIR_WARN_SIZE_GB`).
  - Writes `~/.kapsis/.disk-usage-cache` (JSON: `bytes`, `at`, `threshold_gb`) for downstream consumers.
  - Set threshold to `0` to disable.
- **Constants** (`scripts/lib/constants.sh`): `KAPSIS_DEFAULT_SNAPSHOTS_TTL_DAYS=14`, `KAPSIS_DEFAULT_DIR_WARN_SIZE_GB=50`.
- **Dashboard sync** — `~/.kapsis/snapshots/` is now scanned by `DiskUsageStore` alongside the other categories so leaks become visible before they're terminal (`dashboard/server/src/{config,store/disk}.ts`).
- **Docs** — `docs/CLEANUP.md` gets the new resource row, "Snapshots" + "Disk Pressure Warning" subsections, and env-var entries.
- **Tests** — `tests/test-cleanup-snapshots.sh` (18 tests, static + functional) registered in the `cleanup` and `validation` categories and in `QUICK_TESTS`.

## Out of scope (deliberate)

- Preflight integration of the disk-pressure check — the cache file is written so a future preflight hook can read it cheaply, but wiring it in is deferred until cleanup-time warnings prove insufficient.
- Launch-time orphan sweep in `launch-agent.sh` — TTL sweep alone covers the agreed scope.
- Replacing the trap-driven `backend_cleanup()` rm — the TTL sweep makes that leak tolerable in the meantime.

## Test plan

- [x] `bash tests/test-cleanup-snapshots.sh` — all 18 tests pass.
- [x] `bash tests/test-cleanup-store-repair.sh` — dispatcher regression check, 18/18 pass.
- [x] `bash tests/test-cleanup-vm-health.sh` — 17/17 pass.
- [x] `shellcheck scripts/kapsis-cleanup.sh scripts/lib/constants.sh tests/test-cleanup-snapshots.sh` — clean (no new findings).
- [x] `cd dashboard && bun test` — 226/226 pass; `bunx tsc --noEmit` clean.
- [x] Manual end-to-end repro:
  ```bash
  mkdir -p ~/.kapsis/snapshots/{stale,fresh}; touch -d "30 days ago" ~/.kapsis/snapshots/stale
  ./scripts/kapsis-cleanup.sh --dry-run        # lists stale, skips fresh
  ./scripts/kapsis-cleanup.sh --force          # removes stale, keeps fresh
  cat ~/.kapsis/.disk-usage-cache              # { "bytes": N, "at": T, "threshold_gb": 50 }
  ```
- [ ] CI quick suite to confirm no broader regression.

https://claude.ai/code/session_016uyFJxGxQNC4X7wsGTMUzv

---
_Generated by [Claude Code](https://claude.ai/code/session_016uyFJxGxQNC4X7wsGTMUzv)_


aviadshiber (commented):
## Ensemble code review (3 independent reviewers, majority vote)

Three reviewers analyzed this PR independently. Below are findings that ≥2 of 3 agreed on — single-reviewer outliers were dropped to keep the signal-to-noise high.

### Consensus findings

**1. `check_disk_pressure` crashes on non-numeric `KAPSIS_DIR_WARN_SIZE_GB`** — `scripts/kapsis-cleanup.sh:1028-1034` *(3/3 reviewers)*

If a user sets e.g. `KAPSIS_DIR_WARN_SIZE_GB="50g"`, the zero-check `[[ "$threshold_gb" -eq 0 ]] 2>/dev/null && return 0` silently fails (the `&&` chain is exempt from `set -e`), and the next line `threshold_bytes=$((threshold_gb * 1073741824))` aborts the script. Because `check_disk_pressure` is the last call in `main()` and runs *after* `print_summary`, the cleanup work completes successfully but `kapsis-cleanup` exits non-zero with no diagnostic — a footgun for the exact callers this PR is trying to protect (slack-bot dispatch, cron wrappers that check `$?`).

One-line fix:
```bash
[[ "$threshold_gb" =~ ^[0-9]+$ ]] || return 0
```
…right after the zero-check, before the arithmetic.

**2. Per-entry symlink guard is dead code** — `scripts/kapsis-cleanup.sh:961` *(2/3 reviewers)*

`[[ -L "$snap_dir" ]] && continue` never fires, because the glob `"$SNAPSHOTS_DIR"/*/` yields paths with a trailing slash, and `[[ -L "path/" ]]` follows the symlink and evaluates false. The harm is bounded — `rm -rf path/` against a symlink-to-dir is a no-op, not a follow-and-delete — so this isn't a security exploit. But it's a misleading comment claiming a guard that doesn't exist, and the same dead pattern is already inherited from `clean_conversations` / `clean_sanitized_git` / `clean_worktrees` / `clean_sandboxes`. Worth a follow-up issue to fix the family.

**3. Test tmpdirs leak on failure** — `tests/test-cleanup-snapshots.sh` (every `mktemp -d` site) *(2/3 reviewers)*

The trailing `rm -rf "$snap_dir"` in each test is skipped if an earlier `assert_*` fails, so `kapsis-389-*` dirs accumulate under `$TMPDIR` on CI fleets with intermittent failures. A `trap "rm -rf '$snap_dir'" RETURN` per test, or registering with the framework's cleanup hook, closes it. Non-blocking.

### Positive verification (consensus)

- **Dispatch refactor is correct** (3/3): every existing selective flag (`--ssh-cache`, `--containers`, `--logs`, `--volumes`, `--images`, `--branches`, `--vm-health`, `--repair`, `--project`/`--agent`) was traced through the new `else { if WORKTREES; if SNAPSHOTS }` block — all reach it with both flags false, so the block is a no-op. No regression in the old "only run the requested action" contract.
- **Env-var resolution chain works** (2/3): docs' `KAPSIS_SNAPSHOTS_TTL_DAYS` is honored via the `${SNAPSHOTS_TTL_OVERRIDE:-${KAPSIS_SNAPSHOTS_TTL_DAYS:-${KAPSIS_DEFAULT_SNAPSHOTS_TTL_DAYS:-14}}}` chain.
- **Dashboard sync is complete** (2/3): `paths()` return type is inferred, so adding `snapshots` propagates automatically; all four test files updated their `paths` literals; `DiskUsageEntry.category` is `string` and the UI renders categories dynamically — no shared-types or UI gap.
- **Arithmetic overflow is not a concern**: bash uses signed 64-bit; even pathological GB values stay well below `INT64_MAX`.

### Ensemble verdict: **APPROVE_WITH_NITS**

Reviewer tally: 2× APPROVE_WITH_NITS, 1× REQUEST_CHANGES (driven by finding #1). The non-numeric env-var crash is the only finding worth addressing before merge — a one-line regex guard. Findings #2 and #3 are fine as follow-ups.

The PR is solid: it cleanly mirrors `clean_conversations()`, tests cover both static structure and functional behavior (18 passing), the dispatch refactor is verified non-regressive, and the dashboard surface is correctly extended.

---
*Generated by ensemble review (3 independent reviewers, majority vote synthesis)*

---

aviadshiber:
## Ensemble review disposition

- **Finding 1** (non-numeric `KAPSIS_DIR_WARN_SIZE_GB` crashes `check_disk_pressure`): verified real → fixed in `310f045` — numeric regex guard right after the zero-check, plus regression test `test_disk_pressure_rejects_non_numeric_threshold`.
- **Finding 2** (per-entry symlink guard is dead code): verified real → fixed in `e48d25e` — the trailing-slash glob indeed makes `[[ -L "path/" ]]` always false. One correction to the review: the harm is *not* bounded — `rm -rf` on a trailing-slash symlink path follows the link and deletes the target directory's contents (verified empirically), so the live guard matters. The guard now strips the slash before testing, with a regression test seeding a symlinked entry pointing at an aged victim dir. The same pre-existing pattern in `clean_conversations()` et al. is left for the follow-up issue the review suggested.
- **Finding 3** (test tmpdirs leak when an assert fails early): verified real → fixed in `e48d25e` — each functional test registers a `RETURN` trap (house pattern from `test-config-verifier.sh`) instead of a trailing `rm -rf`.

Validation: `shellcheck -S warning` clean on both files; `tests/test-cleanup-snapshots.sh` 20/20; full `--quick` suite green.

---

PR #394, opened 2026-05-30
fix(commit): exclude .claude/settings.json and *.bak artifacts from user commits (#391)

### LINKED ISSUE #391: Container artifacts committed to user branch: CLAUDE.md injection, settings.json mutation, .mvn backup
## Problem

When the post-container commit runs `git add -A`, it sweeps up files the Kapsis entrypoint touched during container setup. These end up in the user's feature branch commit and have to be manually cleaned up.

## Observed artifacts (from agent `d2846e`, products repo, 2026-05-26)

| File | What happened |
|------|---------------|
| `CLAUDE.md` | Kapsis entrypoint injected 31-line activity-gist instructions block at the end of the project's CLAUDE.md |
| `.claude/settings.json` | Agent enabled `"ENABLE_LSP_TOOL": "1"` during the session — mutated the project's Claude Code settings |
| `.mvn/extensions.xml.bak2` | Kapsis backed up `extensions.xml` during Maven setup, leaving a `.bak2` file in the tree |

## Impact

The user had to create a follow-up cleanup commit (`870e675e`) to remove these files before the PR was clean. This is confusing and easy to miss in code review.

## Suggested fixes

1. **CLAUDE.md gist injection** — write the instructions to a temp file outside the workspace (e.g. `/tmp/kapsis-gist-instructions.md`) or inject into `.kapsis/` which should already be gitignored.
2. **`.claude/settings.json`** — scope setting mutations to the container's home directory, not the repo-tracked `.claude/settings.json`.
3. **`.mvn/extensions.xml.bak2`** — either don't create the backup, or add `*.bak*` / `.mvn/*.bak*` to the Kapsis-injected `.gitignore` additions before commit.

## Alternative global fix

Before `git add -A`, diff the working tree against the base and exclude any files that were only touched by Kapsis infrastructure (paths matching `.kapsis/`, `.claude/`, `*.bak*`, `CLAUDE.md` gist-injection pattern).

aviadshiber:
## Summary

Trimmed re-roll of this PR on top of current `main`, resolving the conflicts and splitting issue #391 into two disjoint, independently mergeable PRs:

- **This PR (#394)** — commit-exclude and ephemeral-pattern filtering (below).
- **PR #413** — the gist-block strip for `CLAUDE.md`/`AGENTS.md`, re-implemented with **SHA-256 provenance hardening** (Issue #408): sentinel constants, sentinel-wrapping in `inject-status-hooks.sh`, `strip_kapsis_injections()` and its invocation in `post-container-git.sh` now live there. All strip logic (and the `stripped_injections` status field + dashboard plumbing) has been **dropped from this branch** so the two PRs no longer overlap.

## What this PR still covers

1. **`.claude/settings.json` commit exclude** — the file is mutated in-session by Kapsis LSP/plugin hook injection; it is now part of `KAPSIS_DEFAULT_COMMIT_EXCLUDE` (`.claude/settings.json` + `**/.claude/settings.json`) and is unstaged before commit even when git-tracked.
2. **`*.bak` ephemeral patterns (non-overridable)** — `**/*.bak` and `**/.mvn/*.bak*` in `KAPSIS_DEFAULT_EPHEMERAL_PATTERNS` cover Maven plugin backups such as `.mvn/extensions.xml.bak2`. Deliberately scoped (not `**/*.bak*`) so names like `README.bakery.md` or `notes.bak.swp` are never filtered (ensemble review finding).
3. **Glob support in `_pattern_to_regex()`** — a single `*` inside a pattern converts to `[^/]*` (matches within one path component, never across `/`), required by the new patterns.
4. **`KAPSIS_EXTRA_COMMIT_EXCLUDE`** — appends user patterns without redeclaring the defaults; `KAPSIS_COMMIT_EXCLUDE` still replaces the list wholesale (so a default can also be intentionally removed).
5. **Docs** — `docs/CONFIG-REFERENCE.md`, `docs/GIT-WORKFLOW.md`, `CLAUDE.md` sections for the exclusion behavior.
6. **Tests** — `tests/test-kapsis-artifact-filter.sh` (9 quick tests: exclude-list contents, tracked-file unstaging, extra-exclude append, `.bak` pattern matching/over-match guards, full integration scenario), registered in the `git` category and `QUICK_TESTS`.

## What was removed (now owned by #413)

- `KAPSIS_GIST_MARKER_BEGIN`/`_END` + legacy-heading constants in `scripts/lib/constants.sh`
- Sentinel-wrapping changes in `scripts/lib/inject-status-hooks.sh`
- `strip_kapsis_injections()` / `_strip_sentinel_gist_block()` / `_strip_legacy_gist_block()` and the `commit_changes()` hook in `scripts/post-container-git.sh`
- `stripped_injections` status field (`scripts/lib/status.sh`) and its dashboard mirror (`dashboard/shared`, `dashboard/server`, `dashboard/ui`)
- All strip-related tests

## Test plan

- `shellcheck` clean on all touched scripts
- `./tests/run-all-tests.sh --quick` passes
- `bash tests/test-kapsis-artifact-filter.sh` — 9/9 pass

Closes #391 (together with #413 for the injected-block strip).


aviadshiber (commented):
## Ensemble review (majority vote)

Three independent reviewer agents analyzed this PR with different framings (correctness / security & design / style & tests). They did not see each other's findings. Below are the findings where **≥2 reviewers concurred**, followed by notable single-reviewer findings.

**Verdict tally:** 2× REQUEST-CHANGES, 1× APPROVE-WITH-NITS → **majority: REQUEST-CHANGES** (posting as a COMMENT review since GitHub forbids requesting changes on one's own PR).

The core design (sentinel-strip + COMMIT_EXCLUDE + EPHEMERAL_PATTERNS layering) is sound and well-tested for the happy path. The blockers are concrete edge cases that defeat the fix on real-world inputs and a few defensive-coding misses.

---

### Consensus findings — Major (≥2 reviewers)

#### 1. Unguarded `cd "$worktree_path"` in `strip_kapsis_injections` (3/3 reviewers)
`scripts/post-container-git.sh:92`

Every other PWD-changing function in this file uses `cd "$path" || return 1` (see lines 355, 826, 849, 890). The new function uses bare `cd`. Two failure modes:
- Under `set -e` from `post-container-git.sh`'s top-level, a bad path aborts the whole post-container pipeline before subsequent error reporting can run.
- Called from a context where `set -e` is suppressed (e.g. the new test file, or a future caller wrapped in `|| _rc=$?`), the function silently runs in the *caller's* CWD and will edit any `CLAUDE.md` / `AGENTS.md` sitting there.

Fix: `cd "$worktree_path" || { log_error "strip_kapsis_injections: cannot cd to $worktree_path"; return 1; }`.

#### 2. CRLF / trailing-whitespace defeats the awk strip → block leaks into commit (2/3)
`scripts/post-container-git.sh:107-111`

`awk '$0 == begin'` requires byte-exact whole-line equality. If `CLAUDE.md` ever passes through CRLF conversion (Windows clone, prettier-formatted markdown, copy through a CRLF-converting transport) the marker line is `<!-- KAPSIS_GIST_BEGIN -->\r` and never equals the literal — strip silently no-ops and the gist block is committed verbatim. Same hazard for trailing whitespace.

Fix: replace exact equality with `$0 ~ "^[[:space:]]*<!-- KAPSIS_GIST_BEGIN -->[[:space:]]*\\r?$"` (and same for END), and add tests for both CRLF and whitespace-suffix.

#### 3. `[[ -s tmp_file ]]` guard leaks the block when file is 100% injection (2/3)
`scripts/post-container-git.sh:113-124`

If `CLAUDE.md` was created by Kapsis (file did not pre-exist; agent never edited anything outside the gist block), the strip produces an empty tmp file. The guard refuses to overwrite — "would be empty after strip — leaving original" — so the **entire gist block survives and is staged**. Intent of the guard (avoid data loss when awk fails) is reasonable but misses the all-injection case.

Fix: differentiate "empty because awk failed" (check `awk` exit status) from "empty because file was 100% injection" (write the empty file or `rm`).

---

### Consensus findings — Minor (≥2 reviewers)

#### 4. Test coverage gaps (3/3)
`tests/test-kapsis-artifact-filter.sh`

13 tests cover the happy path well but security- and robustness-relevant cases are missing:
- Multiple `BEGIN`/`END` pairs in one file (does it strip first, all, or get confused?)
- Unbalanced markers (BEGIN with no END — current awk silently truncates the tail)
- CRLF line endings (see #2)
- Marker line with trailing whitespace
- Marker inside a fenced code block (e.g. a doc that documents this feature)
- All three reviewers independently asked for the multi-block case in particular.

#### 5. Docs not updated (2/3)
- `docs/GIT-WORKFLOW.md` still describes the pre-#391 pipeline (`git add -A` → validate); add the strip phase and link to the sentinel contract.
- `docs/CONFIG-REFERENCE.md` "Default Excluded Patterns" should mention `.claude/settings.json` is now default-excluded and document the override (which forces redeclaring all defaults — see single-reviewer finding below).
- `CLAUDE.md` "Things to Avoid" / commit-exclude sections are not updated.

#### 6. Test sourcing pollutes `~/.kapsis/logs/` (2/3)
Sourcing `scripts/post-container-git.sh` from the test triggers `log_init "post-container-git"` at source time, creating a real log file in the user's home for every test run (visible in test output). Either unset `_KAPSIS_LOGGING_LOADED` before source, or point `KAPSIS_LOG_DIR` to a tmpdir in `setup_test_repo`.

---

### Notable single-reviewer findings

These didn't cross the 2-vote threshold but are concrete enough to call out:

**Reviewer A (correctness):**
- **Missing-END marker silently truncates tail.** `skip` is set to 1 and never reset; every line after a stray BEGIN is dropped. The `-s tmp_file` guard doesn't catch this because earlier lines are preserved. Track whether `end` was actually seen and discard the rewrite if not. (*Major*)
- **Backward-compat gap.** Old-format `CLAUDE.md` files (with `---` separator instead of HTML sentinels) are never stripped AND never re-injected (the idempotency grep on `"Kapsis Activity Gist"` finds the old body and skips). Net: pre-PR users keep leaking the block forever. PR body acknowledges this — recommend a one-time migration or a fallback strip keyed on the `# Kapsis Activity Gist` heading. (*Major*)
- **Marker idempotency check is fragile.** Grepping for `"Kapsis Activity Gist"` would skip injection for any user CLAUDE.md that happens to mention the phrase. Better to key on the sentinel comment which user content is unlikely to contain. (*Major*)
- **`mktemp` cross-filesystem `mv` is not atomic** on K8s / macOS user-namespaced volumes. Use `mktemp "$(dirname "$md_file")/.kapsis-strip-XXXXXX"`. (*Minor*)

**Reviewer B (security & design):**
- **Adversarial sentinel spoofing.** An agent that wants to hide a chunk of its own changes from the user's commit can wrap them in `<!-- KAPSIS_GIST_BEGIN --> ... <!-- KAPSIS_GIST_END -->` and the host will happily strip them. The host has no provenance check on who wrote the markers. Mitigation: record exact byte ranges or a per-session HMAC at injection time. (*Major*)
- **`**/*.bak*` over-blocks.** Regex `(^|/)[^/]*\.bak[^/]*$` matches `README.bakery.md`, `notes.bak.swp`, fixture data like `tests/fixtures/foo.bak`, DBA schema-migration backups, and dotfile-manager backups. The pattern is in `EPHEMERAL_PATTERNS` which is documented as "non-user-overridable" — silent data loss for any project that legitimately commits these. Recommend scoping (`.mvn/**/*.bak*` + `**/*.bak`) or moving to `COMMIT_EXCLUDE` so users can override. (*Major*)
- **No escape hatch for `.claude/settings.json`.** Teams commonly want to commit project-level Claude Code config (`permissions.allow`, hooks, MCP servers). The only override (`KAPSIS_COMMIT_EXCLUDE`) is whole-replacement, forcing users to redeclare `.gitignore`, `**/.gitignore`, `.gitattributes`, `**/.gitattributes`. Recommend `KAPSIS_EXTRA_COMMIT_EXCLUDE` or an explicit "include even if default-excluded" list. (*Major*)
- **Strip is invisible to audit/status.** Host-side mutation of user-facing commit content with no `audit_event`, no `status.json` field, no commit-message marker. Per CLAUDE.md "Dashboard Sync Rule" this should be surfaced. (*Major*)
- **Sentinel strings hardcoded in two files** — drift risk. Define once in `scripts/lib/constants.sh`. (*Nit*)

**Reviewer C (style & tests):**
- **New test file is mode 0644, not 0755.** Sibling tests are all 0755. The PR's own "Test plan" command (`bash tests/test-kapsis-artifact-filter.sh`) happens to work because it invokes bash explicitly, but `./tests/test-kapsis-artifact-filter.sh` and any future direct-exec call would fail. (*Major* — concrete and trivially verifiable)
- **`commit_changes` doesn't check `strip_kapsis_injections` return.** Combined with the unguarded `cd`, this can abort the pipeline before the existing `[[ -d $worktree_path ]]` checks produce a useful error. (*Major*)
- **Stale comment in `_pattern_to_regex` header block** no longer documents `*` semantics; readers don't know `**/*.bak*` is supported. (*Minor*)
- **Asymmetric inline #391 comment** — `KAPSIS_DEFAULT_COMMIT_EXCLUDE` has a 2-line rationale, the matching `**/*.bak*` entry has none. (*Minor*)
- **Tests use hand-rolled `if grep … log_fail … return 1`** instead of the framework's `assert_file_contains` / `assert_file_not_contains`. `log_fail` alone does not set `_ASSERTION_FAILED`, so a future maintainer dropping the explicit `return 1` would silently pass. (*Major*)
- **`setup_test_repo` ends with a dead `echo "$TEST_REPO"`** that no caller captures — leaks the path into test logs. (*Minor*)
- **`main()` reimplements `print_test_header`** and bypasses `KAPSIS_TEST_QUIET=1` suppression. (*Minor*)
- **PR title puts `(#391)` in the `(scope)` slot** of Conventional Commits; the body already says `Closes #391`. (*Nit*)

---

### Suggested minimal fix set to unblock merge

If you want to land this with the smallest delta that addresses the consensus findings + the highest-impact single-reviewer items:

1. Guard the `cd` in `strip_kapsis_injections` with `|| return 1`.
2. Make awk marker match tolerant of `\r` and trailing whitespace (and add a CRLF test).
3. Track whether END was seen in awk; on missing-END, leave file unchanged and log a warning.
4. Either commit the strip when result is empty, OR delete the file when result is empty — currently neither path is taken.
5. Backward-compat fallback strip for the old `---` format, OR a one-time migration; at minimum document the manual-cleanup workaround prominently.
6. Narrow `**/*.bak*` (scope to `.mvn/**` + `**/*.bak` exact-suffix) OR move to `COMMIT_EXCLUDE` so users can override.
7. `chmod +x tests/test-kapsis-artifact-filter.sh`.
8. Define sentinel strings once in `constants.sh`.
9. Audit-event and `status.json` field for the strip (per Dashboard Sync Rule).

#7 and #8 are 30 seconds. #1, #2, #3, #4 are small targeted fixes. #5, #6, #9 are the design-level items that warrant author judgment.

---

_Ensemble methodology: 3 independent reviewer agents (correctness, security/design, style/tests) given the full diff without seeing each other's output. Findings cross-tabulated by topic; consensus = ≥2 reviewers concurring on a finding. Majority verdict surfaces above; single-reviewer findings reported separately for context._


---

aviadshiber:
## Ensemble review follow-up — dispositions (commit efe4ef4)

All consensus findings and the high-impact single-reviewer items are addressed in efe4ef4; validation at the bottom.

### Consensus — Major
| # | Finding | Disposition |
|---|---------|-------------|
| 1 | Unguarded `cd` in `strip_kapsis_injections` | **Fixed in efe4ef4** — `cd … \|\| { log_error; return 1; }`; `commit_changes` now checks the return (warns, never aborts the user's commit) |
| 2 | CRLF / trailing whitespace defeats the awk strip | **Fixed** — markers matched with `^[[:space:]]*…[[:space:]]*$` (covers `\r`); CRLF + trailing-whitespace tests added |
| 3 | `-s` guard leaks block when file is 100% injection | **Fixed** — awk exit status now distinguishes "strip failed" from "legitimately empty"; the empty result is written (injection only appends to existing files, so this restores the pre-injection empty file) |

### Consensus — Minor
| # | Finding | Disposition |
|---|---------|-------------|
| 4 | Test coverage gaps | **Fixed** — suite now 26 tests: multi-block, unbalanced BEGIN (file byte-identical), CRLF, trailing whitespace, fenced-code-block (documented limitation: strip is not Markdown-aware, but never truncates), 100%-injection, stray END, legacy formats, byte-exact round-trip, `.bak` over-match guards |
| 5 | Docs not updated | **Fixed** — GIT-WORKFLOW.md (PHASE 4 pipeline + new "Infrastructure Artifact Filtering" section), CONFIG-REFERENCE.md (defaults incl. `.claude/settings.json`, non-overridable ephemeral patterns, override/extend docs), CLAUDE.md ("Commit Artifact Filtering" section) |
| 6 | Test pollutes `~/.kapsis/logs/` | **Fixed** — `KAPSIS_LOG_TO_FILE=false` before sourcing (same pattern as test-post-container-git.sh); verified no log file is created |

### Single-reviewer findings
- **A / missing-END truncates tail** — fixed: awk `END { exit skip }`; on unbalanced markers the file is left unchanged with a warning.
- **A / backward-compat gap (old `---` format)** — fixed: conservative legacy strip keyed on the `# Kapsis Activity Gist` heading with a `---` separator above it; refuses (file unchanged + manual-cleanup warning) if another H1 follows the heading. Combined with the idempotency fix below, legacy files get re-injected with sentinels during the session and end fully clean. PR-body "Known limitation" is superseded by this.
- **A / fragile idempotency grep** — fixed: keyed on the `KAPSIS_GIST_MARKER_BEGIN` sentinel, not the prose.
- **A / mktemp cross-fs mv** — fixed: same-directory `mktemp "${md_file}.kapsis-strip-XXXXXX"` → atomic rename.
- **B / adversarial sentinel spoofing** — **deferred**: needs a provenance design (per-session HMAC or recorded byte ranges at injection time). Bounded severity: the agent already controls worktree content arbitrarily; spoofing only lets it hide its own edits from the auto-commit, which the worktree diff still reveals. Can open a follow-up issue.
- **B / `**/*.bak*` over-blocks** — fixed: narrowed to `**/*.bak` (exact suffix) + `**/.mvn/*.bak*` (Maven backups at any depth); regression tests assert `README.bakery.md` / `notes.bak.swp` / `config.xml.bak2`-outside-`.mvn` are never filtered.
- **B / no escape hatch for `.claude/settings.json`** — fixed: new `KAPSIS_EXTRA_COMMIT_EXCLUDE` (additive, documented); CONFIG-REFERENCE.md also documents that `KAPSIS_COMMIT_EXCLUDE` is wholesale-replacement, which is how a team re-includes `settings.json`.
- **B / strip invisible to audit/status** — **fixed for status, deferred for audit**: new `stripped_injections` status.json field (`status_set_stripped_injections`), dashboard shared `AgentStatus` type + AgentDetail render updated per the Dashboard Sync Rule. Audit event deferred: the audit hash chain is per-session and only initialized by in-container hooks (`audit_init` in kapsis-status-hook.sh); a host-side post-container event would start a disconnected chain — needs its own design. The strip is logged (`log_success` with file list) meanwhile.
- **B / sentinels hardcoded in two files** — fixed: `KAPSIS_GIST_MARKER_BEGIN`/`_END` + `KAPSIS_GIST_LEGACY_HEADING` in `scripts/lib/constants.sh`, sourced by both (with a non-readonly fallback in inject-status-hooks.sh for environments without constants.sh).
- **C / test file mode 0644** — fixed: 0755.
- **C / `commit_changes` ignores strip return** — fixed (warn-and-continue; aborting would lose the user's work).
- **C / stale `_pattern_to_regex` comment** — fixed: documents `*` → `[^/]*`.
- **C / asymmetric `#391` comment on `**/*.bak*`** — fixed: rationale comment added on the ephemeral patterns.
- **C / hand-rolled asserts** — fixed: converted to framework asserts (`assert_file_contains`/`_not_contains`, `assert_equals`, `assert_true`/`false`), which also removed the repeated cleanup/return blocks.
- **C / dead `echo "$TEST_REPO"`** — fixed.
- **C / `main()` reimplements `print_test_header`** — fixed: uses `print_test_header` (quiet-mode aware) + `print_summary`.
- **C / PR title `(#391)` in scope slot** — no change: the Conventional Commits scope is `(commit)`; `(#391)` is a trailing issue reference in the description, which is the standard GitHub convention.

### Bonus
The strip now also consumes the single blank line the injector writes before `BEGIN`, making inject→strip **byte-exact** — previously a whitespace-only `CLAUDE.md` diff (one trailing blank line) survived every gist-injecting session, which is exactly the noise class #391 targets. Covered by a new round-trip test that runs the real `inject_gist_instructions()` and asserts `cmp` equality.

### Validation
- `shellcheck -S warning` clean on all modified scripts (matches CI)
- `tests/test-kapsis-artifact-filter.sh` 26/26; `test-validate-staged-files` 15/15; `test-ephemeral-artifact-filtering` 9/9; `test-git-excludes` 7/7; gist/status hook suites green
- `./tests/run-all-tests.sh --quick`: failure set identical to an `origin/main` baseline run in the same sandbox (14 pre-existing environment failures — signing proxy, no keychain, `/kapsis-status` mount); zero regressions from this branch
- CI on efe4ef4: all 22 checks green (incl. Container Tests, dashboard-sync `check`, dashboard test + compile matrices)

---
_Generated by [Claude Code](https://claude.ai/code/session_01GyUwb9P9abPA2WnUAqnCUU)_

---

aviadshiber:

### Critical (merge-blocking)

- **`tests/run-all-tests.sh:78` — new test file registered in `--category` dispatch but not in `QUICK_TESTS`, so CI's `--quick` gate never runs it.**
  The PR only appends `test-kapsis-artifact-filter.sh` to the `validation)` and `git)` branches inside `get_tests_for_category()` (used by `--category <name>`). The `QUICK_TESTS` variable at line 78 — the list actually used by `--quick` mode (`tests_to_run="$QUICK_TESTS"` at line 157) — is untouched by this diff and does not contain `test-kapsis-artifact-filter.sh`. Since `.github/workflows/ci.yml` invokes `./tests/run-all-tests.sh --quick` as the fast CI gate, the new regression tests (including the `.bak`/`.mvn/*.bak*` over-match guards) never execute there; they only run in the slow full/container job or when manually invoked (`bash tests/test-kapsis-artifact-filter.sh`). This contradicts the PR's stated test plan ("`./tests/run-all-tests.sh --quick` passes"), which is true only because the file is silently skipped, not because it ran. Per repo convention, new no-container tests must be added to `QUICK_TESTS` *and* a category. Fix: append `test-kapsis-artifact-filter.sh` to the `QUICK_TESTS=` string on line 78.

### Non-critical

None.

None — both raw findings (from `correctness-reviewer` and `test-reviewer`) described the same underlying bug and were merged into the single critical item above after verification against the diff and file contents on `origin/claude/sharp-cori-2IzsJ`.

### Scope verification notes

- Confirmed **no leftover gist/sentinel/strip logic** remains in this trimmed PR: no `GIST_MARKER`/`SENTINEL`/`LEGACY_HEADING` references in `scripts/lib/constants.sh`, no `strip_kapsis_injections`/`_strip_sentinel_gist_block`/`_strip_legacy_gist_block` in `scripts/post-container-git.sh`, no sentinel wrapping in `scripts/lib/inject-status-hooks.sh`, and no `stripped_injections` field in `scripts/lib/status.sh`. That logic is correctly deferred to PR #413.
- `_pattern_to_regex()`'s new `*` → `[^/]*` glob conversion correctly implements `**/*.bak` and `**/.mvn/*.bak*` for the documented cases (e.g. matches `.mvn/extensions.xml.bak2`, does not match `README.bakery.md` at the top level) — glob semantics check out for the stated exclude/ephemeral patterns.
- Test registration was otherwise correct: the `validation` and `git` category branches in `get_tests_for_category()` were both updated to include `test-kapsis-artifact-filter.sh`.

aviadshiber:
✅ **Round-1 critical resolved** (commit `4051572`): `test-kapsis-artifact-filter.sh` is now registered in `QUICK_TESTS` (line 78) in addition to the validation/git categories — the `--quick` CI gate now runs it (9/9 passing locally).

PR #395, opened 2026-06-05
fix(conversations): populate ~/.kapsis/conversations/ with agent transcripts (#390)

### LINKED ISSUE #390: ~/.kapsis/conversations/ stays empty — no transcripts captured for kapsis-agent-debugging workflows
## Symptom

`~/.kapsis/conversations/` is **always empty** (0 bytes, no files) on a long-running install:

\`\`\`
$ du -sh ~/.kapsis/conversations/
  0B  /Users/aviad.s/.kapsis/conversations
$ ls ~/.kapsis/conversations/
$  # (empty)
\`\`\`

Meanwhile sibling directories have grown to:
- `snapshots/` — 852 dirs, 109 GB (87 days of agent runs)
- `worktrees/` — 7 dirs, 19 GB
- `specs/` — 510 files
- `status/` — 2129 entries

So agents *have* been running constantly — but nothing lands in `conversations/`.

## Why this matters

The **kapsis-agent-debugging** skill documents conversation inspection as a first-class debugging workflow:

> *\"Systematic debugging guide for Kapsis container agents — hung processes, crash analysis, **conversation inspection**, liveness monitor behavior...\"*

When an agent hangs or returns a bad result, the documented recovery path is to read the agent's conversation transcript. Currently there is nothing to read — every post-mortem requires going back to the requesting client (Slack/CLI) and trying to reconstruct what happened from external logs.

## Expected

Either:
1. The Claude CLI agent's transcript should be persisted to `~/.kapsis/conversations/<agent_id>.{md,jsonl}` on agent exit (success or failure), and **especially on failure**, OR
2. The directory should be removed from the repo if it is a vestigial path no longer used, so debugging skill docs aren't misleading.

(1) is the strictly more useful option — debugging an agent that returned no output is much easier with the transcript than without.

## Notes

- `~/.kapsis/audit/` does receive entries (1 MB) but those are policy/permission audits, not conversation content.
- The `references/cleanup.md` table doesn't list `conversations/` at all (same gap as `snapshots/` — see #389) so its expected lifecycle isn't documented.

## Suggested fix

- Pipe the Claude CLI agent's stdout/stderr transcript to a per-agent file in `conversations/` on container exit.
- Document the path in `references/cleanup.md` along with a retention policy (probably TTL-based, same as the snapshots fix in #389).
- Add `--conversation <agent_id>` to `kapsis` CLI so the debugging workflow can pull it without users guessing filenames.

aviadshiber:
Closes #390.

## Problem

`~/.kapsis/conversations/` was **always empty** despite being created and bind-mounted per agent. The documented post-mortem debugging workflow — *"read the transcript when an agent hangs"* — had nothing to read.

**Root cause:** The directory infrastructure existed (created, mounted, TTL-cleaned) but no code ever wrote to it. `exec "$@"` in `entrypoint.sh` had no output redirection, and the host-side container output buffer (`$container_output`) was deleted in `launch-agent.sh` without first being copied anywhere.

**Secondary bug:** `clean_conversations()` in `kapsis-cleanup.sh` called `get_file_mtime "$conv_dir"` on a *directory* — `get_file_mtime` gates on `[[ ! -f ]]` and returns empty for dirs, so the `|| continue` silently skipped every entry. Conversation directories were never cleaned up despite the 7-day TTL.

---

## Ensemble brainstorm

Three agent perspectives were consulted before settling on this design:

| **A** | `tee` before `exec` in `entrypoint.sh` — captures clean agent-only output | ✅ Adopted as **secondary/fallback** |
| **B** | Set `CLAUDE_CONFIG_DIR=/home/developer/.claude` — redirects Claude's native JSONL | ❌ Ruled out: interacts with existing `~/.claude` overlay setup; claude-cli only |
| **C** | Copy the already-buffered `$container_output` host-side before `rm -f` | ✅ Adopted as **primary** — zero extra I/O during run, agent-agnostic |

---

### `scripts/launch-agent.sh`
- **`run_agent()`** — Before `rm -f "$container_output"`: copies the already-buffered container stdout+stderr to `~/.kapsis/conversations/<agent-id>/transcript.txt`. Header records agent ID, exit code, timestamp. ANSI codes stripped. 50 MB tail-cap (bootstrap noise is at the head; the agent's actual work is at the tail).
- **`_cleanup_with_completion()`** — Saves partial transcript before deleting `_CONTAINER_OUTPUT_TMP` so SIGTERM/crash paths also produce a transcript. No-op when the normal path already wrote it.

### `scripts/entrypoint.sh`
- Secondary capture: `tee -a` before `exec "$@"` writes `agent-output.txt` to the conversations mount. Covers direct container invocations that bypass `launch-agent.sh`. Silently skipped when dir absent/unwritable — never blocks agent startup.

### `scripts/kapsis-cleanup.sh`
- `clean_conversations()` line 895: `get_file_mtime` → `get_dir_mtime`. Fixes the secondary bug so TTL-based cleanup actually fires on conversation directories.

### `docs/CLEANUP.md`
- Add **Conversations** row to the "What Gets Cleaned" table.

### `tests/test-conversation-transcript.sh` (new, 13 tests)
Covers all three artifact classes:
- Constants: TTL defined and numeric, `CONTAINER_CONVERSATIONS_PATH` correct
- Transcript write: file created, header present, content verbatim
- Transcript write: ANSI codes stripped
- Transcript write: 50 MB cap keeps **tail**, not head (regression test for truncation direction)
- Transcript write: no-op when conv dir absent or buffer empty (safety guards)
- Interrupt path: partial transcript written before buffer deletion
- Interrupt path: does not overwrite existing transcript from normal path
- Cleanup: `get_dir_mtime` works on dirs (`get_file_mtime` regression test)
- Cleanup: stale dir age exceeds TTL; fresh dir age does not

## Test plan

- [x] `bash tests/test-conversation-transcript.sh` — 13/13 pass
- [x] `bash -n` syntax check on all modified files — clean

## Known limitation

The entrypoint `tee` capture (`agent-output.txt`) may not fire for every invocation pattern — in particular, `exec 1> >(tee …) 2>&1` uses process substitution which is bash-specific and is a no-op if the conversations dir mount is absent. The primary host-side capture (`transcript.txt`) covers all cases regardless of in-container behavior.

https://claude.ai/code/session_017uh2nQsGTfyH7q4WBQE9Tj

---
_Generated by [Claude Code](https://claude.ai/code/session_017uh2nQsGTfyH7q4WBQE9Tj)_


aviadshiber (commented):
## Ensemble Review (5 independent reviewers, majority vote)

Five independent reviewers analyzed this PR; below are findings that appeared in **≥3 of 5** reviews. All five recommended **Request changes** (posted as COMMENT — GitHub blocks self-requested-changes).

### Overview

The PR fixes Issue #390 by adding two transcript-capture paths (host-side in `run_agent` + trap; in-container `tee -a` in `entrypoint.sh`) and swaps `get_file_mtime` → `get_dir_mtime` in `kapsis-cleanup.sh`. The core symptom (empty `~/.kapsis/conversations/`) and the cleanup-skip regression are correctly addressed. The cleanup one-liner is verified-correct: `get_dir_mtime` exists in `compat.sh`, constants resolve, and the documented 7-day TTL matches `KAPSIS_DEFAULT_CONVERSATIONS_TTL_DAYS`. `AGENT_ID` is validated `^[a-zA-Z0-9_-]+$` upstream so no path-traversal risk.

### Consensus findings (majority vote)

| # | Severity | Finding | Votes |
|---|----------|---------|-------|
| 1 | **MAJOR** | `bsd-sed-ansi` — `sed 's/\x1b\[[0-9;]*m//g'` uses GNU-only `\x1b`; on macOS BSD sed this is treated literally, so ANSI stripping is a **silent no-op on macOS hosts**. Reuse `_sanitize_strip_ansi` from `scripts/lib/sanitize-files.sh:272`, or use ANSI-C quoting `$'\x1b'`. — `scripts/launch-agent.sh:2617,3202` | 5/5 |
| 2 | **MAJOR** | `simulated-helpers-no-coverage` — All 13 tests exercise `_simulate_transcript_save` / `_simulate_cleanup_transcript_save` re-defined **inline in the test file**, not the production blocks. Reverting the `run_agent`/`_cleanup_with_completion` changes would leave tests green. Source `launch-agent.sh` and invoke the real functions, or test end-to-end with a stub backend. — `tests/test-conversation-transcript.sh:38-86` | 5/5 |
| 3 | **MAJOR** | `agent-output-unbounded` — In-container `tee -a "$_KAPSIS_AGENT_OUT"` has no cap, no rotation, no truncate-on-start. Across container restarts or resumed runs `agent-output.txt` grows without bound, eventually filling the per-agent volume. Unlike the host-side path it has no 50 MB cap. — `scripts/entrypoint.sh:2012-2013` | 5/5 |
| 4 | **MAJOR** | `tee-process-leak` — `exec 1> >(tee -a …) 2>&1; exec "$@"` orphans the tee subshell when bash is replaced by the agent; it is never `wait`ed on, can outlive teardown, and may lose final buffered bytes on fast-exit agents. — `scripts/entrypoint.sh:2013` | 5/5 |
| 5 | **MAJOR** | `osc-partial-ansi-strip` — Regex `\[[0-9;]*m` matches only CSI SGR. OSC sequences (titles, hyperlinks `ESC]8;…`), cursor moves (`ESC[…H/J/K`), bracketed-paste markers, and `\r` redraws (heavily used by claude-cli for progress) survive into the "clean" transcript. — `scripts/launch-agent.sh:2617,3202` | 5/5 |
| 6 | **MAJOR** | `binary-output-corruption` — `sed` runs unconditionally on the full container stdout/stderr stream; under non-UTF-8 locales sed can fail on binary bytes and the trailing `2>/dev/null \|\| true` silently drops the entire transcript. Prefix with `LC_ALL=C`. — `scripts/launch-agent.sh:2617,3202` | 4/5 |
| 7 | **MINOR** | `raw-echo-logging` — New blocks bypass `scripts/lib/logging.sh` for operational output and silently swallow errors with `2>/dev/null \|\| true`; the trap path has no logging at all on success. Project rule (`CLAUDE.md` §Code Style) is "never raw echo for operational output". | 4/5 |
| 8 | **MINOR** | `tee-blocks-on-full-volume` — If the conversations volume fills, `tee -a` blocks on write and the agent's stdout/stderr stalls (deadlock risk with no detection). The loose `[[ -w … ]]` guard only checks at startup. — `scripts/entrypoint.sh:2011-2013` | 3/5 |
| 9 | **MINOR** | `trap-redundant-dir-check` — Trap uses `${AGENT_ID:-unknown}` while normal-path uses bare `${AGENT_ID}`; on unset AGENT_ID under trap, partial transcripts pollute a shared `conversations/unknown/` namespace. Trap also re-derives `_trap_conv_dir` rather than caching the value computed at mount time. — `scripts/launch-agent.sh:2609` | 3/5 |
| 10 | **NIT** | `doc-ttl-magic-number` — `docs/CLEANUP.md:77` hardcodes "TTL 7 days" instead of referencing `KAPSIS_DEFAULT_CONVERSATIONS_TTL_DAYS`; drift risk if the default changes. | 3/5 |

### Minority finding worth surfacing

One reviewer flagged that `dashboard/ui/src/views/AgentDetail.tsx:301` still tells the user *"Kapsis provisions the directory but only populates it when the agent CLI is configured to write transcripts there"* — directly contradicts the new behavior and triggers the **Dashboard Sync Rule** in `CLAUDE.md`. Worth confirming whether that banner needs updating in this PR.

### Suggested fix priorities (before merge)

1. Fix `bsd-sed-ansi` (broken on macOS) and `simulated-helpers-no-coverage` (zero protection against future regression) — these together mean the fix is **untested on the platform where it's most likely to break**.
2. Cap or rotate the in-container `agent-output.txt`, or drop the secondary capture and rely solely on the host-side path which already has the 50 MB cap.
3. Add `LC_ALL=C` to the `sed` pipelines and broaden the regex (OSC + non-SGR CSI), or — better — reuse `_sanitize_strip_ansi`.

### Verdict

5/5 reviewers: **Request changes**. The cleanup TTL fix is ready; the transcript-capture path needs the items above before merge.

---
*Ensemble review: 5 independent reviewers (Claude), majority vote (≥3 of 5).*

---

aviadshiber:
## Ensemble review disposition (fixed in 2995fee)

All 10 consensus findings + the minority dashboard finding verified **real** and addressed. The capture pipeline was restructured around a new shared `scripts/lib/transcript.sh`, used by both the normal path and the cleanup trap:

1. **`bsd-sed-ansi`** → ESC/BEL/CR are embedded via ANSI-C quoting (`$'\x1b'`) so BSD sed gets literal bytes, never a GNU-only `\x1b` escape. (`_sanitize_strip_ansi` wasn't directly reusable — it's file-in-place and count-returning, not a stream filter — so the lib applies the same technique to the stream pipeline; rationale documented in the lib header.)
2. **`simulated-helpers-no-coverage`** → tests rewritten to source and exercise the production `transcript_save` / `transcript_save_partial` / `transcript_strip_ansi` functions; the inline simulation helpers are gone. 15/15 pass.
3. **`agent-output-unbounded`** → the in-container `tee -a` capture is removed from `entrypoint.sh` entirely (the review's own suggested option); the host-side path with its 50 MB tail-keeping cap (`KAPSIS_TRANSCRIPT_MAX_BYTES` override) is the single capture.
4. **`tee-process-leak`** → moot — no tee in the container anymore.
5. **`osc-partial-ansi-strip`** → coverage now spans CSI incl. private modes (`?2004h/l`), OSC-to-BEL (titles, OSC-8 hyperlinks), residual bare ESC bytes, and CR redraws (mid-line CR → newline so progress bars become successive lines).
6. **`binary-output-corruption`** → `LC_ALL=C` on the sed/tr stages; no blanket `|| true` swallowing the stream.
7. **`raw-echo-logging`** → operational output goes through `log_warn`/`log_debug`; failures are logged, not discarded.
8. **`tee-blocks-on-full-volume`** → moot — see 3.
9. **`trap-redundant-dir-check`** → trap and normal path share the same resolved conversations dir and agent-id fallback; partial saves no-op safely and never overwrite a completed transcript.
10. **`doc-ttl-magic-number`** → `docs/CLEANUP.md` references `KAPSIS_DEFAULT_CONVERSATIONS_TTL_DAYS`.
- **Minority (dashboard)** → `AgentDetail.tsx` empty-state copy now describes the actual behavior (transcript written when the session ends).

Validation: `shellcheck -S warning` clean on all touched scripts; `test-conversation-transcript.sh` 15/15 against production code; full `--quick` suite — the only PR-related failure was a missing exec bit on the rewritten test file (fixed, amended); the remaining local failures are this sandbox's pre-existing environmental set (no keychain/Podman), unrelated to the PR. The dashboard change is a single JSX text node; CI's dashboard job covers the build.

---

PR #396, opened 2026-06-06
test(container): add real-image hook tests + rename misleading e2e/integration labels (#381)

### LINKED ISSUE #381: test: existing 'e2e' / 'integration' tests don't exercise the container image — convert to real container-based tests
# Issue Draft (for aviadshiber/kapsis)

**Suggested title:** `test: existing "e2e" / "integration" tests don't exercise the container image — convert to real container-based tests`

**Suggested labels:** `tests`, `tech-debt`, `infrastructure`

**Body:**

---

## Problem

PR #380 fixed a bug where `scripts/lib/inject-plugin-hooks.sh` was added to the source tree (commit `8d5eea8`) and sourced from `scripts/entrypoint.sh`, but never `COPY`ed into the container image — so `install_plugin_hooks()` silently log_debug-skipped at runtime for every agent in every container build for ~6 days. The bug was completely outside the test envelope despite the repo having multiple tests with `e2e` / `integration` in their names.

The reason: **none of those tests actually build or run the container image.** They all source the lib scripts directly from the host source tree under `$KAPSIS_ROOT/scripts/lib/`.

Audit done as part of #380:

| Test file | Sources from host? | Touches image? |
|---|:-:|:-:|
| `tests/test-e2e-claude-hooks.sh` | ✅ | ❌ |
| `tests/test-integration-all-hooks.sh` | ✅ | ❌ |
| `tests/test-integration-claude-hooks.sh` | ✅ | ❌ |
| `tests/test-integration-gist-hook.sh` | ✅ | ❌ |
| `tests/test-plugin-hook-injection.sh` | ✅ | ❌ |

Smoking gun in each: `export KAPSIS_HOME="$KAPSIS_ROOT"` and `source "$LIB_DIR/<x>.sh"` against the source tree, not against `/opt/kapsis/lib/` inside a container.

Commit `5cd2435 refactor(tests): rename e2e→integration for simulated tests, real tests` acknowledged the naming-vs-reality gap and renamed most of them. But the underlying coverage gap (no real container-based assertion) wasn't addressed. The leftover `test-e2e-claude-hooks.sh` was missed by the rename, ironically.

## Why this matters beyond #380

Any future scripts/lib/ file that:

1. Gets sourced from `scripts/entrypoint.sh`
2. Doesn't get a matching `COPY` line in `Containerfile`

…will recreate the exact same silent-skip bug class. The build-time `RUN` guard added in #380 catches the specific shape "entrypoint references missing lib file." But it does NOT catch:

- A hook script in `scripts/hooks/` referenced by an installed adapter but not COPYed
- A new agent type that needs additional in-container scripts
- Permission/ownership drift (file present but unreadable by the developer user)
- Behavioral regressions (file present, script runs, but output is wrong)

A proper container-image test suite catches all of these because it asserts on **the runtime state the user actually sees**.

## Prior art (what good looks like)

Two tests already exercise the real image and should be the template for converting the rest:

- `tests/test-container-libs.sh` — original prior art, added after the analogous `status.sh`-not-in-image bug (`kapsis-DEV-209078`)
- `tests/test-container-plugin-hooks.sh` — added in PR #380 for plugin-hook injection

Both:
- Use `$KAPSIS_TEST_IMAGE` (the real built `kapsis-sandbox`)
- Use `run_in_container` / direct `podman run` to execute inside the image
- Use `skip_if_no_overlay_rw` to gracefully skip on hosts without container support
- Assert on real runtime state (file presence, function callability, settings.json contents)

## Proposed remediation

Two phases. **Default policy: when a container test covers the same functionality, delete the host-only counterpart** — don't pay CI/token cost for duplicate coverage. Only keep a host-only test if it exercises logic that the container test cannot reasonably set up (e.g., unit-level edge cases of a single lib function that would need an excessive fixture).

### Phase 1 — For each currently-fake test, add the container counterpart, then DELETE the host-only one unless it has unique coverage

| Existing host-only test | Add real-image test | Host-only test disposition |
|---|---|---|
| `test-e2e-claude-hooks.sh` | `test-container-claude-hooks.sh` — actually run the entrypoint's hook injection inside the image, then assert the final `settings.json` | **DELETE** — wholly subsumed |
| `test-integration-all-hooks.sh` | `test-container-all-hooks.sh` — all three injectors run inside image, single output `settings.json` with all expected keys | **DELETE** — wholly subsumed |
| `test-integration-claude-hooks.sh` | covered by `test-container-claude-hooks.sh` above | **DELETE** — wholly subsumed |
| `test-integration-gist-hook.sh` | `test-container-gist-hook.sh` — fixture gist input → assert `gist.txt` produced inside container | **DELETE** — wholly subsumed |
| `test-plugin-hook-injection.sh` | `test-container-plugin-hooks.sh` ✅ done in #380 | **KEEP unique parts only** — the 31KB host file covers whitelist-filter edge cases, malformed `hooks.json` handling, idempotency, etc. that the 5 container tests don't. Strip any test case that overlaps with the container suite; rename remainder to `test-inject-plugin-hooks-unit.sh` to reflect actual scope. |

Audit each "DELETE" entry before removal — if any test case in the host file genuinely covers something the new container test can't, peel it into a small `test-<x>-unit.sh` file and keep that. Default is delete, exception is keep.

Run-cost ledger after Phase 1:
- 4 host-only files removed
- 4 new container tests added (each ~3–5s warm, ~30s cold)
- 1 host-only file reduced in scope (`test-plugin-hook-injection.sh` → `test-inject-plugin-hooks-unit.sh`, smaller)
- Net: ~30–60s added to CI per cold run; some tokens reclaimed from removed duplicate tests.

### Phase 2 — Fix the naming so it stops creating false confidence

The label that's actually wrong is **e2e**, not "integration." The host-only tests under `test-integration-*.sh` legitimately test multiple components interacting (multiple injectors, multiple hooks, lib + entrypoint logic together) — that's integration testing, just at the **host level** instead of the **container-runtime level**.

What needs to change:

- `tests/test-e2e-claude-hooks.sh` — does NOT exercise the real end-to-end runtime path (no image, no entrypoint). Rename to `test-integration-claude-hooks-host.sh` (or similar) to match its actual scope, OR fold its assertions into the existing `test-integration-claude-hooks.sh` if they overlap.
- The new container-based tests (`test-container-libs.sh`, `test-container-plugin-hooks.sh`, the ones added in Phase 1) ARE the actual e2e tests — they go through the same code path the user's agent does. The `test-container-*` prefix already communicates that; keeping it is fine.
- Document the three-tier convention in `docs/TESTING.md`:
  - `test-<unit>.sh` → single-script / single-function tests
  - `test-integration-<area>.sh` → multi-component host-side integration
  - `test-container-<area>.sh` → real image / runtime / e2e

This avoids the "but we have e2e tests for this!" false confidence that led to PR #380's bug shipping, without erasing the (genuine) value of the host-side integration tests.

### Out of scope for this issue

- A separate CI workflow for full image-build-from-`Containerfile` tests (different question: should CI cache or rebuild from scratch each PR?). Tracked separately if needed.

## Acceptance criteria

- [ ] Each of the four fake-e2e/integration tests has a real-container counterpart that uses `$KAPSIS_TEST_IMAGE` and `run_in_container` / direct `podman run`
- [ ] Each new container test self-skips via `skip_if_no_overlay_rw` when container support is unavailable
- [ ] Each new container test is registered in the appropriate category in `tests/run-all-tests.sh` and in `QUICK_TESTS`
- [ ] The four existing host-only tests are renamed to drop misleading `e2e` / `integration` from their names
- [ ] `docs/TESTING.md` (or equivalent) is updated to call out the distinction between fast host-only tests and slower real-image tests, with guidance on when to write which

## References

- PR #380: fix(image): ship inject-plugin-hooks.sh in container + build-time guard
- Commit `8d5eea8`: the bug that motivated #380
- Commit `5cd2435`: the partial rename that recognized the naming gap
- `kapsis-DEV-209078`: the earlier `status.sh`-not-in-image bug that motivated `test-container-libs.sh`

aviadshiber:
Closes #381.

## Problem

The existing `test-e2e-claude-hooks.sh`, `test-integration-all-hooks.sh`, `test-integration-claude-hooks.sh`, and `test-integration-gist-hook.sh` all **sourced inject scripts from the host source tree** and never ran inside the container image. This meant a missing `COPY` line in `Containerfile` went undetected for 6 days (the bug class from PR #380 — `inject-plugin-hooks.sh` silently no-oped at runtime for every agent run).

The root issue: any future `scripts/lib/*.sh` added and sourced by `entrypoint.sh` but missing a `COPY` line falls into the same silent-no-op trap, and none of the existing "e2e" / "integration" tests would catch it.

---

## Ensemble brainstorm

Three agent perspectives were consulted before settling on this design:

| **A** (Correctness-First) | Container tests must assert functional outcomes (not just file presence) — hook commands must point to `/opt/kapsis/hooks/`, not host paths | ✅ Adopted: every container test asserts a runtime side-effect, not just `test -f` |
| **B** (Minimal CI Cost) | Combine multiple injection scenarios into one container invocation via heredoc to minimize container starts | ✅ Adopted for `test-container-all-hooks.sh` — runs 3 injectors + all assertions in one `podman run` |
| **C** (Faithful Regression) | Tests must fail if a COPY line is removed and the image is rebuilt — structural + functional assertion pair required | ✅ Adopted: each file starts with structural (file presence) then functional (settings.json mutation) |

**Synthesis**: rename host tests to `test-host-*` (don't delete — they remain fast unit/regression guards sourcing from host), add 3 container counterparts that use `$KAPSIS_TEST_IMAGE` and catch packaging regressions.

---

### New container-tier tests (3 files)

Each uses `skip_if_no_overlay_rw`, `setup_container_test`, and direct `podman run` with a bind-mounted fixture — identical boilerplate to `test-container-plugin-hooks.sh`.

**`tests/test-container-status-hooks.sh`** (5 tests):
- `inject-status-hooks.sh` exists at `/opt/kapsis/lib/` (catches missing COPY)
- lib is executable (chmod 755)
- `inject_claude_hooks` function is sourceable inside the container
- `inject_claude_hooks` produces hook commands pointing to `/opt/kapsis/hooks/` (not host paths) — this is the key assertion: if COPY is missing the function no-ops and the output is empty
- No `settings.local.json` created (#351 regression guard, now in-container too)

**`tests/test-container-gist-hook.sh`** (3 tests):
- Gist hook appears in `settings.json` as PostToolUse[0] when `KAPSIS_INJECT_GIST=true`
- Gist hook command at the injected path is executable inside the container
- Dispatching the gist hook with a git-commit tool event inside the container creates `gist.txt` with correct content — the same end-to-end path as the host-sourced `test-host-inject-gist-hook.sh` but now running against the packaged binary

**`tests/test-container-all-hooks.sh`** (4 tests):
- All three inject libs (`inject-status-hooks.sh`, `inject-lsp-config.sh`, `inject-plugin-hooks.sh`) exist and are executable in the image
- Combined injection (status + LSP + plugin) in one container invocation: asserts `.hooks`, `.lspServers`, `.enabledPlugins` all coexist in valid JSON, no `settings.local.json`
- Plugin marker command appears in the combined `settings.json`
- Idempotency: exactly one Stop entry after double injection

### Renamed host-tier tests (4 files)

Dropped the misleading `e2e` / `integration` prefixes — these tests source scripts from the host tree and do not exercise the packaged image:

| Old name | New name | Removed from QUICK_TESTS? |
|----------|----------|--------------------------|
| `test-e2e-claude-hooks.sh` | `test-host-claude-live-api.sh` | Yes (requires real credentials) |
| `test-integration-all-hooks.sh` | `test-host-inject-all-hooks.sh` | No |
| `test-integration-claude-hooks.sh` | `test-host-claude-mock-api.sh` | No |
| `test-integration-gist-hook.sh` | `test-host-inject-gist-hook.sh` | No |

Headers updated to document the tier, the rename rationale, and cross-references to their container counterparts.

### `tests/run-all-tests.sh`

- `status` category: updated to new `test-host-*` names
- `security` category: added 3 new container tests alongside `test-container-libs.sh` and `test-container-plugin-hooks.sh`
- `QUICK_TESTS`: old integration names replaced with new `test-host-*` names; 3 new container tests added (they self-skip via `skip_if_no_overlay_rw` when Podman is unavailable, exactly like `test-container-plugin-hooks.sh` already in the list)

### `docs/TESTING.md` (new)

Documents the three-tier naming convention (`test-<subsystem>.sh` / `test-host-<subsystem>.sh` / `test-container-<subsystem>.sh`), when to write each tier, coverage map, boilerplate template, fixture bind-mount pattern, and the COPY regression bug class explanation.

---

## Test plan

- [x] `bash -n` syntax check on all new and modified files — clean
- [x] `bash tests/test-host-inject-gist-hook.sh` — 5/5 pass
- [x] `bash tests/test-host-inject-all-hooks.sh` — 8/8 pass
- [x] `bash tests/test-container-status-hooks.sh` — self-skips cleanly (no Podman in this env)
- [x] `bash tests/test-container-gist-hook.sh` — self-skips cleanly
- [x] `bash tests/test-container-all-hooks.sh` — self-skips cleanly
- [ ] Full container test suite (requires Podman + built `kapsis-sandbox` image)

## Known limitation

`test-host-claude-live-api.sh` and `test-host-claude-mock-api.sh` cannot have container counterparts without the Claude CLI binary installed in the image — that would pull in a large npm dependency. These remain host-tier tests; their injection-side coverage is now handled by the container tests.

https://claude.ai/code/session_012spBmp1KAJdDf9aTeDc7FA

---
_Generated by [Claude Code](https://claude.ai/code/session_012spBmp1KAJdDf9aTeDc7FA)_


aviadshiber (commented):
## Ensemble review (majority vote)

Three reviewers with non-overlapping scopes — correctness, style/conventions, security/CI risk — were consulted. Verdicts: **COMMENT, COMMENT, APPROVE** → majority **COMMENT**. No reviewer found a blocker; the PR's design and packaging are sound. One coverage gap and four NITs are worth addressing before merge.

### Findings

**1. CONCERN — `tests/test-container-status-hooks.sh:127-132` — coverage gap on hook-script COPYs**

The "hook commands point to `/opt/kapsis/hooks/`" assertion only verifies that `inject_claude_hooks` emitted the expected path strings into `settings.json`. Those strings are hardcoded via `jq --arg` in `scripts/lib/inject-status-hooks.sh:46-47,93-126` and have no dependency on whether the underlying hook scripts exist at the path. If `COPY scripts/hooks/kapsis-status-hook.sh` or `COPY scripts/hooks/kapsis-stop-hook.sh` were removed from `Containerfile:553-554`, the test would still pass — defeating the PR's stated goal of catching missing-COPY regressions for this hook tier.

The gist test closes this gap correctly via `test_gist_hook_command_is_executable_in_container` (`test-container-gist-hook.sh:78-125`) using `[[ -x "$gist_cmd" ]]`. Suggested fix: read each hook command back out of `settings.json` and assert `-x` (or at least `-f`) on the resolved path inside the container.

**2. NIT — `tests/test-container-gist-hook.sh:184` — dispatch exit masked**

The hook dispatch is invoked with `|| true`, swallowing any non-zero exit. The downstream `[[ -f gist.txt ]]` is the real assertion, but capturing the exit code (or stderr) would meaningfully aid diagnosis when the test fails.

**3. NIT — `tests/run-all-tests.sh:78` — asymmetric QUICK_TESTS membership after rename**

`test-e2e-claude-hooks.sh` was in `QUICK_TESTS`; the renamed `test-host-claude-live-api.sh` is not. Since the test self-skips without `CLAUDE_CODE_OAUTH_TOKEN`, observable behavior is unchanged — but worth confirming the drop was intentional vs. an editing artifact. The mock-API counterpart was correctly re-added.

**4. NIT — `docs/TESTING.md` not registered in `CLAUDE.md` Documentation Map**

Every other `docs/*.md` has a row in CLAUDE.md's Documentation Map table. Add a row pointing to the new testing guide so future agents discover it.

**5. NIT — `tests/test-container-status-hooks.sh:46-58` — misleading diagnostic**

`test_inject_status_hooks_lib_executable` checks only `test -x`. If the lib is missing entirely (the actual regression case), the failure message reads "NOT_EXECUTABLE" rather than "MISSING". The preceding `_lib_exists` test covers presence so this is cosmetic — just a friendlier error string.

### What the reviewers agreed worked well

- All 3 new container test files match the established `test-container-plugin-hooks.sh` boilerplate exactly: `setup_container_test`, `KAPSIS_TEST_IMAGE` resolution, `--userns=keep-id --security-opt label=disable`, `skip_if_no_overlay_rw` guard, `mktemp -d` fixtures with `trap … RETURN` cleanup.
- Rename completeness verified — no dangling references to old `test-e2e-*` / `test-integration-*` names in CI workflows, Makefile, packaging, or docs.
- Containerfile lines 539-555 do COPY the libs/hooks the tests assert on (premise of the PR is real).
- Idempotency claim verified: `test-container-all-hooks.sh:244-249` calls `inject_claude_hooks` twice and asserts `jq ".hooks.Stop | length" == 1`.
- All new shell files pass shellcheck cleanly with `#!/usr/bin/env bash` + `set -euo pipefail`, `[[ ]]` conditionals, quoted variables, no GNU-only flags.
- Podman flags match the prior-art baseline — no privilege escalation, no `--cap-add`, no seccomp loosening. Bind mounts use mktemp'd fixtures only; no host credentials reachable.
- CI workflow rename is complete and consistent; the live-API job remains gated by `if: github.event_name != 'pull_request'` and `environment: e2e-live-api`, so the OAuth token is unreachable from fork PRs.
- Dashboard Sync Rule not triggered — test-only PR, no new CLI flags / status fields / audit events.

---

aviadshiber:
## Ensemble review disposition (fixed in d29d86a)

1. **Coverage gap — hook COPYs unverified** (CONCERN): verified real — the assertions only matched the `jq --arg` path strings. Fixed: a new `test_injected_hook_commands_are_executable_in_container` reads each hook command back out of the generated `settings.json` and asserts `-x` on the resolved path inside the container, mirroring the gist test's pattern. Removing either `COPY` from the Containerfile now fails the test.
2. **`|| true` masks dispatch exit** (NIT): fixed — dispatch rc and output are captured and included in the `GIST_NOT_CREATED` diagnostic; the file-existence check remains the real assertion (hooks may exit non-zero on benign events).
3. **QUICK_TESTS asymmetry** (NIT): **intentional, now documented in place** — `test-host-claude-live-api.sh` self-skips without `CLAUDE_CODE_OAUTH_TOKEN`, but on dev machines where the token *is* set it would make real Anthropic API calls; the quick tier must stay fast and credential-free. Its mock-API counterpart is in QUICK_TESTS. A comment above the list now records this so it doesn't look like an editing artifact.
4. **`docs/TESTING.md` not in the Documentation Map** (NIT): fixed — row added (and the stale "15 guides" count corrected to 18).
5. **Misleading NOT_EXECUTABLE diagnostic** (NIT): fixed — three-state markers (`LIB_MISSING` / `LIB_PRESENT_NOT_EXECUTABLE` / `LIB_EXECUTABLE`), deliberately non-overlapping so the substring-based asserts can't cross-match.

Validation: `shellcheck -S warning` + `bash -n` clean on all modified files; container tests skip gracefully without Podman locally — CI's Container Tests job is the verifier for the new in-container assertions.

---

PR #397, opened 2026-06-07
fix(overlay): move macOS overlay upper/work dirs off virtio-fs to VM-native named volume (#376)

### LINKED ISSUE #376: Move overlay upper/work dirs off virtio-fs to eliminate AVF cache-coherency bug class
## Context

PR #375 mitigates the **trigger** for the macOS AVF virtio-fs cache-coherency bug (Apple FB16008360, podman#23061) by serializing concurrent agent launches. That kills the same-second cluster failures but doesn't eliminate the bug class — single agents under sustained I/O can still hit it.

## Bug class (full)

Each kapsis container's `/workspace` overlay puts its `upper`/`work` directories on a host-side path under `~/.kapsis/.ai-sandboxes/<id>/` — which lives on the virtio-fs share backed by AVF. The overlay's `mkdirat`/`getxattr`/`metacopy` lookups round-trip through the AVF virtio-fs server, which has known concurrency races (documented across podman#23061, podman#24725, docker/for-mac#6690).

Symptoms inside the VM:
```
overlayfs: failed to create directory .../<id>/work/work (errno: 13); mounting read-only
overlayfs: failed to get metacopy (-2)
```

→ surfaces as `exit_code=4 error_type=mount_failure` to the agent.

## Proposed structural fix

Move overlay metadata (`upper` and `work` directories) off the virtio-fs share onto **VM-native ext4** via a Podman named volume. The `/workspace` BIND mount can stay virtio-fs (read-only or worktree-style); only the overlay-internal metadata needs to move.

### Concrete change

In `setup_overlay_sandbox` (and the equivalent in `setup_worktree_sandbox` where overlay dirs are also created for `.claude` CoW staging):

Today:
```bash
SANDBOX_DIR="${SANDBOX_UPPER_BASE}/${SANDBOX_ID}"  # = $HOME/.ai-sandboxes/... (virtio-fs)
UPPER_DIR="${SANDBOX_DIR}/upper"
WORK_DIR="${SANDBOX_DIR}/work"
```

Proposed: keep the host-visible path for debuggability, but bind it FROM a Podman named volume backed by the VM's own ext4. Container sees `/workspace`'s overlay layers on local ext4 → no AVF round-trips for overlay metadata.

Implementation sketch:
1. Create / reuse a `kapsis-overlays` named volume per agent (or shared with subpaths).
2. Mount it into the container at the same logical path the overlay backend expects.
3. Keep the `host-visible` symlink pointing at the volume's contents so existing debugging tools (kapsis-status, kapsis-cleanup, dashboard) still work.

## Why this matters

- Eliminates the **bug class** rather than the trigger — single agents under sustained I/O won't hit it either.
- Removes the dependency on the launch-lock mitigation in PR #375 (the lock would become a defense-in-depth no-op rather than the load-bearing fix).
- Aligns with Docker for Mac's chosen approach (Docker moved layer storage off virtio-fs years ago for this exact reason).

## Out of scope

- The /workspace bind itself can stay virtio-fs (it's the only way the host-side checkout reaches the container).
- Container-internal scratch (/tmp etc.) is already VM-native ext4.

## Dependencies

- PR #375 (launch serialization) — should land first as the short-term mitigation; this issue is the structural follow-up.

## Tracking

- Apple Feedback FB16008360
- podman/podman#23061
- podman/podman#24725
- docker/for-mac#6690

aviadshiber:
Closes #376.

## Problem

The AVF virtio-fs cache-coherency bug (Apple FB16008360, podman#23061) causes sporadic `overlayfs: failed to create directory ... (errno: 13)` and `overlayfs: failed to get metacopy (-2)` errors on macOS, surfacing as `exit_code=4 error_type=mount_failure` to the agent.

**Root cause:** `~/.kapsis/.ai-sandboxes/<id>/upper` and `/work` sit on the virtio-fs share. The overlay driver's `mkdirat`/`getxattr`/`metacopy` lookups round-trip through the AVF helper process, which has known concurrency races. PR #375 serialized concurrent launches (mitigation); this PR eliminates the bug class.

## Ensemble brainstorm

Three perspectives consulted before settling on the design:

| **A** (Security) | `--cap-add SYS_ADMIN` is broad — is there a narrower capability? | ❌ FUSE mounts inside rootless Podman require SYS_ADMIN; narrower alternatives don't work. Gated by `KAPSIS_OVERLAY_USE_VOLUME=false` opt-out. |
| **B** (Correctness) | After container exits, export named volume → `UPPER_DIR` via `podman volume export \| tar`, so `post_container_overlay()` works unchanged | ✅ Adopted — single tar extraction to `SANDBOX_DIR` gives exact `upper/` + `work/` layout |
| **C** (Minimal Risk) | macOS-only change; Linux path completely untouched; existing `:O,upperdir=` path preserved as fallback | ✅ Adopted — `is_macos()` guards every new code path |

### `scripts/lib/constants.sh`
- `KAPSIS_OVERLAY_VOLUME_SUFFIX="-overlay"` — volume name suffix (mirrors the `-status` pattern from Issue #276)

### `scripts/launch-agent.sh`

**`setup_overlay_sandbox()`** — on macOS, sets `OVERLAY_VOLUME=kapsis-<id>-overlay` and creates only `SANDBOX_DIR` as a post-run export target. Host `upper/`/`work/` dirs are **not** pre-created (they live in the named volume). Linux path unchanged.

**`generate_volume_mounts_overlay()`** — on macOS, mounts the project read-only as `/lower` and the named volume as `/overlay` instead of the `:O,upperdir=…,workdir=…` virtio-fs bind. `entrypoint.sh`'s existing `setup_fuse_overlay()` (new-layout branch, checks for `/overlay`) merges them into `/workspace` via fuse-overlayfs running on VM-native ext4 — zero virtio-fs round-trips on overlay metadata.

**`build_container_cmd()`** — on macOS overlay mode, adds `--device /dev/fuse --cap-add SYS_ADMIN` (required for FUSE mounts inside rootless containers).

**`generate_env_vars()`** — on macOS overlay mode, sets `KAPSIS_USE_FUSE_OVERLAY=true` so `entrypoint.sh` takes the fuse-overlayfs path.

**`export_overlay_volume_to_host()`** — new function called just before `post_container_overlay()`. Uses `podman volume export | tar` to extract the named volume's `upper/` and `work/` subtrees into `SANDBOX_DIR`, so all downstream logic (scope validation, change summary, rsync hint) works without modification.

**`cleanup_agent_volumes()`** — adds `overlay` to the macOS-only suffix list so the named volume is removed at session end.

### `tests/test-overlay-volume-sandbox.sh` (new, 9 tests)

Host-tier tests (no container launched, no Podman required):
- macOS: `OVERLAY_VOLUME` name is `kapsis-<id>-overlay`
- macOS: `SANDBOX_DIR` created, host `upper/`/`work/` **not** created
- macOS: volume mounts contain `/lower:ro` and `/overlay`, NOT `:O,` or `upperdir=`
- macOS: `/workspace` absent from volume mounts (entrypoint handles it)
- macOS: `KAPSIS_OVERLAY_USE_VOLUME=false` opt-out falls back to kernel OverlayFS
- Linux: host `upper/`/`work/` created, `OVERLAY_VOLUME` empty
- Linux: mounts contain `:O,` `upperdir=`, NOT `/lower:ro`
- Constant defined and equals `-overlay`

### `tests/run-all-tests.sh`
- New test added to `validation` category and `QUICK_TESTS`

## What is NOT changed

- `entrypoint.sh` — already has the new-layout branch (`if [[ -d "/overlay" ]]`) from prior work; no changes needed
- Linux behaviour — `is_macos()` gates every new path
- Worktree mode — untouched
- `post_container_overlay()` — unchanged; export step feeds it the same directory layout

## Test plan

- [x] `bash tests/test-overlay-volume-sandbox.sh` — 9/9 pass
- [x] `bash tests/test-input-validation.sh` — 13/13 pass (no regression)
- [x] `bash tests/test-status-reporting.sh` — 27/27 pass (no regression)
- [x] `bash -n` syntax check on all modified files — clean
- [ ] macOS end-to-end: launch agent with `--overlay-mode`, verify no exit_code=4 under concurrent load (requires macOS + Podman VM)

## Known limitation

Container-tier tests (actual fuse-overlayfs mount inside `kapsis-sandbox` image) require Podman + a built image. These are deferred to a follow-up `test-container-overlay-volume.sh` test file — the same pattern as `test-container-plugin-hooks.sh` from PR #380.

https://claude.ai/code/session_01LzNqWjJz92QmDfnxCTSFGa

---
_Generated by [Claude Code](https://claude.ai/code/session_01LzNqWjJz92QmDfnxCTSFGa)_


aviadshiber (commented):
## Ensemble Code Review — PR #397

Three independent reviewers (security, correctness, style/tests) audited this PR. The findings below are the **majority-vote consensus** (issues flagged by ≥2 reviewers), followed by single-reviewer flags worth surfacing.

The design is sound: the entrypoint-side `setup_fuse_overlay()` already supports `/overlay/upper`+`/overlay/work` (verified at `scripts/entrypoint.sh:799-802`), so a `podman volume export` of the named volume produces a tar with `upper/` and `work/` at top level that extracts cleanly into `SANDBOX_DIR` — `UPPER_DIR`/`WORK_DIR` are still assigned in the macOS branch, downstream `post_container_overlay()` works unchanged, ordering is correct (export precedes cleanup), and dry-run/Linux paths are preserved. 9 host-tier tests pass; `bash -n` clean.

---

### Majority Findings (≥2 reviewers)

**1. [MEDIUM] Named volume persists across reruns of the same `AGENT_ID`** — `scripts/launch-agent.sh:1285`, lifecycle in `main()`

Named volumes are not removed by `--rm` — only by `cleanup_agent_volumes` at session end. If a previous run crashed before cleanup, or the user reruns with `--agent-id <same>`, leftover `upper/`/`work/` content pollutes the new fuse-overlayfs upper layer. `post_container_overlay`'s change diff then sees stale files from the prior run. (Flagged by reviewers A and B.)

Suggested fix: In `setup_overlay_sandbox`, before launch, `podman volume rm -f "$OVERLAY_VOLUME" 2>/dev/null || true` (or refuse to start if it pre-exists), so each fresh run begins with a clean upper layer.

**2. [LOW] `2>/dev/null` on both ends of the export pipeline hides diagnostics** — `scripts/launch-agent.sh:3557-3559`

```bash
podman volume export "$OVERLAY_VOLUME" 2>/dev/null \
    | tar -xf - 2>/dev/null \
    || export_rc=$?
```

When the export fails, the user only sees `Overlay volume export failed (rc=N)` — no signal whether the VM was down, the volume name was wrong, or tar ran out of disk. (Flagged by reviewers A and B.)

Suggested fix:
```bash
local err; err=$(mktemp)
podman volume export "$OVERLAY_VOLUME" 2>"$err" \
    | tar -xf - -C "$SANDBOX_DIR" 2>>"$err" \
    || { export_rc=$?; log_warn "export stderr: $(cat "$err")"; }
rm -f "$err"
```

**3. [LOW–MEDIUM] Test inlines verbatim copies of production functions — drift risk** — `tests/test-overlay-volume-sandbox.sh:65-94`

The test re-declares `setup_overlay_sandbox()` and `generate_volume_mounts_overlay()` verbatim. Sibling tests like `test-vfkit-watchdog.sh:60` source the production library and only stub logging. Production changes will pass tests even if behavior diverges. (Flagged by reviewers B and C.)

Suggested fix: Extract the two functions to a small lib (e.g., `scripts/lib/overlay-sandbox.sh`) and source it from both `launch-agent.sh` and the test, matching the vfkit-watchdog/status-sync pattern. Or, source `launch-agent.sh` with a `KAPSIS_LAUNCH_AGENT_LIB_ONLY=1` guard.

---

### Single-Reviewer Notable Findings (worth surfacing)

**4. [HIGH — security only] `tar -xf -` on untrusted volume content has no symlink/path-traversal hardening** — `scripts/launch-agent.sh:3557-3559`

A buggy or compromised agent could write `/overlay/upper/foo` as a symlink to `/Users/$USER/.ssh/authorized_keys`; on host extraction, subsequent writes follow the symlink and escape `SANDBOX_DIR`. The repo's `scripts/lib/status-sync.sh:58-110` already documents this exact attack class and uses a hardened extract pattern.

Suggested fix: Extract into a fresh `mktemp -d` under `SANDBOX_DIR` with `--no-same-owner --no-same-permissions`, reject symlinks/devices via a `find -type l -delete` post-pass (or use `bsdtar --insecure=no`), then move regular files into place.

**5. [MEDIUM — security only] `--cap-add SYS_ADMIN` is not gated by `KAPSIS_SECURITY_PROFILE`** — `scripts/launch-agent.sh:2419-2426`

SYS_ADMIN is the broadest Linux capability and undermines the `strict`/`paranoid` profile contract. Inside the user namespace it still enables `mount(2)`, `pivot_root`, and overlay tricks. The PR has no path that says "I asked for paranoid but got SYS_ADMIN added back."

Suggested fix: When `KAPSIS_SECURITY_PROFILE` ∈ {strict, paranoid}, refuse `KAPSIS_OVERLAY_USE_VOLUME=true` (or warn + downgrade). Document the trade-off in `docs/SECURITY-HARDENING.md`.

**6. [MEDIUM — style/docs only] New env var undocumented** — Per CLAUDE.md, user-facing config changes should land in `docs/`. `KAPSIS_OVERLAY_USE_VOLUME` appears nowhere in `docs/`. Suggested fix: add it to `docs/CONFIG-REFERENCE.md`, mention the macOS named-volume variant in `docs/ARCHITECTURE.md`, and note the `-overlay` volume name in `docs/CLEANUP.md`.

**7. [LOW] Misleading comment** — `scripts/launch-agent.sh:3545` claims "the first path component is stripped" but no `--strip-components` is used. The behavior is correct (tar entries are already relative to volume root), only the comment is wrong.

---

### Items Verified Clean

- `KAPSIS_USE_FUSE_OVERLAY` env-var wiring → `entrypoint.sh:782` ✓
- `UPPER_DIR`/`WORK_DIR` still assigned in macOS branch → downstream code works ✓
- Ordering: export at `:3224` precedes cleanup at `:3235` ✓
- Dry-run honored, Linux path unchanged ✓
- Volume suffix matches cleanup name ✓
- Style adherence (4-space indent, `[[ ]]`, quoted vars, log_* helpers) ✓
- `bash -n` syntax clean on all 4 files ✓

---

**Recommendation:** address findings 1, 2, and 4 before merging (#1 is a real footgun, #4 is the only security-blocking concern surfaced). #3, #5, #6, #7 are quality follow-ups that can land in this PR or a follow-up at the author's discretion.


---

aviadshiber:
## Ensemble review response

All 7 findings verified against HEAD — all REAL — and addressed in `b77a24d` (code/tests) + `0f17eb5` (docs):

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | Named volume persists across reruns of same AGENT_ID | **Fixed in b77a24d** — `setup_overlay_sandbox` now checks `podman volume exists` and `volume rm --force`s a leftover `kapsis-<id>-overlay` before launch, so every session starts with a clean upper layer (covers crashed runs and `--keep-volumes`) |
| 2 | `2>/dev/null` on both ends of the export pipeline | **Fixed in b77a24d** — stderr captured to a mktemp file and surfaced via `log_warn` on non-zero rc |
| 3 | Test inlines verbatim copies of production functions | **Fixed in b77a24d** — extracted to `scripts/lib/overlay-sandbox.sh`, sourced by both `launch-agent.sh` and the test (vfkit-watchdog/status-sync pattern); suite grew 9 → 17 tests |
| 4 | `tar -xf -` lacks symlink/path-traversal hardening | **Fixed in b77a24d** — extraction into a fresh staging dir with `--no-same-owner --no-same-permissions`, then sanitized (device/fifo/socket nodes dropped; absolute and `..`-escaping symlinks stripped, safe intra-tree relative links like `node_modules/.bin` preserved), then moved into `SANDBOX_DIR`. Unit-tested incl. a hostile-symlink end-to-end export |
| 5 | `--cap-add SYS_ADMIN` not gated by security profile | **Fixed in b77a24d** — `resolve_overlay_volume_mode` warns and downgrades to kernel OverlayFS under `strict`/`paranoid` (SYS_ADMIN is never added back); trade-off documented in SECURITY-HARDENING.md §2.3.1 |
| 6 | `KAPSIS_OVERLAY_USE_VOLUME` undocumented | **Fixed in 0f17eb5** — CONFIG-REFERENCE.md (new section + env var), ARCHITECTURE.md (named-volume variant under Overlay Mode), CLEANUP.md (`-overlay` in the per-agent volume table) |
| 7 | Misleading "first path component is stripped" comment | **Fixed in b77a24d** — comment rewritten; tar entries are already relative to the volume root |

Side effect worth noting: the lib extraction grew the `launch-agent.sh` header past line 250, which exposed a brittle `head -250` usage check in `tests/test-volume-cleanup.sh`; it now anchors on the `usage()` function body (same commit).

Validation: `shellcheck -S warning` clean on all modified scripts; `test-overlay-volume-sandbox.sh` 17/17; `test-volume-cleanup.sh` 10/10; quick-suite failures match the pre-existing no-Podman/no-keychain environment baseline. Dashboard check: the disk store already enumerates volumes via the generic `name=kapsis-` filter, so the `-overlay` volume is covered without dashboard changes.

https://claude.ai/code/session_01GyUwb9P9abPA2WnUAqnCUU

---

aviadshiber:
## SYS_ADMIN removed — overlay now constructed by the Podman runtime (`fac5a56`)

Per the security requirement, the macOS overlay-volume mode no longer mounts fuse-overlayfs inside the agent container. It now uses the **same mechanism as the existing Linux overlay path**: a podman overlay volume mount

```
-v "$PROJECT:/workspace:O,upperdir=<vol>/upper,workdir=<vol>/work"
```

with `<vol>` = the per-agent named volume's VM-side mountpoint. The overlay is assembled by the Podman runtime in the VM's rootless namespace **before the agent starts**, so:

- **`--cap-add SYS_ADMIN` and `--device /dev/fuse` are gone** — the agent container keeps the full cap-drop posture. This closes the CVE-2022-0185-class `fsconfig()` escape surface that a namespaced SYS_ADMIN grant exposes.
- The mode is now available under **all security profiles, including `strict`/`paranoid`** — the warn-and-downgrade gate is removed.
- Reliability properties are unchanged: upper/work (the overlay metadata that triggers the AVF virtio-fs bug, FB16008360 / podman#24725) stay on VM-native ext4; stale-volume reset and the hardened `podman volume export` recovery path are untouched.
- New: `_prepare_overlay_volume` pre-creates `upper/` + `work/` inside the volume via a throwaway container (podman does not `MkdirAll` custom overlay dirs).

Floor: Podman ≥ 4.0 (overlay mounts with custom `upperdir`/`workdir`, containers/podman#12712).

Validation: shellcheck `-S warning` clean; `test-overlay-volume-sandbox.sh` 21/21 (incl. new assertions that no cap-add/device flags are emitted and that strict/paranoid don't downgrade); `test-volume-cleanup.sh` 10/10. **One real-macOS smoke run is still recommended** before relying on the mode: `upperdir` pointing at the volume's VM-side `_data` path is interpreted server-side in the VM and can't be exercised in CI.

---

PR #398, opened 2026-06-08
feat(preflight): VM memory sizing advisor + host memory pressure gate (#377)

### LINKED ISSUE #377: Kapsis-managed VM memory sizing (config-driven, not manual `podman machine set`)
## Context

The macOS AVF virtio-fs cache-coherency bug (Apple FB16008360, mitigated by PR #375 / structural fix tracked in companion issue) is **amplified** by host memory pressure: when the host has near-zero free RAM + heavy swap, jetsam throttles the AVF helper process backing the VM's virtio-fs server, widening the cache race window and increasing failure frequency.

Currently the user's only knob is `podman machine set --memory N && podman machine restart` — a manual, disruptive operation that requires killing in-flight agents.

## Ask

Make VM memory sizing a kapsis configuration concern, not a manual `podman machine` workaround.

## Suggested design

1. **Preflight advisor** — `kapsis preflight` already exists; extend it to:
   - Read host total RAM (sysctl on macOS, /proc/meminfo on Linux).
   - Read current VM allocation (`podman machine inspect`).
   - Read concurrent-agent intent from agent-sandbox.yaml (e.g., a new `max_parallel_agents` config key).
   - Warn / error if the VM is sized below a sane threshold for the planned concurrency, e.g. `VM mem < (1 GB base + 0.5 GB per parallel agent)`.

2. **Auto-tune (opt-in)** — config key `vm_memory: auto` in the global kapsis config would have preflight calculate a recommended VM size based on (host_total_RAM × headroom_factor) and apply it via `podman machine set --memory` before launch. Requires `podman machine stop && start` so should only run when no agents are active (kapsis already tracks active-agent count for autoheal).

3. **Preflight gate on jetsam pressure** — on Darwin, read `memory_pressure -l warn` or `vm_stat` swap usage; if host is in elevated pressure state, refuse new launches (or warn) with a clear message linking to this issue and the AVF bug context.

## Why kapsis-side and not user-managed

- Manual `podman machine set` requires stopping the VM, killing all in-flight agents.
- Different workloads (parallel agents, large repos, mvn-cache pressure) want different VM sizes.
- The amplifier-vs-trigger distinction is non-obvious to users; kapsis is the right layer to encode it.

## Dependencies

- Companion: PR #375 (short-term launch serialization mitigation).
- Companion: structural overlay-off-virtio-fs issue (eliminates the bug class entirely; this memory-config issue becomes nice-to-have once the structural fix lands).

## Notes

User's specific case: M4 Pro 16 GB host running a 16 GiB Podman VM = the AVF helper is jetsam's #1 target. Halving the VM to 8 GiB would significantly drop failure frequency on this hardware, but the right answer is making kapsis aware of host capacity rather than asking the user to remember the formula.

aviadshiber:
Closes #377.

## Problem

macOS Podman VM memory is configured once at `podman machine init` time and never revisited. When the VM is under-sized relative to actual workload (especially parallel agents), or when it consumes too much of the host RAM, the AVF helper process becomes jetsam's top eviction candidate — widening the virtio-fs cache-coherency race window (Apple FB16008360) and increasing `mount_failure` (exit_code=4) frequency.

The user's only knob was a manual, un-guided `podman machine set --memory N` invocation. There was no guidance on what N should be, no connection to Kapsis workload intent, and no warning when the current size was wrong.

## Ensemble brainstorm

Three perspectives consulted before settling on this design:

| **A** (Correctness) | Auto-tune via `podman machine set` at launch time is disruptive — requires VM restart, kills all in-flight agents | ❌ Defer auto-tune to a follow-up; advisor-only for this PR |
| **B** (Minimal scope) | Warn-only + print exact remediation command; macOS-only; silent no-op on Linux | ✅ Adopted |
| **C** (Integration/Architecture) | Follow `check_disk_space()` pattern exactly — env-var overrides, `preflight_warn`, constants in `constants.sh` | ✅ Adopted |

**Synthesis**: advisor + jetsam gate in this PR; auto-tune (`vm_memory: auto`) deferred.

### `scripts/lib/constants.sh`
Four new constants for the advisor formula (all env-var-overridable):
- `KAPSIS_DEFAULT_VM_BASE_MEMORY_GB=2` — VM OS + Podman daemon baseline
- `KAPSIS_DEFAULT_VM_PER_AGENT_MEMORY_GB=3` — per-agent process + toolchain heap
- `KAPSIS_DEFAULT_VM_MAX_HOST_PCT=80` — VM:host ratio above which AVF helper is jetsam's #1 target
- `KAPSIS_DEFAULT_VM_SWAP_WARN_PCT=50` — elevated host swap threshold

### `scripts/preflight-check.sh`

**`check_podman_vm_memory(config_file)`** (macOS only):
- Reads `vm.max_parallel_agents` from `agent-sandbox.yaml` (or `KAPSIS_MAX_PARALLEL_AGENTS` env)
- Computes recommended minimum: `base_gb + per_agent_gb × max_parallel_agents`
- **WARN 1**: VM memory below threshold → prints exact `podman machine set --memory N` command with stop+start
- **WARN 2**: VM > max_host_pct of host RAM → jetsam amplifier warning with safe recommended size
- Silent no-op when VM memory cannot be read (missing machine, stopped machine, parse failure)
- Linux: returns 0 immediately (no VM to size)

**`check_host_memory_pressure()`** (macOS only):
- Reads `sysctl vm.swapusage` for current swap usage
- **WARN**: swap ≥ KAPSIS_DEFAULT_VM_SWAP_WARN_PCT → elevated pressure message linking to Apple FB16008360
- Silent when no swap is configured, or when `sysctl` is unavailable (Linux, older macOS)

Both checks wired into `preflight_check()` immediately after `check_podman()`.

### `agent-sandbox.yaml.template`
New `vm:` section documents `max_parallel_agents` with the formula in a comment.

### `tests/test-preflight-vm-memory.sh` (new, 17 tests)

| Group | Tests |
|-------|-------|
| Constants | 4 — each constant defined and numeric |
| Linux no-op | 2 — both checks return without warnings on Linux |
| VM memory advisor | 5 — below threshold, jetsam risk, remediation command content, agent scaling, adequate VM |
| Memory pressure | 4 — high swap warns, low swap ok, no swap ok, unreadable VM skipped |
| Wiring | 2 — both checks present in `preflight_check()` |

All 17 pass. Existing `test-preflight-check.sh` (23 tests) and `test-disk-space-guard.sh` (10 tests) — no regressions.

## Sample output (macOS, 4GB VM, 16GB host)

```
[WARN] VM memory 4GB < recommended 5GB for 1 parallel agent(s)
[WARN]   AVF virtio-fs cache race window widens under memory pressure (Apple FB16008360)
[WARN]   To resize (requires VM restart — kills in-flight agents):
[WARN]     podman machine stop podman-machine-default && podman machine set --memory 5120 podman-machine-default && podman machine start podman-machine-default
```

## Out of scope (deferred)
- `vm_memory: auto` config key for opt-in auto-tune — requires VM stop/start, tracked separately
- `kapsis preflight --tune-memory` standalone command

https://claude.ai/code/session_01HTB5aL6vQdEbx7Pb59CzPP

---
_Generated by [Claude Code](https://claude.ai/code/session_01HTB5aL6vQdEbx7Pb59CzPP)_


aviadshiber (commented):

Three independent reviewers focused on **(A) correctness**, **(B) project conventions**, **(C) security / edge cases**. Findings where ≥2 reviewers converged are tagged **[majority]**; the rest are notable singletons.

---

### Majority findings

**1. [majority] Swap-pressure heuristic is fragile** — `scripts/preflight-check.sh:558-585`

Three angles converged on the same code:
- The awk parser truncates fractional MB → sub-MB swap (`total = 0.50M`) yields `0` and hits the "No swap configured" branch instead of "pressure: low" (Reviewer A).
- No unit/suffix handling — if Apple ever emits `G` or `K` in `vm.swapusage` (or mixes units between `used` and `total`), the percentage math silently goes wrong (Reviewer A).
- Percentage-only signal produces false positives on large-RAM Macs with tiny dynamic swap files — a 600 MB swap at 51% used fires the same warning as a host that's actually thrashing (Reviewer C).

Suggest: add an absolute floor (e.g., skip when `swap_used_mb < ~512`), explicitly match `M` and log_debug otherwise, or use a more reliable signal (`memory_pressure -l warn`, `vm.compressor_pageouts`).

**2. [majority] Silent fallback when `yq` is missing or YAML key is absent** — `scripts/preflight-check.sh:477-482`

If `yq` isn't installed, or the wrong flavor is installed (mikefarah vs python-yq vs kislyuk), or the YAML key is mistyped, the function silently uses `max_parallel_agents=1`. A user who set `vm.max_parallel_agents: 8` then gets no warning about their under-sized VM, with no signal that their setting was ignored. Add a `log_debug` (or better, `log_info`) noting the resolved value and the source — env / yaml / default.

---

### Notable singletons (high confidence)

**3. [correctness] Check 1 early-return swallows Check 2 and the success log** — `scripts/preflight-check.sh:508-516`

When `vm_mem_gb < recommended_gb`, the function emits the sizing warning and `return 0`. The jetsam check (Check 2) and the `preflight_ok "VM memory OK …"` branch are both unreachable on that path. The function header at lines 449-454 implies the two checks are independent. Either run both in sequence (combining warnings) or document Check 1's priority in the docstring.

**4. [correctness] VM memory comparison loses up to 1 GiB to integer truncation** — `scripts/preflight-check.sh:497, 504, 508`

`vm_mem_gb = vm_mem_mb / 1024` truncates: a 5119 MiB VM rounds to 4 GiB and falsely trips the `4 < 5` threshold. Same class of issue for `host_mem_gb` (line 504), which feeds the jetsam ratio. Comparing in MiB end-to-end (`vm_mem_mb < recommended_gb * 1024`) eliminates the entire boundary issue.

**5. [robustness] `KAPSIS_VM_*` numeric env vars are not validated** — `scripts/preflight-check.sh:469-471`

`KAPSIS_MAX_PARALLEL_AGENTS` and the YAML field get `=~ ^[0-9]+$` guards (lines 475, 480), but `KAPSIS_VM_BASE_MEMORY_GB`, `KAPSIS_VM_PER_AGENT_MEMORY_GB`, `KAPSIS_VM_MAX_HOST_PCT` go straight into `$(( base_gb + per_agent_gb * max_parallel_agents ))`. Bash arithmetic expansion is recursive — a misconfigured CI/profile script setting `KAPSIS_VM_BASE_MEMORY_GB='x[$(touch /tmp/pwn)]'` (with any `x` in scope) executes under `set -euo pipefail`. Self-pwn boundary, but trivial to harden — apply the same regex guard already used two lines below.

**6. [docs] No entries in `docs/CONFIG-REFERENCE.md`** for the new `vm:` section or the 4 `KAPSIS_VM_*` override env vars. Compare the `liveness:` template (CONFIG-REFERENCE.md:893-919). The Dashboard Sync Rule does NOT apply (`status.sh` / `audit.sh` untouched in this PR) — but CONFIG-REFERENCE is the canonical user-facing surface for new config keys.

**7. [convention] `vm:` not validated by `scripts/lib/config-verifier.sh`**

Every peer section (`liveness:`, `git.auto_push:`, `security.userns:`, `resources:`, `lsp_servers:`) has a corresponding `validate_*_config` function. A typo'd `max_paralel_agents` silently falls back to 1 with no warning. Modeling on `validate_liveness_config` (lines 590-670) would be the consistent fix.

**8. [nit] `>` vs `>=` boundary inconsistency**

`check_podman_vm_memory` uses `vm_pct > max_host_pct` (line 522); `check_host_memory_pressure` uses `swap_pct >= swap_warn_pct` (line 580). The constant comment says "above which", suggesting `>`, but the swap check uses `>=`. Pick one.

---

### Verdict

Core advisor design is sound. Test coverage is excellent (17 tests, well-isolated subshell mocks). Project conventions are followed throughout — naming (`KAPSIS_DEFAULT_*`/`KAPSIS_*` override pattern matches `KAPSIS_DEFAULT_VFS_*`), logging via `lib/logging.sh`, arithmetic patterns, `is_macos` from `compat.sh` all match neighbors. ShellCheck is clean at project severity.

Items I'd treat as **must-address before merge**: majority finding #1 (swap heuristic at least needs the absolute floor), #2 (silent yq fallback), and singleton #3 (early-return semantics). The rest are reasonable polish for follow-up.


---

aviadshiber:
## Ensemble review disposition (fixed in 469e068)

All 8 findings verified **real** and addressed:

1. **Swap heuristic fragility** → anchored `M`-suffix awk field parser (fractional MB rounded up instead of truncated to 0; unexpected units ignored rather than misparsed), plus an absolute floor `KAPSIS_VM_SWAP_FLOOR_MB` (default 512 MB, in `constants.sh`) gating the percentage signal so tiny dynamic swapfiles no longer false-positive.
2. **Silent yq/YAML fallback** → resolved `max_parallel_agents` is now logged with its source (`env` / `config` / `default`); non-numeric values warn instead of being silently ignored; missing `yq` logs that the config key wasn't consulted.
3. **Check 1 early-return swallowing Check 2** → the jetsam-ratio check now runs even when the sizing warning fires; warnings combine.
4. **GiB truncation** → all comparisons are MiB end-to-end (`vm_mem_mb < recommended_gb * 1024`).
5. **Unvalidated `KAPSIS_VM_*` env vars before arithmetic** → all numeric overrides go through `_resolve_vm_numeric_override` (regex-guarded, warn-and-default on garbage), closing the arithmetic-expansion injection path.
6. **CONFIG-REFERENCE gap** → `vm:` section + all `KAPSIS_VM_*` overrides documented (modeled on the `liveness:` template), and `agent-sandbox.yaml.template` updated.
7. **No config-verifier coverage** → `validate_vm_config` added to `config-verifier.sh` with positive-integer validation and unknown-key warnings.
8. **`>` vs `>=` inconsistency** → strict `>` used for both percentage thresholds, matching the "above which" comments.

Validation: `shellcheck -S warning` clean on all touched files; `test-preflight-vm-memory.sh` 27/27 (up from 17); `test-config-verifier.sh` 29/29; `config-verifier.sh --all --test` green; full `--quick` suite green.

---

Issue #266, opened 2026-04-20
kapsis-ctl: Go helper binary for container lifecycle operations (strangler-fig)


aviadshiber:
## Problem

Several reliability issues trace back to Kapsis managing containers via bash shelling out to `podman` CLI over SSH:

1. **Podman SSH tunnel instability** — after Mac reboot/sleep, SSH breaks and all container operations fail
2. **`kapsis list` stale state** — shows containers that don't exist in podman
3. **`podman rm -f -t 0` hangs** — containers stuck in "Stopping" because the SSH-based kill can't signal the VM process
4. **No proper container inspect** — bot has to call `podman ps` directly instead of going through Kapsis

## Proposal: Strangler-Fig Pattern

Create a Go binary (`kapsis-ctl`) that talks directly to the Podman socket (libpod API over Unix socket), bypassing SSH entirely. Migrate container lifecycle operations one at a time:

### Phase 1 — Container queries (low risk)
- `kapsis-ctl inspect <agent-id>` → container state, PID, created time
- `kapsis-ctl list` → authoritative container list from podman API
- `kapsis-ctl alive <agent-id>` → boolean liveness check

### Phase 2 — Container management
- `kapsis-ctl stop <agent-id>` → graceful stop with escalation (SIGTERM → SIGKILL)
- `kapsis-ctl logs <agent-id>` → stream container logs
- `kapsis-ctl cp <agent-id> <src> <dst>` → copy files out of container

### Phase 3 — Container creation (largest change)
- `kapsis-ctl run` → replace the podman run command construction in bash

The bash orchestration logic stays in `launch-agent.sh` but calls `kapsis-ctl` for container operations instead of `podman` directly.

## Benefits
- Eliminates SSH tunnel class of bugs entirely
- Single static binary, easy to distribute
- Testable container operations
- Bot can call `kapsis-ctl inspect` instead of raw `podman ps`

## Context

Proposed during a brainstorm session analyzing 4 failed/stuck agents. The architect proposed this as a middle ground between "rewrite Kapsis in Go" (too risky) and "keep shelling out to podman" (known fragile).

PR #400, opened 2026-06-09
feat(kapsis-ctl): Phase 2 — stop, logs, cp subcommands (issue #266)

aviadshiber:
## Summary

Implements Phase 2 of the `kapsis-ctl` strangler-fig plan from issue #266. All three new subcommands bypass the Podman SSH tunnel entirely by talking to the libpod REST API over the Unix socket — which is the root cause of the hanging `podman rm -f -t 0` and stale-state bugs.

### New subcommands

**`stop [-t N] <name>`**
- `POST /v5.0.0/libpod/containers/{name}/stop?t=N` over Unix socket
- Server sends SIGTERM, waits N seconds (default 10), then SIGKILL — the escalation happens server-side so the SSH tunnel is never involved
- Context deadline is set to `grace_period + 15s` so the HTTP response arrives before the context is cancelled
- `ErrAlreadyStopped` (HTTP 304) is treated as success (exit 0)
- Exit codes: `0` = stopped, `1` = error, `3` = not found

**`logs [-f] [-n N] [--since TS] <name>`**
- `GET /v5.0.0/libpod/containers/{name}/logs` over Unix socket
- Demultiplexes Docker's 8-byte framed stream format: stdout frames → `os.Stdout`, stderr frames → `os.Stderr`
- `-f` (follow) uses `signal.NotifyContext` so Ctrl-C / SIGTERM cleanly closes the HTTP connection
- Non-TTY containers (standard kapsis agent run) use the multiplexed format; TTY containers emit raw bytes
- Exit codes: `0` = success, `1` = error, `3` = not found

**`cp <name>:<src> <dst>`**
- `GET /v5.0.0/libpod/containers/{name}/archive?path=<src>` returns a tar stream
- Handles both raw and gzip-compressed archives (auto-detected via magic bytes)
- **Zip-slip protection**: rejects absolute tar entry paths and `../`-traversal entries that would escape `<dst>`
- Exit codes: `0` = success, `1` = error, `3` = not found

### Internal changes

- Extracted `doRequest(method, url)` helper; `get()` delegates to it
- Added `ErrAlreadyStopped` sentinel error
- Added `LogsOptions` struct and `DemuxLogs` (exported for use from `main`)
- Added `validateContainerPath` for URL query-parameter hygiene (rejects null bytes, newlines)
- No external dependencies added — stdlib-only module preserved (CI `cache: false` stays correct)

### Tests

36 new test cases; all 46 tests pass, `go vet` clean:

| Suite | Tests |
|---|---|
| `TestStop_*` | success, alreadyStopped, notFound, serverError, timeoutPassedToServer, negativeTimeoutOmitsParam |
| `TestDemuxLogs_*` | stdoutFrame, stderrFrame, mixedFrames, unknownStreamTypeDiscarded, emptyStream, truncatedHeader |
| `TestLogs_*` | returnsMuxStream, notFound, queryParamsPassedToServer |
| `TestCopyFromContainer_*` | singleFile, nestedFiles, gzipCompressed, notFound, zipSlipParentPath, zipSlipAbsolutePath |
| `TestValidateContainerPath` | absolute, relative, root, empty, null byte, newline, carriage return |

## Test plan

- [ ] `make all-ctl` passes (`vet`, `test`, `build`)
- [ ] `./bin/kapsis-ctl stop --help` shows usage
- [ ] `./bin/kapsis-ctl logs --help` shows usage
- [ ] `./bin/kapsis-ctl cp --help` shows usage
- [ ] CI `go-build` job green (runs on every push, no path gate)

## References

- Issue #266 — kapsis-ctl strangler-fig plan
- Phase 1 (inspect, list, alive) — already shipped on `main`

https://claude.ai/code/session_016jDro2JCW7uiAg4pqsxe4o

---
_Generated by [Claude Code](https://claude.ai/code/session_016jDro2JCW7uiAg4pqsxe4o)_


aviadshiber (commented):
## Ensemble Code Review — Majority Vote

Three independent reviewers audited this PR; the findings below are the **consensus issues** (each flagged by ≥2 reviewers). Single-reviewer findings are listed separately at the end without weight.

### Scope
Phase 2 of `kapsis-ctl` — adds `stop`, `logs`, `cp` subcommands talking to libpod over a Unix socket. Stdlib-only. +982 / −10 across 3 files. Test suite well-structured and grew from 10 → 46 tests.

### Overall
Solid PR with good test coverage of the happy paths and the two explicit zip-slip vectors. The consensus issues are concentrated in `extractTar` (file/dir permission handling, response-body size cap) and the `logs` command (TTY handling, signal handling). None of the consensus findings is a release blocker, but several have user-visible impact.

---

### Consensus Findings (Majority Vote)

#### Bugs

**1. 16 MB `LimitReader` silently truncates `cp` archives — 3/3 reviewers** &nbsp; `podman/client.go` (`CopyFromContainer`)
```go
return extractTar(io.LimitReader(resp.Body, maxResponseBodyBytes), hostDest)
```
`maxResponseBodyBytes = 16 MB` was sized for JSON inspect/list responses. Applied to a tar stream it silently truncates any container directory larger than 16 MB — extraction either errors with a confusing tar EOF or, worse, succeeds for the first entries and drops the rest. The cap is *also* applied **before** gzip decompression, so a malicious 16 MB gzipped archive can decompress to many GB during extraction (decompression-bomb vector). Recommend: remove the cap for `cp` and instead bound *per-entry* `io.Copy` to a configurable limit; or stream with a much higher tar-specific cap (e.g. 10 GB).

**2. `extractTar` forces owner-rwx onto every extracted file and directory — 3/3 reviewers** &nbsp; `podman/client.go`
```go
case tar.TypeDir:
    os.MkdirAll(target, os.FileMode(hdr.Mode)&0o777|0o700)
case tar.TypeReg, tar.TypeRegA:
    os.OpenFile(target, ..., os.FileMode(hdr.Mode)&0o777|0o200)
```
The `| 0o700` / `| 0o200` masks unconditionally add owner permissions. Effects:
- A read-only file (0o444) lands on disk as 0o644 (gains owner-write).
- World-writable bits in a malicious tar entry are *preserved* (only the low 9 bits are masked, no clear of group/other write).

If the goal is "make sure we can re-open the file we just wrote," it's better to apply the mask only to the `OpenFile` call and `os.Chmod` to the archive's intended mode after the write completes.

**3. TTY containers will produce garbled output — 2/3 reviewers** &nbsp; `main.go` (`cmdLogs`)
The Logs doc comment correctly notes that `--tty` containers stream raw bytes without 8-byte frame headers, but `cmdLogs` unconditionally calls `DemuxLogs`. Against a TTY container the first byte is mis-interpreted as `streamType` and the next 7 as a frame size, so output is garbled or dropped. Either inspect the container first (`/inspect` → `Config.Tty`) or expose a `--tty` flag.

**4. Follow mode ignores SIGHUP — 2/3 reviewers** &nbsp; `main.go` (`cmdLogs`)
```go
ctx, cancel = signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
```
A terminal disconnect (parent shell exits) does not cancel the context, so the streaming connection and goroutines leak until the server closes them. Add `syscall.SIGHUP`.

**5. `cmdStop` deadline math is brittle — 2/3 reviewers** &nbsp; `main.go:cmdStop`
```go
ctx, cancel := context.WithTimeout(context.Background(), time.Duration(gracePeriod+15)*time.Second)
```
- A user-supplied `-t -1` (or any negative value, which `Client.Stop` documents as "use server default") produces a 14 s context deadline — small but accidental.
- A user-supplied `-t -100` yields a negative duration → already-expired context → request fails immediately even though the API call would have succeeded.
- The +15 s margin can also be tight on macOS where the SSH socket forwarder adds latency.

Clamp negative timeouts and consider a longer/configurable margin.

#### Security

**6. Symlink and hardlink tar entries silently skipped — 2/3 reviewers** &nbsp; `podman/client.go:extractTar` (`default:` branch)
Symlinks/hardlinks/devices fall through to `default` with a "skip silently" comment. The zip-slip *write* path is safe today, but:
- Users have no way to know their `cp` is incomplete.
- A future maintainer adding symlink support to this branch will not get the zip-slip prefix check that regular files receive.
Recommend returning an error (or at minimum a stderr warning) for unsupported entry types so this never silently happens.

#### Nits / Cleanup

**7. `tar.TypeRegA` is deprecated — 3/3 reviewers** &nbsp; `podman/client.go`
The `archive/tar` package normalizes `TypeRegA` to `TypeReg` internally; the extra case is dead and triggers staticcheck. Drop it.

**8. `DemuxLogs` compares EOF with `==` instead of `errors.Is` — 2/3 reviewers** &nbsp; `podman/client.go:DemuxLogs`
```go
if err == io.EOF || err == io.ErrUnexpectedEOF { return nil }
```
Bare-reader contract returns these unwrapped, so this works today, but `errors.Is` future-proofs against any wrapping intermediary (gzip, HTTP body decoders) and matches the style used elsewhere in the file.

#### Test Gaps

**9. No follow-mode cancellation test — 2/3 reviewers**
No test exercises `Logs(Follow: true)` with a cancelled context, so neither the SIGHUP omission nor any future "Ctrl-C exits 1 not 0" regression would be caught.

**10. No symlink / hardlink tar entry test — 2/3 reviewers**
The two zip-slip tests cover `../` and absolute paths but not the classic symlink-write-through-symlink escape, nor the silent-skip behaviour for non-regular entries.

**11. No archive-larger-than-16-MB / decompression-bomb test — 2/3 reviewers**
Neither the silent-truncation case nor the post-decompression amplification case has coverage.

**12. No mode-preservation / intermediate-dir test — 2/3 reviewers**
`TestCopyFromContainer_singleFile` and `_nestedFiles` only assert content, not modes or that intermediate directories `a/`, `a/b/` were actually created with reasonable permissions — so the mode-clobbering bug above is invisible to the suite.

---

### Single-Reviewer Findings (Not Consensus, FYI)

These reached only one reviewer; weigh as you see fit:

- HTTP 304 is reportedly not returned by libpod's `stop` endpoint (Docker compat API returns it; libpod returns 204) — `ErrAlreadyStopped` may be unreachable against a real Podman server.
- `cp` semantics diverge from `docker cp`/`podman cp`: those tools place a single-file source at `dest` when `dest` does not exist; this implementation always treats `dest` as a directory.
- `usage()` text omits `--since` for `logs` and exit code `2` for inspect/stop/logs/cp.
- `Logs` non-follow mode has a hard 30 s timeout that truncates large `-n 100000` dumps.
- `extractTar` creates `destDir` with mode 0o750 *before* zip-slip checks — left behind on extraction failure.
- No bound on per-frame `size` field in `DemuxLogs` (4 GB max trusted from network).

---

### Approach

Three reviewers were spawned in parallel with independent prompts (one general, one security-focused, one pragmatic). Each produced a structured findings list; I canonicalized titles and matched by file + topic. Findings appearing in ≥2 lists are surfaced above as consensus; outliers are listed transparently below but not weighted.



---

aviadshiber:
## Ensemble review disposition (fixed in 8e6f27f)

All 12 consensus findings verified **real** and fixed, each with test coverage (56 tests/subtests, `-race` clean):

1. `cp` 16 MB LimitReader truncation → cap removed from the tar stream; 10 GiB post-decompression limit enforced per-entry and cumulatively in `extractTar` (+ gzip-bomb test)
2. `extractTar` forced owner-rwx → files created 0600 / dirs 0700 during extraction; exact archive mode (`&0o777`) restored via chmod, dirs children-first (+ mode-preservation test)
3. TTY containers garble `logs` → `cmdLogs` inspects first; `Config.Tty` → raw `io.Copy`, non-TTY → `DemuxLogs`
4. Follow mode ignores SIGHUP → added to `NotifyContext`
5. Negative `-t` deadline math → clamped to libpod's 10 s default; margin widened 15 s → 30 s
6. Symlinks/hardlinks silently skipped → per-entry warning to stderr; unvalidated `Linkname` caveat documented
7. Dead `tar.TypeRegA` case → removed
8. `EOF` compared with `==` → `errors.Is`
9–12. Missing tests (follow-cancellation, link-skip warning, >16 MB + decompression bomb, mode/intermediate-dir preservation) → all added

Single-reviewer items also addressed: `usage()` now documents `--since` and exit code 2; non-follow logs timeout 30 s → 5 min; `cp` host-dest-dir semantics documented in the command doc + usage.

**False positives (no change, with reasoning):**
- *"libpod stop never returns 304, so `ErrAlreadyStopped` is unreachable"* — libpod's stop endpoint does return `304 container already stopped`; the handling is kept.
- *"DemuxLogs trusts the 4 GB frame size header"* — frames stream through `io.Copy` over a `LimitReader` (32 KiB buffer); no allocation proportional to frame size.

CI is green at `8e6f27f` (kapsis-ctl Build & Test, ShellCheck, CI Success aggregator).

---

PR #401, opened 2026-06-10
fix(operator): enforce AgentRequest security invariants in reconciler

aviadshiber:
## Summary

This PR closes a **Critical** finding from a security scan of the K8s operator: the `ValidatingAdmissionWebhook` that enforces every multi-tenant security invariant was **fully implemented and tested but never wired in** — not registered in `main.go`, and disabled in the kustomization. With it off, the CRD schema was the only enforcement.

**Impact (before this PR):** any user who can create an `AgentRequest` in a watched namespace could:
- run an **arbitrary container image** from any registry (the allowlist was the only guard);
- select a **privileged service account**;
- enable `nestedContainers: true` in **any** namespace, which grants a hostPath mount of `/dev/fuse`, `allowPrivilegeEscalation=true`, and unconfined seccomp/AppArmor — a strong node-escape primitive.

## Fix — defense in depth

Rather than rely solely on a single admission webhook being enabled, the same invariants are now enforced in the reconciler itself:

- **`webhook` package**: new exported `Validate()` and `NewValidatorFromEnv()` constructor reading `KAPSIS_IMAGE_ALLOWLIST`, `KAPSIS_APPROVED_SERVICE_ACCOUNTS`, `KAPSIS_NESTED_CONTAINER_NAMESPACES`. Safe defaults when unset: images restricted to `ghcr.io/aviadshiber/kapsis/`, `serviceAccountName` must be `kapsis-agent`, and `nestedContainers` is **forbidden in every namespace**.
- **Reconciler**: validates each `AgentRequest` in `reconcilePending` *before* building the Job. A rejected request goes terminal-`Failed` with the reason and **no Job is ever created**. This holds regardless of webhook deployment state.
- **`main.go`**: constructs the validator from env, sets it on the reconciler, and registers the admission webhook (gated by `ENABLE_WEBHOOKS`) for admission-time defense in depth.
- **manager manifest**: documents the three new env vars.

The defaults are chosen to be safe-by-default yet non-breaking for normal deployments (which use the project's own agent images and the `kapsis-agent` SA). Operators running other registries set `KAPSIS_IMAGE_ALLOWLIST`.

## Tests

- `webhook`: `NewValidatorFromEnv` defaults + overrides, `splitEnvList`, the exported `Validate` wrapper. All pass.
- `controller`: new envtest spec asserting a validator-rejected request is set to `Failed` with **no Job created** — verified passing (`1 Passed | 0 Failed` under focus).
- `go build ./...`, `go vet ./...` clean; `gofmt` clean on all touched files.

### Note on unrelated pre-existing test failures
Two controller specs (`When a Job has failed` / `…completed`) fail under the local **envtest 1.35** binaries because newer Kubernetes Job-status validation now requires `FailureTarget=true` + `startTime`. These failures **reproduce on the base branch without any of these changes** and are unrelated to this PR.

## Scope / follow-ups not in this PR
This addresses one of three Critical findings from the scan. Still open (larger, architectural): wiring the full webhook **deployment** plumbing (cert-manager + `config/webhook` kustomize, which is incomplete in-tree), and the DNS-only network filter lacking IP-layer egress enforcement. Happy to tackle either next.

https://claude.ai/code/session_01Rikh5cMWFwtsD2AJo7W7ce

---
_Generated by [Claude Code](https://claude.ai/code/session_01Rikh5cMWFwtsD2AJo7W7ce)_

### DIFF
--- a/operator/cmd/main.go
+++ b/operator/cmd/main.go
@@ -211,6 +211,9 @@ func main() {
 			setupLog.Error(err, "Failed to create webhook", "webhook", "AgentRequest")
 			os.Exit(1)
 		}
+		setupLog.Info("Admission webhook registered", "certPath", webhookCertPath)
+	} else {
+		setupLog.Info("Admission webhook NOT registered; reconciler-side validation remains active")
 	}
 	// +kubebuilder:scaffold:builder
 
--- a/operator/cmd/main.go
+++ b/operator/cmd/main.go
@@ -197,11 +197,16 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Register the admission webhook unless explicitly disabled. The reconciler
-	// already enforces the same invariants, so this is admission-time defense in
-	// depth; it requires the ValidatingWebhookConfiguration + serving certs to be
-	// deployed (see config/webhook + cert-manager) to take effect.
-	if os.Getenv("ENABLE_WEBHOOKS") != "false" {
+	// Register the admission webhook only when explicitly enabled. Registering it
+	// makes controller-runtime start the webhook server, which fails the whole
+	// manager when serving certs are absent — and the in-tree manifests
+	// (config/default, config/manager via scripts/k8s-deploy.sh) do not yet ship
+	// the cert-manager + config/webhook plumbing. The reconciler enforces the
+	// same invariants unconditionally, so the webhook is admission-time defense
+	// in depth for deployments that provision certs and the
+	// ValidatingWebhookConfiguration. Once that plumbing lands in-tree, this gate
+	// can move to the kubebuilder `!= "false"` convention.
+	if os.Getenv("ENABLE_WEBHOOKS") == "true" {
 		if err := validator.SetupWebhookWithManager(mgr); err != nil {
 			setupLog.Error(err, "Failed to create webhook", "webhook", "AgentRequest")
 			os.Exit(1)
--- a/operator/cmd/main.go
+++ b/operator/cmd/main.go
@@ -37,6 +37,7 @@ import (
 
 	kapsisv1alpha1 "github.com/aviadshiber/kapsis/operator/api/v1alpha1"
 	"github.com/aviadshiber/kapsis/operator/internal/controller"
+	kapsiswebhook "github.com/aviadshiber/kapsis/operator/internal/webhook"
 	// +kubebuilder:scaffold:imports
 )
 
@@ -178,13 +179,33 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Security invariants enforced by both the reconciler (always) and the
+	// admission webhook (when deployed). Configured from operator environment.
+	validator := kapsiswebhook.NewValidatorFromEnv()
+	setupLog.Info("Security validator configured",
+		"imageAllowlist", validator.ImageAllowlistPatterns,
+		"approvedServiceAccounts", validator.ApprovedServiceAccounts,
+		"nestedContainerNamespaces", validator.NestedContainerNamespaces)
+
 	if err := (&controller.AgentRequestReconciler{
 		Client:             mgr.GetClient(),
 		Scheme:             mgr.GetScheme(),
 		StatusSidecarImage: os.Getenv("KAPSIS_STATUS_SIDECAR_IMAGE"),
+		Validator:          validator,
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "Failed to create controller", "controller", "AgentRequest")
 		os.Exit(1)
+	}
+
+	// Register the admission webhook unless explicitly disabled. The reconciler
+	// already enforces the same invariants, so this is admission-time defense in
+	// depth; it requires the ValidatingWebhookConfiguration + serving certs to be
+	// deployed (see config/webhook + cert-manager) to take effect.
+	if os.Getenv("ENABLE_WEBHOOKS") != "false" {
+		if err := validator.SetupWebhookWithManager(mgr); err != nil {
+			setupLog.Error(err, "Failed to create webhook", "webhook", "AgentRequest")
+			os.Exit(1)
+		}
 	}
 	// +kubebuilder:scaffold:builder
 
--- a/operator/config/manager/manager.yaml
+++ b/operator/config/manager/manager.yaml
@@ -68,6 +68,13 @@ spec:
         #   value: kapsis-agent
         # - name: KAPSIS_NESTED_CONTAINER_NAMESPACES # namespaces allowed to use nestedContainers
         #   value: kapsis-privileged
+        #
+        # Opt-in admission webhook (defense in depth; the reconciler always
+        # enforces the invariants above). Requires serving certs mounted into the
+        # pod and a ValidatingWebhookConfiguration (config/webhook + cert-manager)
+        # — without them the manager fails to start.
+        # - name: ENABLE_WEBHOOKS
+        #   value: "true"
         ports: []
         securityContext:
           readOnlyRootFilesystem: true
--- a/operator/config/manager/manager.yaml
+++ b/operator/config/manager/manager.yaml
@@ -57,6 +57,17 @@ spec:
         # Optional: override the status sidecar image.
         # - name: KAPSIS_STATUS_SIDECAR_IMAGE
         #   value: ghcr.io/aviadshiber/kapsis/status-sidecar:latest
+        #
+        # Security validation (enforced by the reconciler always, and by the
+        # admission webhook when deployed). Defaults: images restricted to
+        # ghcr.io/aviadshiber/kapsis/, serviceAccountName must be kapsis-agent,
+        # and nestedContainers=true is forbidden in every namespace. Override:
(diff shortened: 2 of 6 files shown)


aviadshiber (commented):
## Ensemble review — majority vote (3 reviewers)

Three independent reviewers examined this PR. The findings below were flagged by **at least 2 of 3** reviewers (majority). Single‑reviewer findings were dropped to reduce noise. Severity is the median of the three reviewers' assignments where they overlapped.

> Posted as a comment review (cannot REQUEST_CHANGES on own PR) — treat #1 and #2 as blockers.

---

### 🔴 High

**1. Reconciler-side validation only runs in `reconcilePending`; spec updates bypass enforcement** *(3/3)*
`operator/internal/controller/agentrequest_controller.go` ~L128‑151
`Reconcile` short-circuits to `reconcileWithJob` whenever the Job already exists, and `r.Validator.Validate` is only invoked inside `reconcilePending`. If the admission webhook isn't deployed (the PR body acknowledges the webhook overlay is still off in `config/default/kustomization.yaml`), a user who updates an existing AgentRequest spec (e.g., swaps in a forbidden image, flips `security.nestedContainers=true`, sets `network.mode=open`, adds reserved `KAPSIS_*` env vars) will not be re-validated. The PR's headline claim that invariants hold *"regardless of webhook deployment state"* doesn't hold for the update path. Either move `Validate` to the top of `Reconcile` (before the terminal-phase short-circuit), or refuse spec changes after Job creation, or explicitly document that update-path enforcement requires the webhook.

**2. Default-on webhook registration + missing TLS/kustomization plumbing breaks the default install** *(2/3, plus related findings from all 3)*
`operator/cmd/main.go` ~L204‑209, `operator/config/manager/manager.yaml` ~L54‑71, `operator/config/default/kustomization.yaml` ~L23
`ENABLE_WEBHOOKS` defaults to enabled (only the literal string `"false"` disables it — `0`/`no`/`False` all enable), but the default kustomization still has `#- ../webhook` and `#- ../certmanager` commented out, and `manager.yaml` mounts no serving certs. As a result, `SetupWebhookWithManager` will register a TLS-serving endpoint that controller-runtime's `certwatcher` cannot satisfy, and the manager will fail to start (or stay non-functional) on a fresh `kustomize build config/default | kubectl apply`. Pick one:
- set `ENABLE_WEBHOOKS=false` in `manager.yaml` until the webhook + cert-manager overlay is enabled by default, **or**
- enable `../webhook` and `../certmanager` in the default kustomization and ship working certs, **or**
- gate registration on a non-empty `--webhook-cert-path` rather than an env var.

Also: use `strconv.ParseBool` for the env gate so non-canonical falsy values don't silently keep the webhook on.

### 🟠 Medium

**3. Nil `Validator` silently disables all security validation** *(3/3)*
`operator/internal/controller/agentrequest_controller.go` ~L65‑73
The new gate is `if r.Validator != nil`, with a struct comment noting that nil means "skipped" for tests. The PR's defense-in-depth guarantee depends on this field being set in production. A future refactor, an alternate `main.go`, or a forgotten field in a custom harness would silently re-introduce the very vulnerability this PR closes — with no log line, no warning, and no startup error. Prefer fail-closed: require `Validator` non-nil at `SetupWithManager`, panic in `SetupWithManager` if nil, or have the controller construct a default `NewValidatorFromEnv()` when nil.

**4. Validation rejection makes the CR terminally Failed; spec edits don't recover** *(2/3)*
`operator/internal/controller/agentrequest_controller.go` ~L103‑105 / L136‑151
After rejection the reconciler sets `Phase=PhaseFailed`, which the top-level `Reconcile` then treats as terminal (early return at L103). A user who fixes the spec in place gets no re-validation and no Job — they must delete and recreate. The commit message says "the user must fix and resubmit", but this is non-obvious. Either gate the terminal-phase short-circuit on `status.observedGeneration == metadata.generation` so a spec bump unlocks re-reconciliation, or surface the delete-and-recreate requirement loudly in `Status.Message`.

**5. `NewValidatorFromEnv` does not materialize the `kapsis-agent` SA default into the struct** *(2/3)*
`operator/internal/webhook/agentrequest_validator.go` ~L99‑115
When `KAPSIS_APPROVED_SERVICE_ACCOUNTS` is unset, `ApprovedServiceAccounts` stays nil; the `kapsis-agent` default is then applied lazily inside `validateServiceAccount`. This is inconsistent with how `ImageAllowlistPatterns` defaults are populated at construction time, contradicts the constructor docstring ("Defaults to {kapsis-agent}…"), and the new `TestNewValidatorFromEnv_Defaults` even locks in the discrepancy by asserting the field is empty. The split also means the in-memory config is not self-describing (logs / future `kubectl describe` of operator state will show `approvedServiceAccounts=[]` while validation actually enforces `[kapsis-agent]`). Set all defaults in `NewValidatorFromEnv` and drop the lazy fallback.

### 🟡 Low

**6. `validateConfigMountPaths` accepts relative / non-absolute paths and uses OS-aware `filepath.Clean`** *(2/3)*
`operator/internal/webhook/agentrequest_validator.go` ~L267‑281
`filepath.Clean("etc/shadow")` returns `"etc/shadow"` (no leading `/`), which then fails every `HasPrefix("/etc/")` check and is silently accepted. The CRD likely requires absolute paths, but a validator should not depend on schema validation for safety. Also, `path/filepath` is OS-aware (Windows path separators); `path.Clean` matches container mountPath semantics better. Reject any mountPath that does not start with `/` before the prefix check, and switch to `path.Clean`.

**7. `splitEnvList` separator handling: missing `\r`, no dedup, accidental whitespace splits** *(3/3)*
`operator/internal/webhook/agentrequest_validator.go` ~L117‑129
The `FieldsFunc` predicate covers `,`, ` `, `\t`, `\n` but not `\r`, so values copied from CRLF YAML/Windows env get a trailing `\r` that breaks downstream equality (e.g., `"kapsis-agent\r" != "kapsis-agent"`). The result is also not deduplicated. And splitting on plain spaces means a typo like `KAPSIS_IMAGE_ALLOWLIST="ghcr.io/my org/"` silently expands the allowlist to two unrelated prefixes. Either restrict separators to `,` / `\n` and reject entries containing whitespace, or use `unicode.IsSpace` and deduplicate after splitting.

### 🪶 Nit

**8. `rejectingValidator` should use `errors.New` instead of `fmt.Errorf("%s", …)`** *(2/3)*
`operator/internal/controller/agentrequest_controller_test.go` ~L39‑43
No formatting is happening, and a `%` in `reason` would be interpreted as a directive. Use `errors.New(rv.reason)`.

---

### Findings dropped (single-reviewer, not promoted by majority)
For transparency: each reviewer also raised concerns the others did not — e.g. status leakage in `Status.Error` exposing operator config (1/3), `Status.CompletedAt` not set on rejection (1/3), test coverage gaps around `NetworkPolicy` side-effects and the real validator (1/3 each), prefix-based image allowlist matching without a path separator boundary (1/3), and webhook setup happening before `AddHealthzCheck` (1/3). These may still be worth addressing, but they did not clear the majority bar.

### How to read this review
Findings #1 and #2 are the blockers — the rest are recoverable in follow-up commits on this PR.

---

aviadshiber:
Addressed the ensemble review (pushed in `7c6e464`, rebased on the parallel webhook-gating commit):

**Blockers**
- **#1 update-path bypass** — moved `Validate()` to the top of `Reconcile`, so it runs on every reconcile, not just `reconcilePending`. A spec update that introduces a forbidden image / `nestedContainers` / reserved env is now caught even after a Job exists, and the running Job is deleted (Background propagation) to stop the workload. New envtest spec covers this (Job created → reconcile with rejecting validator → Job gone + `Failed`).
- **#2 broken default install** — handled by the parallel commit `2c35229` (opt-in `ENABLE_WEBHOOKS=="true"`); I rebased onto it rather than clobbering it, keeping only the registered/not-registered log lines and dropping my `strconv`/cert-path variant.

**Medium / Low / Nit**
- **#3** `SetupWithManager` now errors if `Validator` is nil (fail-closed; tests that opt out call `Reconcile` directly).
- **#4** rejected CRs carry a `Status.Message` stating recovery requires delete-and-recreate.
- **#5** `NewValidatorFromEnv` materializes the `kapsis-agent` SA default into the struct; `TestNewValidatorFromEnv_Defaults` updated accordingly.
- **#6** `validateConfigMountPaths` rejects non-absolute paths and uses `path.Clean` (POSIX) instead of `path/filepath`; new `etc/shadow` relative-path test.
- **#7** `splitEnvList` treats CR + all Unicode whitespace as separators and de-duplicates; new CRLF/dedup cases.
- **#8** `rejectingValidator` uses `errors.New`.

`go build`/`vet`/`gofmt` clean; webhook unit tests and the three security controller specs pass under envtest.

I did **not** add `ObservedGeneration`-based re-reconciliation for #4 (the review offered the `Status.Message` alternative) — flagging in case you'd prefer the generation-gating approach instead. The two unrelated `When a Job has failed/completed` specs that fail under envtest 1.35 (Job-status API change requiring `FailureTarget`/`startTime`) are pre-existing on the base branch.

---

PR #412, opened 2026-06-13
feat(security): deny userns+mount escalation in default seccomp profile

aviadshiber:
## Summary

Closes a container-escape gap surfaced during PR #397 research: under the **default `standard`** security profile, `scripts/lib/security.sh` set `[standard_seccomp]=false`, so Kapsis applied **no custom seccomp profile** and relied on Podman's stock default — which **unconditionally permits `unshare(CLONE_NEWUSER)`** and the `mount`/`fsconfig` family. A code-executing agent could therefore create a **nested user namespace** (gaining `CAP_SYS_ADMIN` *relative to it* even after Kapsis drops all host capabilities) and reach the mount-escalation surface — a **CVE-2022-0185 class** primitive — with no capability that Kapsis grants.

This PR makes the default deny that family with **EPERM**, while preserving every other protection from Podman's default profile, and provides a documented opt-out for workloads that legitimately need nested userns/mounts.

> Draft for the owner's review decision (default-on hardening is a behavior change). Submitted as **draft** pending a live podman/macOS run (see (f)).

---

### (a) Exact gap found per-profile (Phase 1)

All three shipped profiles use `defaultAction: SCMP_ACT_ERRNO` (deny-by-default). They are **hand-rolled allowlists, NOT derived from the upstream containers/podman default** — which is why simply flipping `[standard_seccomp]=true` to reuse `kapsis-agent-base.json` would have *regressed* `standard` to a thin allowlist that could miss syscalls real tools need.

| Profile (pre-PR) | Seccomp applied? | `unshare` | `setns` | `mount`/`umount2`/`fsopen`/`fsconfig`/`fspick`/`move_mount`/`open_tree`/`pivot_root`/`mount_setattr` | `clone`/`clone3` | `keyctl`/`add_key` |
|---|---|---|---|---|---|---|
| **`standard` (DEFAULT)** | **none** → Podman stock default | **ALLOW** | **ALLOW** (gated CAP_SYS_ADMIN) | **ALLOW** (unconditional in upstream default) | ALLOW | keyctl ALLOW / add_key absent |
| `strict` | `kapsis-agent-base.json` | **ALLOW** (explicit) | **ALLOW** (explicit) | DENY (absent from allowlist → default ERRNO) | ALLOW | DENY |
| `paranoid` | `kapsis-agent-base.json` | **ALLOW** (explicit) | **ALLOW** (explicit) | DENY | ALLOW | DENY |
| `minimal` | none (by design) | n/a | n/a | n/a | n/a | n/a |
| `kapsis-interactive.json` (debug) | opt-in | **ALLOW** | **ALLOW** | DENY | ALLOW | DENY |

**Load-bearing finding:** `standard` (the default the product ships) left the *entire* family open. `strict`/`paranoid` already denied the mount family but still **explicitly allowed `unshare`+`setns`** — the primitive that re-opens it via a nested userns.

I confirmed against the live upstream `containers/common` default profile that `unshare`, `mount`, `umount2`, `fsopen`, `fsconfig`, `fspick`, `move_mount`, `open_tree`, `pivot_root`, `mount_setattr`, `clone`, `clone3`, `keyctl` are **unconditionally allowed**; only `setns` and `bpf` are CAP_SYS_ADMIN-gated.

### (b) What I now deny, and the EPERM-not-KILL choice

New **`security/seccomp/kapsis-default-hardened.json`** = the **verbatim upstream containers/common default** (keeps all of upstream's coverage + 7 architectures: x86_64, aarch64, mips*, s390x) with this family **denied**, returning **`EPERM` (`SCMP_ACT_ERRNO`, errno 1)**:

`unshare`, `setns`, `mount`, `umount2`, `fsopen`, `fsconfig`, `fspick`, `move_mount`, `open_tree`, `pivot_root`, `mount_setattr`

Done two ways for belt-and-suspenders: (1) the family is **removed from every allow rule** (so it falls through to the deny default), and (2) an **explicit EPERM block** is prepended. The explicit block uses `errnoRet: 1` (EPERM), which differs from the profile's `defaultErrnoRet: 38` (ENOSYS) — so libseccomp does **not** reject it as "matches default action of filter" (the same pattern upstream uses for `setns`/`bpf`).

**EPERM, not `SCMP_ACT_KILL`**, on purpose: tools that *probe* for userns/mount support (sandboxing libs, language test harnesses) see a clean "operation not permitted" and degrade, instead of being SIGSYS-killed.

The two legacy thin-allowlist profiles (`kapsis-agent-base.json`, `kapsis-interactive.json`) were **also** updated to deny `unshare`/`setns` with EPERM, so every shipped enforcing profile now closes the hole.

### (c) Scope decision + opt-out

- **Scope: default hardened.** `[standard_seccomp]=true`; `get_seccomp_profile()` now returns `kapsis-default-hardened.json` for **standard / strict / paranoid** (preferred over the legacy thin base so we don't regress coverage). `minimal` keeps seccomp off.
- **Opt-out: `KAPSIS_ALLOW_USERNS=true`** (env) or `security.seccomp.allow_userns: true` (YAML; env wins). Swaps to new **`kapsis-default-userns.json`** = the verbatim upstream default (userns allowed) — restores the prior behavior, but as an explicit, auditable profile.
- **Master kill switch** `KAPSIS_SECCOMP_ENABLED=false` still disables seccomp entirely.
- Precedence in `get_seccomp_profile`: audit mode → explicit `KAPSIS_SECCOMP_PROFILE` → `KAPSIS_ALLOW_USERNS` swap → agent-specific → hardened default → legacy base fallback.

### (d) Known breakage (use the opt-out)

- **Nested containerization** — running Podman/Docker *inside* the agent container.
- **`bubblewrap` / `nsjail`** sandboxes.
- **Chromium / Playwright / Electron** — they set up a userns+mount sandbox unless run with `--no-sandbox`.

All three break under the default and must set `KAPSIS_ALLOW_USERNS=true`. Documented in `docs/SECURITY-HARDENING.md` §1.3, `security/README.md` (troubleshooting), and `docs/CONFIG-REFERENCE.md`.

### (e) Residual risk

- **`clone(CLONE_NEWUSER)` is left allowed.** `clone`/`clone3` are the basis of all process/thread creation; denying them breaks everything, and portable arg-filtering of the `CLONE_NEWUSER` bit across `clone`/`clone3` register layouts in static OCI JSON is impractical. Mitigation: the **load-bearing** escalation path is `unshare(CLONE_NEWUSER)` → `mount`/`fsconfig`, and **both ends are denied** — even if an agent reaches a nested userns via `clone`, `mount`/`fsconfig`/`setns` are still EPERM. `clone(CLONE_NEWUSER)` is left to the complementary defenses already in place: rootless userns remap (`--userns=keep-id`), `--cap-drop=ALL`, `--security-opt no-new-privileges`.
- `keyctl`/`bpf` follow upstream (keyctl allowed, bpf CAP-gated) — not the mount-escalation surface this PR targets; can be tightened in a follow-up if desired.

### (f) Validation that still needs a live run

This sandbox has **no `podman`**, so kernel-level enforcement was **not** runtime-verified. Validated here: every modified/new seccomp JSON parses (`json.load`); `shellcheck -S warning` clean on all modified shell; unit tests assert profile selection + that the JSON denies the family (EPERM) while keeping clone/exec; `./tests/run-all-tests.sh --quick` adds **no new failures** vs. an origin/main baseline.

**Owner should run on a real podman/macOS host:**
```bash
podman run --rm --security-opt seccomp=security/seccomp/kapsis-default-hardened.json alpine \
  sh -c 'unshare -Urm true; echo "unshare rc=$? (expect non-zero / Operation not permitted)"; echo ok | cat; echo "normal rc=$?"'
# Expect: unshare denied (EPERM), normal command succeeds.
# And confirm a real agent run (claude/codex) still completes end-to-end.
```

---

### Dashboard Sync

This adds a config key (`security.seccomp.allow_userns` / `KAPSIS_ALLOW_USERNS`) but **no new status field, audit event, exit code, CLI flag, or container resource** — `status.json` (the dashboard's data source) is unchanged, and the dashboard does not currently surface seccomp/security-profile state. No dashboard change required.

### Files
- `security/seccomp/kapsis-default-hardened.json` (new), `kapsis-default-userns.json` (new)
- `security/seccomp/kapsis-agent-base.json`, `kapsis-interactive.json` (deny unshare/setns)
- `scripts/lib/security.sh` (default on + selection logic + summary)
- `scripts/launch-agent.sh` (YAML `allow_userns`), `scripts/lib/config-verifier.sh` (validate it)
- `docs/SECURITY-HARDENING.md`, `docs/CONFIG-REFERENCE.md`, `security/README.md`
- `tests/test-security.sh` (updated + 5 new tests)

https://claude.ai/code/session_01GyUwb9P9abPA2WnUAqnCUU

---
_Generated by [Claude Code](https://claude.ai/code/session_01GyUwb9P9abPA2WnUAqnCUU)_


aviadshiber:
Pushed `31bb203` to fix the two CI failures on `e7160ee`:

**Config Validation + `test-config-verifier.sh` (Unit Tests)** — same root cause. The `security.seccomp.allow_userns` validator extracted the value via `yq 'if (.security.seccomp | type == "object") and (.security.seccomp | has("allow_userns")) ...'`. Under **mikefarah yq v4** (what CI installs) `and` evaluates both operands, so `has()` runs against the absent `.security.seccomp` (null) and errors; with `$(yq … 2>/dev/null)` under `set -e` the failing substitution aborted the whole verifier with no `[FAIL]` line (it died mid-`aider.yaml`, which has no `security` section). The dev sandbox's non-mikefarah yq masked it — which is why the PR body's note about using `has()` to avoid the `false // "null"` falsy-coalesce was validated locally but broke on CI.

The fix keeps the *intent* (distinguish absent from explicit `false`) without `has()`/`type`: plain-path `yq -r '.security.seccomp.allow_userns'` yields `"null"` only when truly absent and `"true"`/`"false"` when set, plus `|| echo null` to stay `set -e`-safe. Mirrors the existing `.security.userns` idiom in the same file. (So the PR body's "avoided via has()" note is now superseded by this commit.)

**`test-liveness-monitor.sh` (Unit Tests)** — unrelated flake: this PR doesn't touch it (byte-identical to `main`), and it passes locally 57/0. The re-run triggered by this push should clear it; if it recurs it's a pre-existing timing flake to track separately.

---

aviadshiber:

Reviewed `fdc70829de1dc5c6a7931bc8daeaa1b342063e44` (branch `claude/harden-seccomp-userns`).

### Critical (merge-blocking)

1. **`security/seccomp/kapsis-default-hardened.json` unconditionally denies `mount(2)` (no `args` filter) and this is now the default for `standard`/`strict`/`paranoid`, breaking Kapsis's own in-container fuse-overlayfs CoW mechanism.**
   `scripts/lib/security.sh` flips `[standard_seccomp]` from `false` → `true`, and `get_seccomp_profile()` returns `kapsis-default-hardened.json` by default. That profile's mount-family rule is `{"names": [..., "mount", ...], "action": "SCMP_ACT_ERRNO", "args": [], "errnoRet": 1}` — no argument filtering, so it can't distinguish a FUSE overlay mount from a privilege-escalation mount. `scripts/entrypoint.sh` calls `fuse-overlayfs -o "lowerdir=...,upperdir=...,workdir=..." "$dst"` inside the container in both `setup_staged_config_overlays()` (staged `.claude`/`.ssh`/`.config` CoW) and `setup_fuse_overlay()` (macOS legacy `/workspace` CoW) — both invoke `mount(2)` from inside the seccomp-confined process. This is a functional regression on essentially every default-profile launch, and it is not listed in the PR's "Known breakage" section (which only calls out nested containerization, bubblewrap/nsjail, and Chromium/Playwright/Electron). No test exercises `setup_staged_config_overlays()`/`setup_fuse_overlay()` under the new hardened default.

2. **The new explicit `unshare`/`setns` deny rule added to the two legacy profiles (`kapsis-agent-base.json`, `kapsis-interactive.json`) uses the same action+errno as those profiles' own filter-wide default, risking a seccomp profile load rejection.**
   Verified via diff: both files have (unchanged) `"defaultAction": "SCMP_ACT_ERRNO", "defaultErrnoRet": 1`. The new rule added by this PR is `{"names": ["setns","unshare"], "action": "SCMP_ACT_ERRNO", "errnoRet": 1, "errno": "EPERM"}` — identical action+errno to the default. The PR's own description explicitly relies on differentiating errno values to avoid this exact problem for the *new* hardened profile ("The explicit block uses errnoRet: 1 (EPERM), which differs from the profile's defaultErrnoRet: 38 (ENOSYS) — so libseccomp does not reject it as 'matches default action of filter'"), but that reasoning is not applied to `kapsis-agent-base.json`/`kapsis-interactive.json`, where both are errnoRet 1. If runc/libseccomp treats this as a no-op or rejects it, `strict`, `paranoid`, and interactive/debug-mode container launches (which use these legacy profiles) could fail to start, or silently fail to actually deny `unshare`/`setns` (defeating the fix for exactly the profiles that previously allowed it explicitly). No podman was available in the PR author's sandbox to runtime-validate seccomp loading (PR body section (f)).

3. **The new/updated security tests never run in CI — `tests/test-security.sh` is registered in neither `QUICK_TESTS` nor the `security` category in `tests/run-all-tests.sh`.**
   Verified on the PR branch: `grep -n "test-security\.sh" tests/run-all-tests.sh` (PR ref) returns nothing. `QUICK_TESTS` (line 78) and the `security)` category list (line 47-48) both omit it; no `.github/workflows/*.yml` invokes it directly. This means the one automated check for the two default-on behavior changes above (`test_hardened_profile_denies_userns_mount`, `test_seccomp_allow_userns_optout`, etc.) is dead code from CI's perspective, and the PR description's claim "`./tests/run-all-tests.sh --quick` adds no new failures" provides no actual signal about the new seccomp logic. Combined with findings 1-2 (real functional/robustness risk, unvalidated on a live podman host) and the complete absence of any container-level runtime test, this is a merge-blocking gap for a default-on, high-blast-radius security change. (Pre-existing: `test-security.sh` was already unregistered on `main`; this PR compounds the problem by adding new load-bearing assertions to a file nothing runs.)

### Non-critical

- **No test covers the `security.seccomp.allow_userns` / `KAPSIS_ALLOW_USERNS` YAML opt-out wiring end-to-end.** Verified: `grep -rn "allow_userns" tests/` (excluding `test-security.sh`'s env-var-only test) finds nothing in `tests/test-config-verifier.sh`, `tests/test-config-resolution.sh`, or `tests/test-agent-config-override.sh`, despite new code in `scripts/lib/config-verifier.sh` (~line 387) and `scripts/launch-agent.sh` (~line 1103) reading this key. This is the only user-facing way (besides the env var) to disable the hardened default for workloads called out as broken by it. (`scripts/lib/config-verifier.sh`, `scripts/launch-agent.sh`)
- **No test asserts `kapsis-agent-base.json`/`kapsis-interactive.json` actually deny `unshare`/`setns`**, even though the PR changes both from `SCMP_ACT_ALLOW` to `SCMP_ACT_ERRNO`. `grep -n "kapsis-agent-base\|kapsis-interactive" tests/test-security.sh` (PR ref) returns no matches — only the new hardened/userns profiles are asserted. A future revert of either legacy file's deny block would pass the full suite silently. (`tests/test-security.sh`)
- **No test asserts `kapsis-default-hardened.json` and `kapsis-default-userns.json` stay in sync** apart from the 11-syscall deny block. Verified via `diff`: the two ~1140-line files are identical except for that block, but nothing catches future drift if one is edited without the other (e.g. an upstream `containers/common` sync). (`security/seccomp/kapsis-default-hardened.json`, `security/seccomp/kapsis-default-userns.json`)
- **No test confirms the `minimal` profile still has seccomp disabled** after `[standard_seccomp]` flips to `true` by default. `tests/test-security.sh` exercises `minimal` for capability-drop and process-isolation but never calls `get_seccomp_profile`/`generate_seccomp_args` with `KAPSIS_SECURITY_PROFILE=minimal`. Pre-existing gap, widened by this PR's new hardened-default branch in `get_seccomp_profile()`. (`tests/test-security.sh`)
- **No test launches an actual container with the new hardened seccomp profile.** All assertions are static JSON/string checks; nothing exercises `podman run --security-opt seccomp=security/seccomp/kapsis-default-hardened.json` to confirm normal agent workloads still boot and run without an unintended EPERM. The PR description itself flags this as unvalidated and recommends the owner run it manually before merge. (`tests/test-security.sh`)

None — all raised scope/coverage claims were verified against the actual PR branch content (`git show origin/claude/harden-seccomp-userns:<path>`) and held up.

aviadshiber:

Reviewed SHA: `fdc70829de1dc5c6a7931bc8daeaa1b342063e44`

### Critical (merge-blocking)

1. **`tests/test-security.sh` is modified with load-bearing new tests, but the file is still never executed by CI** (`tests/run-all-tests.sh`)
   - `QUICK_TESTS` (line 78) does not list `test-security.sh`.
   - `get_tests_for_category("security")` (line 48) lists `test-security-no-root.sh` — a *different* file — not `test-security.sh`.
   - Verified directly on `origin/claude/harden-seccomp-userns`: `grep -c "test-security.sh" tests/run-all-tests.sh` → `0`.
   - This gap pre-dates the PR (file was added 2026-01-04, never registered), but the PR adds the one test that actually matters for this change — `test_hardened_profile_denies_userns_mount`, which parses `kapsis-default-hardened.json` and fails if `unshare`/`setns`/`mount`/etc. leak into an `SCMP_ACT_ALLOW` block or lack an `EPERM` entry. That test currently passes (`bash tests/test-security.sh` → all 43 pass), but a future regression to the hardened profile (the exact CVE-2022-0185-class escalation this PR closes) would ship silently because neither `run-all-tests.sh --quick` nor plain `run-all-tests.sh` (both used in `.github/workflows/ci.yml`) ever invoke it.
   - Fix: add `test-security.sh` to `QUICK_TESTS` and/or the `security` category in `tests/run-all-tests.sh`.

### Non-critical

- **`scripts/lib/security.sh:566` (`print_security_summary`)** — cosmetic: when `KAPSIS_SECCOMP_PROFILE` points at an arbitrary custom JSON, the summary's `else` branch prints `"enforcing, userns+mount DENIED (<basename>)"` unconditionally, without inspecting the custom file's actual content. Verified the `else` branch is reached for any profile that isn't `kapsis-default-userns.json` or `kapsis-audit.json`, including a custom profile that might allow `unshare`/`mount`. Kernel enforcement is correct (whatever the custom file says); only the human-readable status line can be misleading. (minor)

- **`scripts/lib/security.sh` `get_seccomp_profile()`** — if both `kapsis-default-hardened.json` and `kapsis-agent-base.json` are missing/unreadable, the function returns empty and `generate_seccomp_args()` silently emits no `--security-opt`, so the container launches with no seccomp enforcement and no error. Checked packaging: Homebrew's `libexec.install Dir["*"]` installs the whole source tree including `security/`, so this isn't currently reachable via the shipped Homebrew path, but no explicit early validation exists in `preflight-check.sh` (confirmed zero `seccomp` references there) to catch a corrupted/stripped install on any channel. Worth a defensive check but not merge-blocking given the files are shipped and installed by all three packaging paths. (major → downgraded, not merge-blocking)

- **`scripts/launch-agent.sh:1104` / `scripts/lib/config-verifier.sh:399`** — the `KAPSIS_ALLOW_USERNS` YAML-key parsing (`yq -r '.security.seccomp.allow_userns // ""'` vs. the plain-path pattern in `config-verifier.sh`) is duplicated with slightly different idioms across the two files. Correct as written, but worth consolidating into a shared helper for maintainability. (minor)

- None — all raised findings were substantiated at some severity; only the `preflight-check.sh` "silent fallback" item was downgraded from major to non-blocking after confirming all three packaging paths (Homebrew `Dir["*"]`) install the `security/` directory containing the profile JSONs.

Independently re-verified the core security logic: pulled `security/seccomp/kapsis-default-hardened.json` from the PR branch and re-ran the same deny-list check the new test performs — `defaultAction` is `SCMP_ACT_ERRNO`, none of `unshare/setns/mount/umount2/fsopen/fsconfig/fspick/move_mount/open_tree/pivot_root/mount_setattr` appear in any `SCMP_ACT_ALLOW` block, all are present with `errnoRet=1` (EPERM), and `clone/clone3/execve/openat` remain allowed. The enforcement itself is sound — the only blocking issue is that the regression test proving this isn't wired into CI.

aviadshiber:
✅ **Round-1 critical resolved** (commit `014d57f`): `test-security.sh` (incl. `test_hardened_profile_denies_userns_mount`) is now registered in `run-all-tests.sh` so CI actually executes it (43/43 passing locally).

PR #413, opened 2026-06-14
fix(security): provenance-protect gist sentinel strip against rogue-agent spoofing (#408)

### LINKED ISSUE #408: Provenance-protect the commit-strip sentinels against rogue-agent spoofing (#394 follow-up)
## Context
PR #394's host-side commit strip removes content between Kapsis sentinel markers (`<!-- KAPSIS_GIST_BEGIN -->` … `<!-- KAPSIS_GIST_END -->`). The ensemble review noted the host has **no provenance check on who wrote the markers**: a rogue in-container agent can wrap arbitrary content in those sentinels and the host will strip it.

## Threat assessment (bounded, but real)
Because the strip runs at commit time, a rogue agent can only use the markers to **delete its own to-be-committed content** (hide changes from the user's reviewed diff) — it cannot smuggle content *into* the repo this way (stripped content is discarded, not committed). So severity is bounded, which is why it was deferred rather than blocking #394. Still worth closing for the "maximal containment of rogue agents" posture.

## Proposed mitigation (design needed)
Record provenance at injection time so the strip only removes blocks Kapsis itself wrote:
- At injection (`inject_*`/gist-injection path), record either (a) exact byte ranges of the injected block, or (b) a per-session HMAC over the block keyed by a host-only secret, stored host-side (not in the container-visible file).
- At strip time, only remove blocks whose byte-range/HMAC matches what Kapsis recorded; leave agent-authored sentinel look-alikes intact (and ideally flag them).
- Tie the secret/record to the per-agent session so it cannot be replayed across runs.

## Pointers
- Sentinel constants now centralized in `scripts/lib/constants.sh` (#394).
- Injection + strip: gist-injection path and `scripts/post-container-git.sh:strip_kapsis_injections()`.
- Related hardening: host-only forgery-resistant sentinel pattern already used by the vfkit watchdog (`scripts/lib/vfkit-watchdog.sh`, `$TMPDIR` host-private sentinel) — same trust-anchor idea.

_Tracked from the post-merge cleanup of the #394 ensemble review. https://claude.ai/code/session_01GyUwb9P9abPA2WnUAqnCUU_

aviadshiber:
## Problem (closes #408)

PR #394 adds a host-side strip in `post-container-git.sh` that removes Kapsis-injected gist instruction blocks between `<!-- KAPSIS_GIST_BEGIN -->` / `<!-- KAPSIS_GIST_END -->` sentinels from `CLAUDE.md` and `AGENTS.md` before committing.

**Without this fix**, there is no provenance check: a rogue in-container agent can wrap arbitrary content in those markers and the host will silently strip it from the commit diff — hiding the agent's changes from the user.

## Solution

Three-layer defence:

### 1. Host-side pre-injection (`scripts/launch-agent.sh`)
A new `host_inject_gist_instructions()` function runs **on the host, before the container starts** (worktree mode only, when `KAPSIS_INJECT_GIST=true`). It:
- Renders `gist-instructions.md` using the same template substitution as the container's `inject_gist_instructions()`
- Wraps the content in sentinel markers
- Computes SHA-256 of the sentinel-wrapped block
- Writes the hash to `$TMPDIR/kapsis-<agent-id>-gist-proof` — a **host-private path, never bind-mounted into the container**

### 2. Sentinel markers in injection (`scripts/lib/inject-status-hooks.sh`)
`inject_gist_instructions()` now wraps appended content in the sentinel markers (used as fallback when host-side injection is unavailable).

### 3. Provenance-verified strip (`scripts/post-container-git.sh`)
`strip_kapsis_injections()` is called from `commit_changes()` **before staging**. For each sentinel block found in `CLAUDE.md`/`AGENTS.md`:
- Computes its SHA-256
- Compares against the proof file
- **Strips only matching blocks** (verified as Kapsis-injected)
- **Preserves and warns** on mismatching blocks — rogue agent content is NOT stripped and appears in the commit diff for review

The proof file is deleted after first use (one-time token) to prevent replay across sessions.

## Why this defeats the attack

The container **cannot forge** the host-side proof file — `$TMPDIR` is host-private. A rogue agent can:
- Observe the sentinel markers in `CLAUDE.md` (they're visible in the workspace)
- Write its own sentinel blocks around different content

But those blocks will have a **different SHA-256** from what the host recorded. They survive the strip and appear in the PR diff, flagged with `log_warn`.

## Files changed

| File | Change |
|------|--------|
| `scripts/lib/constants.sh` | Add `KAPSIS_GIST_SENTINEL_BEGIN/END` constants |
| `scripts/lib/inject-status-hooks.sh` | Wrap injected block in sentinel markers; update idempotency check |
| `scripts/launch-agent.sh` | Add `host_inject_gist_instructions()` + call after `setup_sandbox` |
| `scripts/post-container-git.sh` | Add `strip_kapsis_injections()` + call in `commit_changes()` before staging |
| `tests/test-gist-injection-provenance.sh` | 6 unit tests covering all paths |

## Test plan

- [x] `bash tests/test-gist-injection-provenance.sh` — 6/6 pass
  - Verified block is stripped ✓
  - Rogue block (SHA mismatch) is preserved and flagged ✓
  - No-op when file has no sentinels ✓
  - Graceful degrade when proof file missing ✓
  - Proof file deleted after first use ✓
  - `inject_gist_instructions()` wraps in sentinels ✓
- [x] `bash tests/test-commit-failure-handling.sh` — 13/13 pass (no regression)
- [x] `bash tests/test-git-excludes.sh` — 7/7 pass
- [x] `bash tests/test-input-validation.sh` — 13/13 pass
- [x] `bash -n` syntax check on all 4 modified scripts passes

https://claude.ai/code/session_01JVNGGZYkge9f5CXrcddU4q

---
_Generated by [Claude Code](https://claude.ai/code/session_01JVNGGZYkge9f5CXrcddU4q)_


aviadshiber:

Reviewed SHA: `81708b7f1fea602e8bd84803c7da62ad9a634417`

### Critical (merge-blocking)

1. **`scripts/post-container-git.sh:320`** — GNU-only `sed -e :a -e '/^\n*$/{$d;N;ba}'` in `strip_kapsis_injections()` fails under BSD sed (macOS, the primary dev/target platform per CLAUDE.md). Reproduced locally:
   ```
   sed: 2: "/^\n*$/{$d;N;ba}": unexpected EOF (pending }'s)
   ```
   Because this runs inside a command substitution, the failure is swallowed and `new_content` becomes empty, so `printf '%s\n' "$new_content" > "$file"` (line 321) blanks CLAUDE.md/AGENTS.md — destroying real file content on every macOS run that strips a verified gist block. The PR's own new test (`tests/test-gist-injection-provenance.sh`) fails when actually executed on macOS for this reason.
   **Disposition:** already fixed in stacked PR #427 (replaces the sed pipeline with a portable `awk` equivalent — confirmed present in #427's diff). Absorb that fix into #413 before merge, then rebase #427 on top.

2. **`scripts/post-container-git.sh:267,318`** — `stripped_count` is declared once outside the `for file_rel in "CLAUDE.md" "AGENTS.md"` loop and never reset per file. The rewrite gate `if [[ "$stripped_count" -gt 0 ]]` therefore uses the cumulative count across both files: if CLAUDE.md has a verified block stripped, AGENTS.md gets rewritten via the same reconstruction/sed pipeline even when AGENTS.md itself had nothing verified to strip — combined with finding #1, this risks blanking a file that should have been left byte-for-byte untouched.
   **Disposition:** already fixed in stacked PR #427 (per-file flag instead of the global counter — confirmed present in #427's diff). Absorb into #413, then rebase #427.

### Non-critical

- **`tests/run-all-tests.sh`** — `tests/test-gist-injection-provenance.sh` is added but not registered in `QUICK_TESTS` or any category array, so it never runs in CI (and would not have caught the two criticals above even on Linux runners, where GNU sed masks bug #1). Confirmed absent via grep. PR #427 registers it in both `QUICK_TESTS` and a category — acceptable as a same-stack follow-up rather than blocking #413, given the disposition guidance above.
- **`scripts/lib/constants.sh:51-52`** — `KAPSIS_GIST_SENTINEL_BEGIN` / `KAPSIS_GIST_SENTINEL_END` are defined but never referenced anywhere in the codebase (confirmed via grep — zero matches outside the declaration). The sentinel strings are hardcoded elsewhere (`inject-status-hooks.sh`, `post-container-git.sh`) instead. Minor maintenance smell; not blocking.
- **`scripts/post-container-git.sh:~301`** — When a legitimate container-side fallback injection has no matching host proof file (e.g. overlay mode, missing template, hash failure), it falls into the "unverified/possible rogue injection" branch and logs a warning even though the content is genuinely Kapsis-injected. Fail-safe (content is preserved, never hidden), but can produce alert-fatigue false positives. Worth a follow-up but not merge-blocking.
- **Test coverage gap** — no test exercises the two-file (CLAUDE.md + AGENTS.md both containing sentinel blocks) scenario that would have caught the global-counter bug (#2 above), nor a malformed/truncated proof-file case. Recommend adding both when picking up #427.

None — all raised scope/severity claims were verified as accurate against the diff and the current codebase.

aviadshiber:

Reviewed SHA: `81708b7f1fea602e8bd84803c7da62ad9a634417`

### Critical (merge-blocking)

1. **BSD-sed incompatibility blanks CLAUDE.md/AGENTS.md on macOS strip path** — `scripts/post-container-git.sh` (~line 320)
   `new_content=$(printf '%s' "$new_content" | sed 's/[[:blank:]]*$//' | sed -e :a -e '/^\n*$/{$d;N;ba}')` uses a GNU-only sed label/branch construct. Empirically reproduced on this macOS host: BSD sed errors `sed: 2: "/^\n*$/{$d;N;ba}\n": unexpected EOF (pending }'s)` and the pipeline emits empty output.
   Verified the actual runtime consequence (not just the sed failure): `strip_kapsis_injections()` is invoked from `commit_changes()`, which itself is called as `commit_changes ... || _commit_rc=$?` (line ~1047). Per bash's documented `set -e` exemption for functions invoked as part of an `||`/`&&` list, **errexit is suppressed for the entire call tree**, confirmed with a minimal repro mirroring the real call structure (`outer() { inner; }; outer || _rc=$?` — script does NOT abort, continues with `RC=0`). So the earlier "crashes the whole script" framing is not what happens; instead `new_content` silently becomes empty and `printf '%s\n' "$new_content" > "$file"` **overwrites CLAUDE.md/AGENTS.md with a blank file whenever a verified gist block is legitimately stripped** — real data loss on the happy path, on Kapsis's primary target platform (macOS).
   **Disposition: absorb the fix from stacked PR #427 (replaces the sed pipeline with a portable awk implementation — confirmed present in #427's diff) into #413, then rebase #427.**

2. **Per-file rewrite gated on cumulative/global `stripped_count`** — `scripts/post-container-git.sh` (~line 267, 318)
   `local stripped_count=0` / `local suspicious_count=0` are declared once before the `for file_rel in "CLAUDE.md" "AGENTS.md"` loop and never reset per iteration. The write-back gate `if [[ "$stripped_count" -gt 0 ]]; then ... printf '%s\n' "$new_content" > "$file"; fi` reads the running total across both files, so if CLAUDE.md has a verified strip (stripped_count→1) and AGENTS.md separately has only unverified/suspicious blocks, AGENTS.md is still rewritten via the reconstructed `new_content` even though nothing should have been stripped from it — combined with bug #1 this can blank a file that was never supposed to change.
   **Disposition: absorb the fix from stacked PR #427 (gates rewrite on a per-file flag instead of the global counter — confirmed present in #427's diff) into #413, then rebase #427.**

### Non-critical

- `tests/test-gist-injection-provenance.sh` is added (223 lines, 6 cases) but is **not registered** in `tests/run-all-tests.sh` (absent from `QUICK_TESTS` and every category array) — confirmed via diff inspection. It therefore never runs in CI and would not have caught bug #1/#2 above. Note: stacked PR #427 registers this file — acceptable as a same-branch follow-up, but should not ship long-term unregistered.
- `scripts/lib/constants.sh:52-53` defines `KAPSIS_GIST_SENTINEL_BEGIN` / `KAPSIS_GIST_SENTINEL_END` but confirmed via repo-wide search that **no other file references them** — the sentinel strings are hardcoded elsewhere (`inject-status-hooks.sh`, `post-container-git.sh`). Minor maintenance risk if the sentinel format ever changes; consider using the constants or removing them.
- No test exercises the multi-file (CLAUDE.md **and** AGENTS.md both containing sentinel blocks) or multi-block-per-file scenario — exactly the case that exposes bug #2. Worth adding once #427's per-file fix lands.
- Container-side fallback injection (no host proof file present, e.g. overlay mode or missing template) is treated as a "possible rogue injection" and logged as `log_warn`, even though it may be legitimate Kapsis-authored content. Fail-safe (content is preserved, never silently dropped) but could cause alert fatigue eroding trust in genuine rogue-injection warnings. Flagged as plausible/minor, not blocking.

- The initial framing from one reviewer ("GNU-only sed idiom crashes the entire post-container-git.sh script … the failing command substitution aborts the whole script") was reproduced and found **inaccurate on the specific mechanism**: because the failing code path is inside `strip_kapsis_injections()`, called from `commit_changes()`, which is itself invoked as `commit_changes ... || _commit_rc=$?`, bash's `set -e` exemption for functions in an `||` list means the script does **not** abort — it silently continues and blanks the file instead (folded into Critical #1 above with the corrected mechanism).

aviadshiber:
✅ **Round-1 criticals resolved** (commit `3d1ec05`): absorbed the two strip-correctness fixes from stacked #427 — (1) BSD-sed pipeline replaced with portable awk (was silently blanking rewritten CLAUDE.md/AGENTS.md on macOS), (2) rewrite now gated on a per-file counter instead of the global `stripped_count`. Added BSD-sed regression coverage (7/7 provenance tests passing). #427 has been rebased on top (`--force-with-lease`).

PR #423, opened 2026-07-03
fix(cleanup): guard clean_images against in-use/service images and GC dangling layers proactively

### LINKED ISSUE #418: kapsis-cleanup: `clean_images` removes in-use service images (no allow-list / in-use guard)
## Summary

`scripts/kapsis-cleanup.sh::clean_images()` removes every Podman image whose repository matches the regex `(^|/)kapsis-`, with **no protection** for images currently referenced by a running or stopped container, and **no allow-list** for known service images (e.g. `kapsis-slack-bot`, `kapsis-claude-cli`, `kapsis-sandbox`). This means downstream tools that build service images on top of the Kapsis base layers and run them as long-lived hosts (e.g. a Slack bot, a CI runner) get their own images clobbered as soon as `kapsis-cleanup --images` (or `--vm-health` under inode pressure) fires.

Affected lines: [`scripts/kapsis-cleanup.sh:708-757`](https://github.com/aviadshiber/kapsis/blob/master/scripts/kapsis-cleanup.sh#L708-L757), invoked from `_vm_remediate` at [`:1342-1357`](https://github.com/aviadshiber/kapsis/blob/master/scripts/kapsis-cleanup.sh#L1342-L1357) and from the `--all` / `--images` paths.

## Concrete failure case

In our Slack-bot integration:

1. `rebuild-bot-image.sh` builds the 3-layer chain (`kapsis-sandbox` → `kapsis-claude-cli` → `kapsis-slack-bot`). Podman creates dozens of intermediate `<none>` layers during the build, so total image count crosses the bot's housekeeping threshold (50).
2. Bot restarts; its housekeeping immediately invokes `kapsis-cleanup --images --force` (or `--vm-health --force` when inode usage spikes).
3. `clean_images` lists everything matching `(^|/)kapsis-` and `podman rmi`'s all of them, including `kapsis-slack-bot:latest` that the bot itself was about to launch agents from.
4. Bot's next agent launch fails with pre-flight error: `Kapsis image not found: kapsis-slack-bot:latest`.

Reproducer (any host with the 3-layer service stack built):

\`\`\`bash
podman images | grep kapsis-          # confirm service images present
kapsis-cleanup --images --force        # nukes them all
podman images | grep kapsis-           # all gone
\`\`\`

### Layer 1: in-use guard

Before \`podman rmi\`, check whether any container (running OR stopped) references the image. Cheap, naturally protects active runs.

\`\`\`bash
is_image_in_use() {
    local image_id=\"\$1\"
    podman ps -a --no-trunc --format '{{.ImageID}}' 2>/dev/null \\
        | grep -q \"^\${image_id}\" && return 0
    return 1
}
\`\`\`

### Layer 2: service-image allow-list

Env-driven with sensible defaults baked in. Bot consumers don't have to opt in.

\`\`\`bash
KAPSIS_KEEP_IMAGES=\"\${KAPSIS_KEEP_IMAGES:-kapsis-slack-bot,kapsis-claude-cli,kapsis-sandbox}\"
\`\`\`

Also expose \`--keep-image NAME\` (repeatable) CLI flag for ad-hoc protection.

### Why both

| Scenario | Layer 1 | Layer 2 |
|---|---|---|
| Active agent run, image referenced by container | ✅ | ✅ |
| Idle service (e.g. Slack bot between runs) | ❌ | ✅ |
| \`--vm-health\` auto-remediation under CRITICAL inodes | depends on running state | ✅ |
| Custom user service image | needs \`--keep-image\` | ✅ |

## Scope of change

- \`scripts/kapsis-cleanup.sh\`: modify \`clean_images()\` (~30 lines) — add \`is_image_in_use()\` helper, parse \`--keep-image\` and \`KAPSIS_KEEP_IMAGES\` env, skip matched images.
- \`tests/test-image-cleanup-allowlist.sh\` (new) — 4 cases: in-use container, env allow-list, \`--keep-image\` flag, image without protection still removed.
- Extend \`tests/test-cleanup-vm-health.sh\` — verify CRITICAL-inode remediation respects the allow-list.
- \`README.md\` / docs — document \`KAPSIS_KEEP_IMAGES\` and \`--keep-image\`.
- \`CHANGELOG.md\` — entry under next version.

## Bot-side workaround (already applied)

In our slack-bot housekeeping (\`~/.claude/skills/slack-cli/scripts/handlers/housekeeping.py\`):

- Stopped calling \`kapsis-cleanup --images\` — switched to direct \`podman rmi\` with a bot-side keep-list (\`kapsis-slack-bot\`, \`kapsis-claude-cli\`, \`kapsis-sandbox\`), no \`--force\`.
- Dropped \`--force\` from the \`kapsis-cleanup --vm-health\` invocation — diagnostic only.

This unblocks our deployment but is a workaround. Real fix belongs in \`kapsis-cleanup\`.

## Related changelog note

v2.35.2 / #415 (\"Match \`localhost/\` prefixed images in clean_images\") expanded the matching pattern. The underlying lack of an allow-list pre-dates that change; #415 just made it bite harder in rootless Podman setups where every locally-built image is \`localhost/\`-prefixed.

### LINKED ISSUE #421: kapsis-cleanup: orphaned <none> build layers never GC'd → Podman VM disk fills to 100% → total Podman outage
## Summary

The image-cleanup logic (kapsis `clean_images` **and** the downstream Slack-bot housekeeper) never garbage-collects the orphaned `<none>:<none>` build layers left behind by repeated 3-layer rebuilds. They accumulate without bound until the Podman VM's root disk fills to 100%, at which point **Podman goes completely down** — not just slow, fully offline. This is the mirror image of #418 (which is about removing *too much*); this issue is about removing *too little*.

## The outage (2026-07-02, real incident)

The `podman-machine-default` VM (applehv, 40 GiB disk) filled to **100%**:

```
/dev/vda4   40G   40G   132K  100%  /
```

Consequences, in order:

1. `podman.service` cannot open its sqlite state DB — with the disk full it can't create the WAL/journal file:
   `Error: updating container state: unable to open database file: no such file or directory`
2. `podman.service` crash-loops (exit 125) on every socket activation.
3. systemd trips the socket trigger-limit: `podman.socket: Trigger limit hit, refusing further activation` → `failed (Result: trigger-limit-hit)`.
4. **Every** host-side Podman call fails: `unable to connect to Podman socket: ... ssh: rejected: connect failed (open failed)`.
5. A `podman machine stop && start` does **not** fix it — a restart reclaims no disk.

### What was actually on the disk

```
overlay dir:  42G
image count:  260
  257  <none>:<none>          ← orphaned build layers, never pruned
    1  localhost/kapsis-slack-bot:latest
    1  localhost/kapsis-sandbox:latest
    1  localhost/kapsis-claude-cli:latest
dangling=true count:  0       ← the 257 are NOT dangling
```

Manual recovery required: free a few MB by hand (`rm -rf /var/cache/fwupd/*`) so the DB can open, then remove the 257 orphaned layers → disk 100% → 26%.

## Root cause: nobody GCs the orphaned `<none>` layers

Each `rebuild-bot-image.sh` run rebuilds `kapsis-sandbox → kapsis-claude-cli → kapsis-slack-bot`. When a new tagged image supersedes the old one, the old image's layers become **untagged (`<none>:<none>`) but non-dangling** (they still sit in a parent chain / retain child references), so:

- **`podman image prune`** (dangling-only) reports **0** — confirmed: `dangling=true` count was 0 with 257 `<none>` present.
- **The Slack-bot housekeeper explicitly skips them.** `_prune_podman_images()` (`skills/slack-cli/scripts/handlers/housekeeping.py:93-94`) does:
  ```python
  if "<none>:<none>" in ref:
      continue   # "so a concurrent rebuild's FROM <id> references stay intact"
  ```
  The intent (protect an in-flight rebuild's `FROM <id>`) is reasonable, but the effect is that orphaned build layers are **never** eligible for removal.
- **The triggers are count/size-based, not disk-based.** kapsis fires cleanup when image count crosses a threshold (bot default 50) or on inode pressure; neither reflects actual free bytes on `/dev/vda4`, and the dominant consumer (`<none>` layers) is skipped anyway. So cleanup "runs" and reports `0 removed` while the disk marches to 100%. Observed bot log the same day: `Image prune removed 0 images but count 260 exceeds limit 50 — all candidates may be in use / matched keep-list`.

This is the same intermediate-`<none>`-layer phenomenon called out in #418's failure case, viewed from the other end: #418 = the count threshold triggers cleanup that nukes in-use service images; **this issue = the layers that inflate that count are themselves never reclaimed**, so the disk fills.

## Proposed fix

**1. Age-based orphan GC for `<none>` layers.** Replace the blanket "skip all `<none>`" with "prune `<none>` images older than a grace window (e.g. 2h) that are not referenced by any container and are not an ancestor of a kept/allow-listed image." The grace window preserves a concurrent rebuild's fresh `FROM <id>` references (the original intent) without letting week-old orphans pile up. Podman naturally refuses to remove ancestors of tagged images, so allow-listed service images stay intact.

**2. Disk-usage trigger.** Add a free-space check on the VM root (`df /`) and escalate cleanup when usage crosses a high-water mark (e.g. ≥ 85%), independent of image count. Count-based limits do not track bytes.

**3. Post-rebuild cleanup hook.** After `rebuild-bot-image.sh` successfully tags the new service images, GC the now-superseded old layers in the same run (in-use-guarded), so orphans are collected at the moment they're created rather than waiting for a threshold.

**4. Guardrails (shared with #418).** All of the above must keep the in-use guard + `KAPSIS_KEEP_IMAGES` allow-list from #418 so the fix for under-removal doesn't reintroduce over-removal.

## Companion change (separate repo)

The Slack-bot housekeeper `_prune_podman_images()` skip at `housekeeping.py:93-94` needs the matching age-based logic (that repo is on Bitbucket, tracked separately). Filing here because kapsis owns the image lifecycle and `kapsis-cleanup.sh` is the shared cleanup entrypoint.

## Related

- #418 — `clean_images` removes in-use service images (over-removal). This issue is the complementary under-removal / disk-fill failure; a complete fix for the image lifecycle should address both together.

aviadshiber:
## Summary

Fixes the two mirror-image bugs behind the 2026-07-02 Podman outage: `kapsis-cleanup --images` destroyed images that WERE in use (#418) while never reclaiming the dangling `<none>` layers that were NOT in use and actually filling the disk (#421).

### Bug 1 — `clean_images` removes in-use service images (#418)

`clean_images()` enumerated every `kapsis-*` image and ran `podman rmi "$image_id" &>/dev/null || true` — no in-use check, exit code swallowed, and every candidate counted as "cleaned" even when removal failed. During the incident it removed the running `kapsis-slack-bot` service image.

**Fix — three guard layers, honest reporting:**

1. **In-use pre-check (primary):** image IDs referenced by ANY container (running or stopped, via `podman ps -a`) are skipped. If the query itself fails, cleanup **fails closed** — zero named-image removals for that invocation (required because `_vm_remediate` invokes `clean_images` unattended at CRITICAL inode pressure). ID formats are reconciled via `_normalize_image_id` (strips `sha256:`, truncates to 12 chars).
2. **Podman dependency refusal (secondary):** the `|| true` is gone and `--force` is never passed, so podman's native `image used by container` / `has dependent children` refusals transitively protect parent layers (e.g. `kapsis-sandbox` beneath an in-use child).
3. **Keep-patterns (tertiary):** `KAPSIS_IMAGE_KEEP_PATTERNS` (ERE vs `repository:tag`; default protects `kapsis-slack-bot`, `kapsis-claude-cli`, `kapsis-sandbox`). Set explicitly empty (`''`) to disable — unset-vs-empty handled via `${VAR-default}`.

Skipped/refused removals now print `[SKIPPED]` with a reason and are **never** counted in `ITEMS_CLEANED`, so the summary can no longer claim space was freed when it was not.

> **Deliberate behavior change:** working (in-use or keep-pattern-protected) images now survive `--images`/`--all`. Escape hatches: `KAPSIS_IMAGE_KEEP_PATTERNS=''` or direct `podman rmi`. Documented in `usage()` and `docs/CLEANUP.md`.

### Bug 2 — dangling build layers never GC'd proactively (#421)

Every image rebuild orphans the previous build's layers as `<none>:<none>` images; 257 of them piled up and filled the Podman VM disk to 100%. The only reclaim path was buried inside the destructive `--images` action.

**Fix — extracted `_prune_dangling_layers()` with three trigger points:**

- **`--prune-dangling`** — new standalone flag: cheap, natively in-use-safe (never `--force`), ideal for cron/downstream housekeepers; double-prune guarded when combined with `--images`/`--all`.
- **`--vm-health` WARNING tier** — `_vm_remediate` now prunes dangling layers as soon as health degrades to WARNING (threshold-based trigger), while the heavier `clean_images` stays gated to CRITICAL.
- **`clean_images`** — still prunes dangling layers, and deliberately does so **even when the in-use query fails**: `podman image prune` is natively in-use-safe, so reclaim capability survives a degraded podman (the actual space hog in the incident was `<none>` layers).

Only genuinely removed layers are counted (re-count after prune), and every mutation site has a `[DRY-RUN]` branch.

## Testing

- New `tests/test-cleanup-images.sh` (12 cases, quick/no containers) drives the real script against a PATH-shimmed fake `podman` with a call log: in-use skip, fail-closed on `ps -a` failure (with dangling prune surviving), keep-pattern default/empty semantics, rmi-refusal honest reporting, sha256:/64-char ID normalization, dry-run mutation-freedom, standalone `--prune-dangling`, dangling-query fail-closed, plus static assertions (swallowed-rmi pattern absent, no `rmi --force`, `_vm_remediate` WARNING-tier gating, flag registration). Registered in both the `validation` category and `QUICK_TESTS`.
- `shellcheck` clean on all touched scripts; `./tests/run-all-tests.sh --quick` green (incl. `test-cleanup-vm-health.sh`, `test-dry-run-completeness.sh`).
- Manual smoke on macOS with real podman 5.8.2: `--images --dry-run` skips all three protected service images; `--prune-dangling --dry-run` clean; verified `ps -a` vs `images` ID format assumption (both 12-char on 5.8.2; normalization is a no-op there and handles older 64-char/`sha256:` formats).

## Dashboard sync

No new `status.sh` fields or `audit.sh` event types (dashboard-sync CI unaffected). The new `--prune-dangling` CLI flag gets a follow-up issue tagged `dashboard-sync` per the CLAUDE.md rule (option 2).

Closes #421
Closes #418


aviadshiber:

No correctness, security, or quality issues were found. One test-coverage gap was confirmed.

### Critical (merge-blocking)

None.

### Non-critical

- **tests/test-cleanup-images.sh** (minor, test-coverage) — No behavioral test exercises the combined `--images --prune-dangling` (or `--all --prune-dangling`) double-prune guard in `scripts/kapsis-cleanup.sh:1857` (`if [[ "$PRUNE_DANGLING" == "true" ]] && [[ "$CLEAN_IMAGES" != "true" ]] && [[ "$CLEAN_ALL" != "true" ]]`). Verified against `origin/fix/421-image-gc-guard:tests/test-cleanup-images.sh`: the suite covers `--prune-dangling` standalone (`test_prune_dangling_flag_standalone`, line 277) and static registration/flag-wiring checks (lines 347–403), but no test runs `_run_cleanup --images --prune-dangling --force` (or `--all --prune-dangling`) and asserts `grep -c "image prune -f" "$CALL_LOG"` equals `1`. A future refactor of that guard (e.g. inverting the condition) could silently double-count/double-invoke the prune with no automated test catching it. Suggest adding one such case before merge or as fast follow-up.

None — the single substantive finding from the four reviewers was verified accurate against the branch tip and kept as non-critical.

---
Reviewed SHA: `34bf2c989c08f484e3f39686d40da7cd707af5cf`

aviadshiber:

Reviewed at `34bf2c9` (fix/421-image-gc-guard).

### Critical (merge-blocking)

1. **`tests/test-cleanup-images.sh:186-227,317-327` — RUN_RC exit-code assertions are dead code, always pass regardless of the script's real exit status.**
   Every call site invokes `_run_cleanup` inside a command substitution — `output=$(FAKE_PS_EXIT=1 ... _run_cleanup --images --force)`. Command substitution forks a subshell, so the `RUN_RC=$?` assignment made *inside* `_run_cleanup` (line 108) only updates the subshell's copy of `RUN_RC`; the outer `RUN_RC` (initialized to `0` at line 100) is never touched. Verified empirically with a minimal repro:
   ```bash
   RUN_RC=0
   _run_cleanup() { RUN_RC=0; false || RUN_RC=$?; }
   output=$(_run_cleanup)
   echo "$RUN_RC"   # prints 0, not 1
   ```
   This means `test_fail_closed_on_ps_failure` (line 199: "Script must still exit 0 on fail-closed path"), `test_rmi_refusal_not_counted_as_cleaned` (line 226), and `test_prune_query_fail_closed` (line 327) can never fail no matter what the script's actual exit code is. A future regression that makes any of these fail-closed paths crash or exit non-zero would ship silently with a fully green test suite. Fix: capture the exit code outside the substitution, e.g. redirect to a temp file and read `$?` directly in the caller's shell instead of inside a function invoked via `$(...)`.

### Non-critical

1. **`scripts/kapsis-cleanup.sh:1482-1497` (minor) — `_vm_remediate` double-prunes dangling layers on a CRITICAL-tier run, producing a misleading second "no dangling images" report.**
   Confirmed in code: line 1482 `if [[ "$VM_HEALTH_STATUS" != "HEALTHY" ]]; then ... _prune_dangling_layers; fi` fires for CRITICAL too (CRITICAL satisfies `!= HEALTHY`). Then line 1492 `if [[ "$VM_HEALTH_STATUS" == "CRITICAL" ]]; then ... clean_images; fi` runs seconds later, and `clean_images()` itself unconditionally calls `_prune_dangling_layers` again at line 886 ("Deliberately runs even when the in-use query failed"). Net effect on a CRITICAL run: the first call does the real prune and reports reclaimed space; the second (inside `clean_images`) re-queries and finds nothing left, printing "No dangling images to prune" in the Images section — confusing for an operator reading the CRITICAL-tier remediation output. The PR's own double-prune guard (`scripts/kapsis-cleanup.sh:1857`, `if [[ "$PRUNE_DANGLING" == "true" ]] && [[ "$CLEAN_IMAGES" != "true" ]] && [[ "$CLEAN_ALL" != "true" ]]`) only covers the standalone CLI-flag combination in `main()`, not this WARNING→CRITICAL sequential path inside `_vm_remediate` — the exact unattended auto-remediation scenario this PR is meant to harden. Not merge-blocking (functionally harmless — no `--force`, no incorrect state — just a confusing log line), but worth a follow-up: skip the WARNING-tier prune when already CRITICAL, or skip the internal prune in `clean_images` when it was just run.

2. **`tests/test-cleanup-images.sh` (major, but not independently blocking given the critical finding above) — no test exercises the combined `--prune-dangling --images` (or `--all`) invocation**, which is exactly the double-prune guard the PR description calls out as a deliberate new behavior (`main()` line ~1857). Only the standalone `--prune-dangling` and standalone `--images` paths are tested (`test_prune_dangling_flag_standalone` et al.). If the `!= "true"` guard is ever inverted or the `CLEAN_ALL` check dropped, `image prune -f` would run twice per invocation with no test catching it.

None — both reviewer findings (correctness-reviewer, test-reviewer) were confirmed against the code at the head SHA. security-reviewer and quality-reviewer reported no findings.

aviadshiber:
✅ **Round-1 critical resolved** (commit `51bdb70`): `RUN_RC` is now captured from `$?` after the command substitution at every call site, making the exit-code assertions live (12/12 passing locally, shellcheck clean).

PR #425, opened 2026-07-03
fix(entrypoint): skip fuse-overlayfs for config staging when FUSE is unavailable (#397 drift)

### LINKED ISSUE #419: entrypoint: `fuse-overlayfs` is invoked even though launcher no longer passes /dev/fuse (PR #397 drift) — noisy 'Overlay failed' on every staged config
## Summary

The launcher and the entrypoint disagree about overlay strategy after PR #397.

[`scripts/launch-agent.sh:2511-2515`](https://github.com/aviadshiber/kapsis/blob/master/scripts/launch-agent.sh#L2511-L2515) explicitly states:

> Issue #376 / PR #397 rework: the macOS named-volume overlay is assembled by the Podman runtime inside the VM (`:O,upperdir=...` on the volume's mountpoint — see lib/overlay-sandbox.sh), exactly like the Linux `:O` path. **No capability or device is added for it**: the cap-drop contract of every security profile stays intact in all sandbox modes.

But [`scripts/entrypoint.sh:670`](https://github.com/aviadshiber/kapsis/blob/master/scripts/entrypoint.sh#L670) still calls `fuse-overlayfs` directly for every staged config:

```bash
_overlay_stderr=$(fuse-overlayfs -o "lowerdir=${src},upperdir=${upper},workdir=${work}" "$dst" 2>&1)
```

Since `/dev/fuse` is no longer passed into the container, every invocation fails with:

```
Overlay failed for <path> (rc=1): fuse: device not found, try 'modprobe fuse' first
fuse-overlayfs: cannot mount: No such file or directory
```

The entrypoint then falls back to `atomic_copy_dir`, so the configs land — just without CoW isolation and with a wall of scary warnings on every launch.

## Reproducer

Any agent launched on macOS post-2.35.1 (or wherever PR #397 was integrated). Example from a Slack-bot integration (data from `~/.kapsis/conversations/aec7bb/transcript.txt`, run on 2026-06-26):

\`\`\`
[entrypoint] Overlay failed for .claude (rc=1): fuse: device not found, try 'modprobe fuse' first
[entrypoint] Overlay failed for .ssh (rc=1): fuse: device not found, try 'modprobe fuse' first
[entrypoint] Overlay failed for .slack (rc=1): fuse: device not found, try 'modprobe fuse' first
[entrypoint] Overlay failed for .kube (rc=1): fuse: device not found, try 'modprobe fuse' first
[entrypoint] Overlay failed for .config/.jira (rc=1): fuse: device not found, try 'modprobe fuse' first
[entrypoint] Overlay failed for .config/bkt (rc=1): fuse: device not found, try 'modprobe fuse' first
[entrypoint] Overlay failed for .config/mp (rc=1): fuse: device not found, try 'modprobe fuse' first
[entrypoint] Overlay failed for .jk (rc=1): fuse: device not found, try 'modprobe fuse' first
[entrypoint] Overlay failed for .candlekeep (rc=1): fuse: device not found, try 'modprobe fuse' first
\`\`\`

Scope: 64 of 91 recent agent transcripts in this deployment show the same pattern. Host VM has \`fuse\` loaded (\`fuse 233472 2 virtiofs\`) and \`/dev/fuse\` exists, but it's not propagated to the container — which is the explicit decision in PR #397.

## Impact

1. **Operational noise**: every launch logs ~9 alarming \`Overlay failed\` warnings that operators have to learn to ignore, masking *real* mount issues.
2. **Lost CoW isolation for staged configs**: \`.ssh\`, \`.config/bkt\`, \`.config/.jira\`, \`.kube\`, \`.claude\` end up as **atomic copies** instead of overlays. Writes inside the container modify the copy and never reach the host, which is the intended security boundary — but the cost is that any later code expecting a fuse-overlayfs mount sees a directory whose backing is just a copy. Cosmetic for most usage, but a latent footgun for anything that introspects the mount.
3. **Misleading debugging signal**: the cryptic \`fuse: device not found\` makes integrators chase phantom \`/dev/fuse\` config issues that don't exist.

### Option A (preferred): make the entrypoint follow the launcher's decision

Detect \`/dev/fuse\` (or test \`fuse-overlayfs --version\` for usability) once at startup. If absent, skip the fuse-overlayfs branch entirely and go straight to \`atomic_copy_dir\` with an \`INFO\`-level log (\"CoW overlay unavailable; staging via atomic copy\") instead of nine separate \`WARN\` lines.

\`\`\`bash
# Near top of setup_staged_config_overlays() in scripts/entrypoint.sh
local _have_fuse=false
if [[ -c /dev/fuse ]] && command -v fuse-overlayfs >/dev/null 2>&1; then
    _have_fuse=true
fi

for relative_path in \"\${configs[@]}\"; do
    ...
    if [[ -d \"\$src\" ]]; then
        if \$_have_fuse; then
            # current fuse-overlayfs path
        else
            # straight to atomic_copy_dir without warning
            atomic_copy_dir \"\$src\" \"\$dst\"
        fi
    fi
done
\`\`\`

### Option B: also use the Podman-native overlay for staged configs (full convergence with PR #397)

Mount each staged config as \`<host_path>:<container_path>:O,upperdir=...\` directly from the launcher instead of bind-mounting to \`/kapsis-staging\` and overlaying inside the container. Bigger change, but it would unify the design and restore CoW for staged configs on macOS.

## Bot-side context (no action required from kapsis)

In our slack-bot integration the noise is downstream of fix work-around — agents still run, credentials still land, but the warnings consistently mislead users (and us) into thinking the agent's auth setup was broken when it wasn't. Last week a Q2-self-review agent attributed missing Bitbucket/Jira/Thanos data to \"Kapsis sandbox blocks\" when the actual cause was a separate \`/workspace\` virtio-fs mount drop — these noisy fuse warnings made the misattribution easier.

## Related

- PR #397 (issue #376) — moved macOS overlay upper/work dirs off virtio-fs
- Issue #418 (filed by me last week) — \`clean_images\` allow-list

Happy to send a PR for Option A if it's the direction you'd take.

aviadshiber:
Closes #419

## Problem

After PR #397, the launcher no longer passes `/dev/fuse` or `SYS_ADMIN` into containers (grep of `scripts/` confirms zero matches). But `setup_staged_config_overlays` in `scripts/entrypoint.sh` still unconditionally invoked:

```bash
fuse-overlayfs -o "lowerdir=...,upperdir=...,workdir=..." "$dst"
```

for every staged config directory. Each invocation failed with `fuse: device not found, try 'modprobe fuse' first`, printing a wall of scary `Overlay failed` / `Falling back to atomic copy` warnings (~9 pairs per launch) before falling back to `atomic_copy_dir`. Configs still landed — only the CoW hop and the noise were the problem.

## Fix (minimal and surgical)

1. **New `_fuse_available` probe** next to `setup_staged_config_overlays`: returns success only if `/dev/fuse` exists AND `fuse-overlayfs` is on `PATH`.
2. **Computed ONCE** (cached in `fuse_ok`) before the staging loop, not per iteration.
3. **FUSE unavailable** (the norm post-#397): go straight to `atomic_copy_dir` for each staged config dir, with a single INFO line for the whole run — `FUSE unavailable — staging configs via atomic copy (expected post-#397)` — instead of a WARN pair per config. The issue-#328 `chmod u+rw` post-copy fixup is applied on this path too, exactly as it was on the old fallback path.
4. **FUSE available** (Linux setups that still pass `/dev/fuse`): the existing fuse-overlayfs path and its per-entry warn-and-fall-back behavior are byte-for-byte preserved.

## Explicit scoping (per issue history)

This PR deliberately does **NOT** touch `setup_fuse_overlay()`, `KAPSIS_USE_FUSE_OVERLAY`, or anything outside `setup_staged_config_overlays` and the new probe helper. A previous PR wrongly deleted `setup_fuse_overlay` as dead code and had to be reverted — it is exercised in Linux CI. The change here is confined to the staged-config installation loop.

## Tests

- **New quick test `tests/test-staging-fuse-probe.sh`** (no container required, registered in `run-all-tests.sh` under `--quick` and the `validation` category). It extracts the production `_fuse_available` + `setup_staged_config_overlays` from `entrypoint.sh` (same sed-extraction pattern as `test-workspace-mount-validation.sh`, with the hardcoded `/kapsis-*` paths rewritten to a temp root so the loop runs unprivileged) and asserts:
  - `_fuse_available` fails when `fuse-overlayfs` cannot be resolved
  - the probe is evaluated exactly **once** per staging run (3 staged configs, 1 probe call)
  - FUSE-unavailable path: single INFO line, **no** `Overlay failed` warnings, `fuse-overlayfs` never invoked, configs land in `$HOME` via atomic copy with content intact
  - FUSE-available path: `fuse-overlayfs` invoked once per staged directory, no copy fallback
  - FUSE-available + mount failure: pre-existing per-entry warning (with captured stderr) and atomic-copy fallback preserved
- **`tests/test-container-claude-staging.sh`** (container e2e) updated: its in-container `awk` extraction of `setup_staged_config_overlays` now also extracts `_fuse_available` so the helper resolves. That test asserted overlay-or-fallback as "both acceptable outcomes", so no behavioral assertions needed changing; full end-to-end coverage of the staging path remains in that container test.

## Validation

- `shellcheck` clean on all touched files
- `./tests/run-all-tests.sh --quick` passes
- `bash tests/test-staging-fuse-probe.sh` passes (4/4)

dashboard_sync: n/a — no status/audit surface change (log-noise fix only).


aviadshiber:

Reviewed SHA: `53bf94378c4678f1ea7d011c0e352bfc57848fc1`

### Critical (merge-blocking)
None. `setup_fuse_overlay()` and `KAPSIS_USE_FUSE_OVERLAY` are confirmed untouched by this PR (grep of the full diff finds zero references to either).

### Non-critical
- **tests/test-staging-fuse-probe.sh:82-131, 197-224** — The real, production `_fuse_available()` success path (FUSE device present + `fuse-overlayfs` binary installed → returns 0) is never exercised anywhere in the suite. `run_staging()`'s own doc comment (line ~79) advertises a `"real"` mode to keep the production probe live, but every actual call site (`run_staging 0 0`, `run_staging 1 1`, `run_staging 0 1`) stubs `_fuse_available` with a hand-written `eval` replacement before `setup_staged_config_overlays` runs, so `test_fuse_available_keeps_overlay_path` (line 199) validates the *consumer* logic (`fuse_ok` routing) but not the real `[[ -e /dev/fuse ]] && command -v fuse-overlayfs` condition. The only direct call to the unstubbed function, in `test_probe_fails_without_fuse_overlayfs_binary`, only checks the failure branch (binary missing via `PATH="$TEST_ROOT/empty-bin"`). Additionally, `tests/test-container-claude-staging.sh`'s two `podman run` invocations (lines 88 and 131) never pass `--device /dev/fuse`, so the container-level test also never reaches the real success branch. A regression that silently breaks the real success condition (e.g. `&&`→`||`, or an inverted device check) would pass CI cleanly. Consider adding one test that runs the real `_fuse_available` against a faked `/dev/fuse`-present + `fuse-overlayfs`-on-PATH environment, using `run_staging real <rc>` as the doc comment already implies.

None — the single scope-adjacent claim (test coverage gap) was verified against the diff and code and held up as described; it was downgraded from "major" to non-critical because it is a test-coverage gap in new test infrastructure, not a defect in the shipped fix logic (`_fuse_available`'s implementation itself is simple, unchanged elsewhere, and directly readable/reviewable at scripts/entrypoint.sh).

aviadshiber:

Reviewed SHA: `53bf94378c4678f1ea7d011c0e352bfc57848fc1`

Verified `setup_fuse_overlay()` (entrypoint.sh:782-865) and `KAPSIS_USE_FUSE_OVERLAY` — confirmed **completely untouched** by this diff. The change is scoped entirely to a new `_fuse_available()` helper and a one-time probe inside `setup_staged_config_overlays()`, exactly as intended for the #419 fix.

### Critical (merge-blocking)

None.

### Non-critical

- **tests/test-staging-fuse-probe.sh:137** (minor, test-reviewer) — `test_probe_fails_without_fuse_overlayfs_binary` only manipulates `PATH` and never controls for `/dev/fuse` presence:
  ```bash
  _fuse_available() { [[ -e /dev/fuse ]] && command -v fuse-overlayfs >/dev/null 2>&1; }
  ```
  Since `[[ -e /dev/fuse ]]` short-circuits the `&&`, on any host/CI runner without `/dev/fuse` (verified: reproduces on macOS — confirmed no `/dev/fuse` on this box) the assertion passes trivially without ever exercising the `command -v fuse-overlayfs` branch the test name promises to cover. Not merge-blocking since the overall behavior (probe fires once, routes to atomic copy) is still covered by the other tests in the same file, but the test's stated coverage claim doesn't hold on FUSE-less hosts. Consider stubbing/forcing `/dev/fuse` presence (or splitting into two independently-stubbed assertions) to make this test host-independent.

None — the sole test-reviewer finding was verified and kept as non-critical (see above). correctness-reviewer, security-reviewer, and quality-reviewer reported no findings.

Issue #382, opened 2026-05-20
Long-running containers silently wedge: `podman exec` hangs, status-sync stops, no exit-code-4 emitted (different mode from #376 launch-storm)


aviadshiber:
## Summary

On long-running kapsis containers we have observed a wedge mode that **none of the existing defences catch**:

- Container reports `Up` in `podman ps`.
- The agent process inside (`claude` PID 1-chain) is alive in `podman top`, burning ~zero CPU.
- `podman exec <container> echo hello` hangs (timeout fires after 10s).
- `/kapsis-status/*` files on the host stop being updated (status-sync.sh appears blocked).
- Liveness monitor never emits `KAPSIS_MOUNT_FAILURE` / exit code 4.
- vfkit is still alive — the v2.24.0 host-side vfkit watchdog correctly does not fire.
- `podman machine list` says `Currently running`, `podman version` works.

Net effect: the bot sees `phase: running` forever, no exit code, and only manual recovery (`podman machine stop && podman machine start`) unblocks it. Graceful machine-stop failed for me; podman fell back to hard stop.

## Why existing fixes don't catch this

| Defence | Status | Why it missed |
|---|---|---|
| v2.31.1 launch-lock (#375) | enabled | This was not a concurrent-launch storm. The two agents involved were ~5h apart. |
| v2.28.8 autoheal cascade (#348/#349) | enabled | Only one wedged container at the time; the autoheal correctly avoided VM restart because the peer was a different state (already exited). |
| v2.24.0 vfkit watchdog | enabled | vfkit did not die. |
| v2.23.8 liveness mid-run mount-check | enabled | Either the check passes from inside the VM (the in-VM mount is actually fine), or its sentinel write to `/kapsis-status/` is what's blocked, and never reaches the host. |
| v2.21.6 pre-launch probe + `probe_mount_readiness` | enabled | Both only run at startup, not mid-run. |

## Repro evidence (agent `22d209`, 2026-05-20)

Sequence on a Mac with kapsis 2.31.3, applehv VM, 16 GiB memory:

- 00:03 UTC: agent `0abf25` launched, ran 90 min, completed cleanly (exit 0, response written, container removed). Its launch-agent post-processing later hung in \`sync_index_from_container\` — separate symptom, possibly related.
- 05:30 UTC: agent `22d209` launched (~5h gap, so launch-lock had nothing to serialize).
[...]

Issue #257, opened 2026-04-19
Detect hung agent after completion — enable liveness + exit code 5 + descendant I/O monitoring


aviadshiber:
## Problem

Agent `4c96e9` completed its work (wrote response to conversation JSONL at 12:18) but the container process never exited. It ran for 2+ hours in a hung state before manual intervention. The liveness monitor did not detect it.

### Why the liveness monitor failed

1. **Liveness is OFF by default** in production configs — the monitor never ran
2. Even when enabled, the timeout was set to 10800s (3h) — too long for this failure mode
3. The monitor checks PID 1's I/O (`/proc/1/io`), but PID 1 is a bash wrapper — not the actual agent process. The bash wrapper's I/O may appear active due to child process accounting
4. The root cause was MCP server child processes (`@upstash/context7-mcp`) staying alive after Claude finished, holding inherited stdio FDs open. Node.js won't exit until all child stdio pipes close.

### Confirmed process state inside hung container

```
Claude (PID 4340): sleeping in epoll_wait, 0 TCP connections
  Child PID 4352: npm exec @upstash/context7-mcp  (MCP server, still alive)
  Child PID 7270: zsh -c <stuck bash tool call>
Response file: 0 bytes (--output-format text writes on exit only)
Conversation JSONL: contains full completed response
```

`CLAUDE_CODE_EXIT_AFTER_STOP_DELAY=10000` was set (launch-agent.sh:1866) but did not help — the MCP server keeps the process alive past the delay.

### 1. Enable liveness by default

Set `liveness: enabled: true` in default config with timeout 900s (not 1800s or 10800s).

### 2. Exit code 5 = "agent completed but process hung"

Current exit code registry:
- 0 = success
- 1 = agent error
- 2 = timeout (wall clock)
- 3 = liveness killed (agent idle, no completion)
- 4 = mount failure (virtio-fs drop)
- **5 = agent completed work but process didn't exit (NEW)**

Detection: liveness monitor kills agent AND conversation JSONL contains a recent assistant response → exit 5 instead of exit 3.

### 3. Monitor process tree descendants, not just PID 1

Currently `_liveness_get_io_total()` reads `/proc/1/io`. After the entrypoint's `exec`, the actual agent runs as a descendant of PID 1. Sum I/O across all descendants:
```bash
for pid in $(pgrep -P 1); do
  read_bytes=$(awk '/read_bytes/ {print $2}' /proc/$pid/io 2>/dev/null)
  # ... sum
done
[...]

PR #426, opened 2026-07-03
fix(watchdog): demote exec-channel watchdog to non-lethal degraded-state reporter with recovery

### LINKED ISSUE #414: exec channel wedge: 3 consecutive podman exec timeouts cause status blackout for rest of agent run
## Summary

During a long-running agent session (62 min, `products` worktree, exit code 0), Kapsis logged:

```
[2026-06-20 22:50:56] [WARN] [launch-agent] KAPSIS_EXEC_CHANNEL_HANG[exec_channel_watchdog]: 3 consecutive podman exec timeouts on kapsis-432b43 — daemon exec channel wedged
```

This appeared ~21 minutes into the run. The container continued executing and eventually exited normally (exit code 0, 56 file changes committed). However, because the exec channel was wedged, status reporting via `podman exec` was broken for the remaining ~40 minutes — the bot had no visibility into agent progress during that window.

## Environment

- Kapsis version: 2.35.1
- Host: macOS (Apple Silicon, AVF/virtio-fs)
- Host swap at launch: 83% (7711 MB / 9216 MB) — pre-flight warned about this
- Podman VM memory: 16384 MB
- Container: `kapsis-slack-bot:latest` (products worktree mode)
- Concurrent agents at the time: 3 active (`2f474f`, `432b43`, `d0f874`)

## Observed behaviour

1. Agent runs normally for ~21 minutes.
2. `exec_channel_watchdog` fires: 3 consecutive `podman exec` calls timed out.
3. Kapsis logs the `KAPSIS_EXEC_CHANNEL_HANG` warning and marks the exec channel as wedged.
4. The container itself keeps running — the AI agent is making progress (56 file changes in the end).
5. Status/progress reporting is blind for the remaining ~40 minutes.
6. Container exits 0. Post-container git operations run normally (though the push later failed due to a pre-receive hook rejecting the commit message — unrelated).

## Reproduction notes

- High host swap (83%) was present at launch. Pre-flight noted this widens the AVF virtio-fs cache race window.
- 3 agents were running concurrently, which may increase Podman daemon load.
- The watchdog triggered after ~21 min, not at startup — suggesting the exec channel degraded over time rather than failing from the start.

## Impact

- **Agent availability**: the exec channel wedge itself doesn't kill the agent — it keeps running. But the bot loses all progress visibility for the duration.
- **Ops concern**: if the exec channel is needed for anything beyond status reporting (e.g. mid-run file extraction, stop signalling), a wedge could have broader effects.

## Questions / requests

1. Is the exec channel watchdog's recovery strategy (mark as wedged, continue) intentional? Could it attempt a reconnect instead?
2. Does the wedge affect `podman stop` signals? When the bot later calls `podman stop kapsis-432b43` (e.g. on timeout), does a wedged exec channel interfere?
3. Is high host swap a known trigger for this? The pre-flight already warns about it — worth adding a harder recommendation to reduce concurrency under swap pressure?
4. Any way to surface a mid-run warning to the user when the exec channel wedges (e.g. a Slack message like "agent is running but status reporting is unavailable")?

## Log excerpt

```
[2026-06-20 22:50:56] [WARN] [launch-agent] KAPSIS_EXEC_CHANNEL_HANG[exec_channel_watchdog]: 3 consecutive podman exec timeouts on kapsis-432b43 — daemon exec channel wedged
[2026-06-20 23:19:48] [INFO] [launch-agent] Timer [container]: 2988s elapsed
[2026-06-20 23:19:48] [INFO] [launch-agent] Container exited with code: 0
[2026-06-20 23:20:20] [INFO] [launch-agent] Agent made 56 file change(s)
```

aviadshiber:
Closes #414

## Problem

The exec-channel watchdog (Issue #382) fired **terminally** after N consecutive `podman exec` probe timeouts: it wrote `exit_code: 4` + `error_type: exec_channel_hang` to status.json and SIGTERMed the agent's `podman run` process. Incident #414 proved both halves wrong:

- The **pkill was ineffective** during a daemon exec wedge — container stop must propagate through the very daemon channel that is wedged. The incident agent survived the kill and worked 40 more minutes.
- The **host-side terminal status write** was the direct cause of a false failure verdict: an agent that later exited 0 with 56 committed changes was reported as a mount-class failure.

## Fix: demote to a non-lethal degraded-state reporter

`scripts/lib/exec-channel-watchdog.sh` is now **observability only** — it never terminates the agent and never writes terminal status:

- Past threshold it enters **DEGRADED**: creates a host-only marker file under `$TMPDIR` (mtime refreshed on every tick — an independent heartbeat), logs a single `KAPSIS_EXEC_CHANNEL_DEGRADED` line, and **keeps probing forever** with exponential backoff capped by `KAPSIS_EXEC_WATCHDOG_BACKOFF_CAP` (default 300s, validated).
- On probe success it removes the marker, logs `KAPSIS_EXEC_CHANNEL_RECOVERED` with the degraded duration, and resets the counter and interval.
- Termination during a wedge is owned by the machinery that actually works: natural `podman run` exit, the vfkit watchdog (#303), the liveness monitor (#257), or the documented manual VM recovery. The lib header explicitly forbids reintroducing shared-VM restarts or unverified daemon-dependent corroboration gates.

`scripts/launch-agent.sh`:

- Sentinel renamed to `_EXEC_DEGRADED_MARKER` (`kapsis-<id>.exec-degraded`); stale-resume pre-clean and cleanup-trap removal retained.
- The post-container override is replaced by a **purely informational** log line that **never modifies EXIT_CODE** (the incident scenario — exit 0 with committed work — can no longer be reclassified).
- Exit 4 always classifies as `mount_failure`; the exec-hang conditional is removed (the vfkit/mount path is the only remaining producer).

`scripts/lib/status.sh`: `exec_channel_hang` documented as **legacy** — retained for historical status files, no longer emitted. The degraded signal deliberately lives out-of-band (host marker + log lines) because status-sync's whole-file volume mirror clobbers host-written status fields on recovery (macOS).

## Dashboard (in-PR sync)

- New `exec-channel` tail rule in `dashboard/server/src/store/health.ts` scanning the log tail for the DEGRADED/RECOVERED markers (same pattern as `KAPSIS_MOUNT_FAILURE`); last event wins. A genuine #382 wedge now surfaces within ~90s as a dashboard **degraded** state instead of a frozen screen or a false terminal failure.
- `terminalRule` keeps mapping legacy `exec_channel_hang` status files to `failed`.
- `types.ts` sync comment updated; no new status fields or error_type values.

## Tests

- `tests/test-exec-channel-watchdog.sh` rewritten to the new contract (15 cases): degraded marker + single DEGRADED line + no pkill + no terminal status, marker mtime heartbeat, recovery, backoff cap, counter reset, orphan protection, input validation (incl. the new backoff-cap knob), static invariant that the lib contains no `pkill` / `status_complete` / `podman machine` in executable code, and EXIT_CODE preservation (0 stays 0, 1 stays 1).
- **Fixes a latent CI gap**: the test file was registered nowhere and never ran — now in the `validation` category and `QUICK_TESTS`.
- `dashboard/server/tests/health-rules.test.ts` extended for the new rule (bun test: 230 pass).

## Guardrails honored

- Kill switch `KAPSIS_EXEC_WATCHDOG_ENABLED=false` retained.
- Backward compatibility: `exec_channel_hang` stays in the dashboard union.
- vfkit watchdog (#303) behavior untouched: sentinel, override block, and exit-4 classification unchanged.
- Corroborated auto-escalation for genuinely dead daemons is deferred to a follow-up (must first empirically verify which podman operations survive the wedge).

## Verification

- `shellcheck` clean on all touched scripts (project `.shellcheckrc`)
- `./tests/run-all-tests.sh --quick` passes, with `test-exec-channel-watchdog.sh` now actually executing
- `bash tests/test-exec-channel-watchdog.sh` — 15/15 pass
- `cd dashboard/server && bun test` — 230 pass; `bun run typecheck` clean


aviadshiber:

### Critical (merge-blocking)

1. **`dashboard/server/src/store/health.ts:127-140`** — The exec-channel dashboard rule loses the DEGRADED signal on long episodes.
   `computeHealth` only scans the trailing 64 KiB of the combined agent log (`startAt = Math.max(0, size - 64 * 1024)`), and `tailRules` derives `execChannel` purely from the last `KAPSIS_EXEC_CHANNEL_DEGRADED`/`KAPSIS_EXEC_CHANNEL_RECOVERED` line seen in that window. But `scripts/lib/exec-channel-watchdog.sh` logs `KAPSIS_EXEC_CHANNEL_DEGRADED` exactly once, on the transition into degraded (`if (( degraded == 0 )); then ... log_warn "KAPSIS_EXEC_CHANNEL_DEGRADED..."`); every later tick only refreshes the marker's mtime (`touch "$degraded_marker_path"`), it never re-logs. Since the log file keeps growing with normal agent activity while the daemon exec channel stays wedged (the PR's own incident narrative describes the agent working 40+ more minutes after the channel wedged), the one-time DEGRADED line scrolls out of the trailing-64 KiB window well before recovery, and `tailRules` then falls through to `execLast === null` → reports `state: "healthy"`. The dashboard silently flips a real, ongoing degraded episode back to a false "healthy" read for the remainder of a long episode — undermining the PR's stated goal of surfacing a #382-style wedge as a persistent dashboard state.
   Verified directly against the diff (`health.ts` lines 79-152) and `exec-channel-watchdog.sh` lines 196-253 (`degraded=0` guard around the `log_warn "KAPSIS_EXEC_CHANNEL_DEGRADED..."` call; subsequent ticks only `touch` the marker).

### Non-critical

- **`tests/run-all-tests.sh:78` / `tests/test-exec-channel-watchdog.sh`** (major, but not a correctness bug) — `test-exec-channel-watchdog.sh` is newly wired into `QUICK_TESTS` (and the `validation` category). The file contains two unconditional `sleep 6` calls (`test_watchdog_degrades_after_threshold`, `test_watchdog_does_not_degrade_on_transient_failure`), several 15s-capped `_wait_for_file`/`_wait_for_gone` poll loops (each backed by a real `sleep 1` per iteration) used repeatedly, an 8s orphan-exit poll, and `test_degraded_probe_backoff` polling up to 40s for 5 backoff-capped probes. Typical-case runtime for this one file alone is in the 45-55s range — a 5x+ blow-out of the documented `--quick` "~10s" budget on every pre-push run (per `CLAUDE.md`'s pre-push checklist). Since the file "was registered nowhere and never ran" before this PR (per the PR description), this PR is what makes the slow runtime real. Suggest either keeping it `validation`-category-only (off `QUICK_TESTS`) or shrinking to event-driven/tighter-capped waits.
- **`tests/test-exec-channel-watchdog.sh:665`** (minor) — `test_informational_branch_preserves_nonzero_exit/_preserves_exit_0/_silent_without_marker` exercise a hand-copied `_simulate_informational_branch()` reimplementation of the condition at `scripts/launch-agent.sh:3295-3296`, not the real script path (also absent from `test-post-container-exit-code.sh`). Matches the existing `test-vfkit-watchdog.sh` `_simulate_override` convention, so this is a latent drift risk rather than a present defect — a future edit to that branch (e.g. reintroducing an `EXIT_CODE` mutation) wouldn't be caught by these tests.
- **`tests/test-exec-channel-watchdog.sh:726`** (minor) — No test covers the start-of-run pre-clean of a stale degraded marker from a prior `--agent-id` resume. Verified `scripts/launch-agent.sh:3085-3086` does `rm -f "$_EXEC_DEGRADED_MARKER"` before starting the watchdog specifically to prevent a leftover marker from a killed/prior run making a resumed run appear degraded on startup — confirmed no test in the file creates a pre-existing marker and asserts it's cleared/ignored before the watchdog's own probes run (only the runtime orphan-protection case is covered).

None — all reported findings were verified against the diff/code and are included above (one critical, three non-critical).

aviadshiber:

### Critical (merge-blocking)

None. The one "major" finding raised by the test-reviewer is a real test-coverage gap but does not block merge on its own (see Non-critical).

### Non-critical

1. **Missing test for launch-agent.sh's stale-marker pre-clean on `--agent-id` resume** — `tests/test-exec-channel-watchdog.sh`
   - `scripts/launch-agent.sh` adds `_EXEC_DEGRADED_MARKER="${TMPDIR:-/tmp}/kapsis-${AGENT_ID}.exec-degraded"; rm -f "$_EXEC_DEGRADED_MARKER" 2>/dev/null || true` immediately before `start_exec_channel_watchdog`, explicitly to guarantee "a leftover marker cannot make a resumed `--agent-id` run appear degraded."
   - Verified: all tests in `tests/test-exec-channel-watchdog.sh` (confirmed by reading the file at the PR head and grepping the whole `tests/` tree for `exec-degraded`/`start_exec_channel_watchdog`) call `start_exec_channel_watchdog` directly with a fresh/caller-controlled marker path, or simulate the informational branch in isolation. None of them pre-create a stale marker file and then exercise the `launch-agent.sh`-level pre-clean-before-watchdog-start ordering. If that `rm -f` line were ever moved after `start_exec_channel_watchdog`, or dropped in a merge conflict, no test would catch the regression.
   - Recommend a follow-up test (can be a small addition, not blocking): pre-touch the marker path, invoke the launch-agent pre-clean snippet (or a thin wrapper exposing it), and assert the marker is gone before the watchdog can observe it.

2. **Stray `stderr` line in `test_degraded_probe_backoff`** — `tests/test-exec-channel-watchdog.sh:400`
   - `probes=$(wc -l < "$TEST_TMP/probe.log" 2>/dev/null || echo 0)` — on the first loop iteration, before the fake `podman` has written a probe, this leaks `.../probe.log: No such file or directory` to stderr even though the test still passes.
   - Reproduced by running the suite directly against the PR branch: `test_degraded_probe_backoff` prints the "No such file or directory" line and then reports PASS.
   - Not a correctness bug (redirection-open failure precedes the `2>/dev/null` taking effect for that specific error class within command substitution), but it's log noise that could mask a real future failure in the same test. Suggest `[[ -f "$TEST_TMP/probe.log" ]] && wc -l < "$TEST_TMP/probe.log" || echo 0`, or pre-touch `probe.log`.

None — both test-reviewer findings held up under verification (confirmed by reading the diff/branch and running the test suite). correctness-reviewer, security-reviewer, and quality-reviewer reported no findings, and spot checks (new tests registered in `tests/run-all-tests.sh`'s `QUICK_TESTS`, no `podman machine` invocation remaining in the watchdog lib) found nothing to add or contradict.

Issue #407, opened 2026-06-13
Audit-trail event for host-side commit artifact strip (#391/#394 follow-up)


aviadshiber:
## Context
PR #394 (fix(commit): prevent Kapsis infrastructure artifacts from landing in user commits, closes #391) added a host-side strip in `scripts/post-container-git.sh` that mutates user-facing commit content (removes injected `CLAUDE.md`/`AGENTS.md` gist blocks between Kapsis sentinels, `.claude/settings.json`, Maven `.bak` files). The ensemble review flagged that this host-side mutation is **invisible to the audit trail**.

What shipped in #394: a `stripped_injections` **status field** (so the dashboard surface is covered per the Dashboard Sync Rule). What was deferred: a proper **audit event** recording that Kapsis altered the committed diff.

## Why it was deferred (not a quick add)
`post-container-git.sh` runs host-side **after the container exits**. It sources `logging.sh`/`status.sh`/`constants.sh`/`sanitize-files.sh` but **not** `scripts/lib/audit.sh`. `audit.sh` exposes a host-callable `audit_log_event()`, but the per-session audit hash-chain is created/`audit_finalize`d during the in-container session. The open design question: does a post-exit host mutation **append to the (possibly already finalized) session chain**, or get recorded as a distinct host-side event class? Appending naively could break `audit_verify_chain`.

## Proposed work
- Decide the chain semantics for post-exit host events (extend the session chain vs. a separate host-event log read by `audit-report.sh`).
- Emit an audit event when a strip actually changes bytes, capturing: file path(s), which pattern matched (gist sentinel / settings.json / `.bak`), and byte counts removed.
- Keep it a no-op (no event) when nothing was stripped, mirroring the existing `stripped_injections` status behavior.
- Surface the new event type in `dashboard/` per the Dashboard Sync Rule (or open a `dashboard-sync` follow-up).

## Pointers
- `scripts/post-container-git.sh` — `strip_kapsis_injections()` / `commit_changes()`
- `scripts/lib/audit.sh` — `audit_init`, `audit_log_event`, `audit_finalize`, `audit_verify_chain`
- Status field already present: `stripped_injections`

_Tracked from the post-merge cleanup of the #394 ensemble review. https://claude.ai/code/session_01GyUwb9P9abPA2WnUAqnCUU_

aviadshiber:
[...]

PR #427, opened 2026-07-03
feat(audit): record host-side commit strip in a non-chained host-events sidecar (#407)

aviadshiber:
> **STACKED ON #413** — base is `claude/sweet-thompson-qf13bs`. Retarget to `main` after #413 merges.

## Summary

Issue #407: the host-side commit strip (#413) mutates the user's committed diff but leaves no audit record. This PR records it — without touching the session hash chain.

## Design: non-chained host-events sidecar

- New `audit_log_host_event()` in `scripts/lib/audit.sh` writes flat JSONL to `$KAPSIS_AUDIT_DIR/<agent-id>-host-events.jsonl`.
- **Structurally chain-safe**: the filename never matches `*.audit.jsonl`, so `audit_verify_chain`, the report file-resolvers, and the dashboard's `AUDIT_FILE_RE` never process it; records carry no `seq`/`prev_hash`/`hash` fields, so nothing can paint a false "chain valid" badge. Mirrors the existing `<agent-id>-alerts.jsonl` sidecar precedent.
- Emitted from `strip_kapsis_injections()` **only when bytes actually changed on disk**, per file: `action, file, blocks_stripped, bytes_removed, removed_sha256, proof_outcome, suspicious_blocks_preserved` — hashes/byte counts only, never removed content.
- **Failure-isolated**: sourcing guarded (`|| true`), call guarded by `declare -f` and `|| log_warn` — a broken audit path can never fail the commit. True no-op when `KAPSIS_AUDIT_ENABLED` is false.
- Sidecars are TTL/size-managed: added to both audit cleanup glob lists.
- `audit-report.sh`: new host-events section (text + JSON), path derived from agent-id directly (deliberately not copying the alerts-sidecar derivation, which mismatches its writer — see follow-up in #407 discussion).
- Dashboard: `HostEvent` type added to `dashboard/shared/src/index.ts`; `types.ts` sync comment updated (dashboard-sync CI). Server/UI surface is a follow-up issue.

## Drive-by fixes to #413 code (found while extending it)

1. **BSD sed incompatibility**: the `sed -e :a -e '/^\n*$/{$d;N;ba}'` cleanup blanks `new_content` under macOS BSD sed (command substitution swallows the failure) — i.e. a verified strip could **destroy file content on Kapsis's primary platform**. Replaced with portable awk.
2. **Per-file rewrite gating**: rewriting was gated on the *global* `stripped_count`, so a later unmodified file could be spuriously rewritten after an earlier strip. Now gated per file.

## Testing

- New `tests/test-host-strip-audit.sh` (9 cases: emission on change, no-emission when unchanged/disabled, chain-verifier isolation, cleanup globs, report rendering, failure isolation) — registered in `security` category + `QUICK_TESTS`.
- Also registered `tests/test-gist-injection-provenance.sh` (#413's test, previously unregistered so it never ran in CI).
- shellcheck `-S warning` clean; full `--quick` suite green; dashboard `bun test` 226/226.

Closes #407


aviadshiber:

Reviewed delta vs base `claude/sweet-thompson-qf13bs` (PR #413) at head `130c9813335a6921dea3342353e4defbe2b3644a`.

### Critical (merge-blocking)
None.

### Non-critical

1. **`scripts/post-container-git.sh:358`** — `expected_sha256` is embedded into the `detail` JSON blob unescaped, unlike `file_rel` which is passed through `json_escape_string` on line 356.
   - Verified the value's provenance: `expected_sha256` originates solely from `launch-agent.sh:1337` (`block_sha256=$(printf '%s' "$block" | sha256_hash)`), written to a `chmod 600`, host-private, non-bind-mounted, one-time-use proof file (`$TMPDIR/kapsis-<agent-id>-gist-proof`) that is deleted after read. In the normal code path this is always a 64-char hex digest — it cannot contain `"`, `\`, or control characters, so it cannot actually produce malformed JSONL through this PR's trust boundary. Downgraded from the original "critical" report to **minor**: still worth wrapping in `json_escape_string` for consistency/defense-in-depth with `file_rel`, but not merge-blocking.

2. **`tests/test-host-strip-audit.sh`** — the PR's stated safety guarantee ("a broken audit path can never fail the commit" — the `declare -f audit_log_host_event` guard, `source ... || true`, and `audit_log_host_event ... || log_warn` in `scripts/post-container-git.sh:344-360`) has no test coverage. No case forces `audit.sh` to fail sourcing, makes `audit_log_host_event` return non-zero, or unsets the function, then asserts `strip_kapsis_injections` still returns 0 and CLAUDE.md/AGENTS.md are still correctly stripped. **(major, confirmed)**

3. **`tests/test-host-strip-audit.sh`** — per-file counters introduced by this PR (`file_stripped`, `file_bytes_removed`, `blocks_stripped`) are only ever exercised with exactly one sentinel block per file (verified against `test_verified_strip_emits_one_event`, `test_strip_both_files_emits_two_events`, `test_per_file_rewrite_guard_regression`). Accumulation across 2+ blocks in the same file is untested. **(major, confirmed)**

4. **`tests/test-host-strip-audit.sh`** — no test greps the host-events sidecar for the actual stripped block text to confirm raw content is never written (only hashes/counts, per the PR description). Emission tests only check that CLAUDE.md itself no longer contains the sentinel. **(minor, confirmed)**

5. **`tests/test-host-strip-audit.sh:409-421`** — `suspicious_blocks_preserved` alongside non-zero `blocks_stripped` is only asserted via a hand-written literal JSON string fed directly to `audit_log_host_event` in `test_audit_report_rendering`, bypassing `strip_kapsis_injections` entirely. No test puts a verified block and a suspicious/rogue block in the *same* file and checks the emitted event reports the correct non-zero `suspicious_blocks_preserved` alongside `blocks_stripped:1`. **(minor, confirmed)**

### Verification notes
- Manually re-read the BSD-sed→awk rewrite (`scripts/post-container-git.sh:330-340`) and the per-file gating (`if [[ "$file_stripped" -gt 0 ]]`) that this PR intentionally fixes on top of #413. The awk trailing-blank-trim logic (`sub(/[[:blank:]]+$/, "")` per line, then trim trailing all-empty lines from the end only) preserves non-trailing content and matches the stated intent; both `printf` calls use `%s` (not the variable as a format string), so no format-string injection risk. No correctness issue found here.
- Test registration checked: `test-host-strip-audit.sh` is correctly registered in both `QUICK_TESTS` and the `security` category list in `tests/run-all-tests.sh`.
- Dashboard Sync Rule checked: `dashboard/server/src/types.ts` and `dashboard/shared/src/index.ts` are updated in this PR to add the `HostEvent` type, satisfying the rule.

- None — all four Phase-1 findings held up on verification; only the severity of the sha256-escaping finding was revised from critical to minor based on the value's provenance (host-generated hex digest, not attacker-reachable through this diff).

aviadshiber:

Reviewed head SHA `130c9813335a6921dea3342353e4defbe2b3644a` (delta vs base `claude/sweet-thompson-qf13bs`, PR #413).

### Critical (merge-blocking)
None.

### Non-critical

- **scripts/lib/audit.sh:533,566 (test-coverage gap)** — The `*-host-events.jsonl` glob added to both `_audit_cleanup_ttl` (~line 533) and `_audit_cleanup_size` (~line 566) has no test coverage. Verified: `tests/test-audit-system.sh::test_audit_cleanup_ttl` / `test_audit_cleanup_size_cap` in this branch (`origin/fix/407-strip-audit-event`) still only build fixtures for `*.audit.jsonl`, `*-alerts.jsonl`, and `*-report.txt` — no `*-host-events.jsonl` fixture or assertion. `tests/test-host-strip-audit.sh`'s 9 cases cover emission/no-emission/isolation/rendering only, none exercise TTL or size-cap eviction of the sidecar. The cleanup code path itself looks correct by inspection (same glob pattern as the other jsonl sidecars), but the PR's stated design goal ("sidecars are TTL/size-managed") ships unverified.

- **tests/test-host-strip-audit.sh:94 (test-coverage gap)** — No test asserts the raw stripped block content is *absent* from `${AGENT_ID}-host-events.jsonl`. Verified in `origin/fix/407-strip-audit-event`: `test_verified_strip_emits_one_event`, `test_strip_both_files_emits_two_events`, `test_audit_log_host_event_unit`, etc. assert presence/correctness of `removed_sha256`, `blocks_stripped`, `bytes_removed`, and secret-masking (`assert_false "grep -q 'sk-supersecret...' "`), but none grep the sidecar to confirm sentinel/gist block text (e.g. `KAPSIS_GIST_BEGIN`) never appears there. Current production code only interpolates hashes/counts, so this holds today, but the invariant ("never write removed content, only hashes/counts") is checked only positively, not negatively — a future regression that accidentally interpolates raw content would pass the existing suite.

None — both test-reviewer findings verified against the PR diff and held up. No findings from correctness/security/quality reviewers.

PR #454, opened 2026-07-12
refactor(build)!: make Maven extension pre-caching pluggable + fix vulnerable transitive jars

aviadshiber:
## Summary

Correction to #452 (already merged). Its `dependencyManagement` override for the `ge-cache` stage's transitive dependencies (dom4j, maven-core, commons-compress, commons-io, jetty-http/server, jsoup, plexus-archiver/utils) **did not work** — confirmed directly in the v2.38.3 release scan, which still showed every one of these at the original vulnerable versions (34→2 Critical improvement came entirely from the `.grype.yaml` Zulu fix, and the go-stdlib CVEs were fixed by the yq bump — but the Maven-cache jars were unaffected).

## Root cause

Verified locally with isolated test poms: `mvn dependency:resolve` downloads the exact same old jar versions whether or not a `<dependencyManagement>` override is present in the resolving project's pom — identical output either way. This is a real Maven-resolver limitation for this specific mojo/scenario, not a syntax mistake in the generated pom (which I also validated as well-formed XML back in #452).

## Fix

Replaced the ineffective `dependencyManagement` block with explicit `rm -rf` of the specific vulnerable version directories from `/opt/kapsis/m2-cache`, run right after the Maven resolve step. Scoped to exact version-pinned paths (e.g. `.../dom4j/dom4j/1.1`, not the whole `dom4j` tree), so this can't accidentally remove a different, already-safe version of the same artifact that happens to also be cached.

## Verification
Built just the `ge-cache` target locally (`podman build --target ge-cache ...`) and inspected the resulting image directly:
- All 13 flagged vulnerable version paths: **gone**.
- `gradle-enterprise-maven-extension:1.20` and `common-custom-user-data-maven-extension:1.12.5` themselves: **still fully cached** (jar+pom+sha1 present), confirming the actual caching purpose of this stage is unaffected.


aviadshiber:
## Ensemble review (added after initial push, per request)

Ran the same 4-agent review used on the earlier PR batch (correctness/security/quality/test reviewers, in parallel) against this fix. Strong 4/4 consensus on one core gap, plus one important security-scope finding:

- **Quality (critical) / Test (high)**: no automated verification that the `rm -rf` actually works — `rm -rf` on a missing path is a silent no-op, so this fix could regress exactly the way #452 silently did, with no CI signal.
- **Security (important)**: this only prevents the vulnerable jars from being *baked into the image*. Since `maven/isolated-settings.xml` is online (not `-o`), an agent running with network mode `filtered`/`open` can still cause Maven to re-resolve the identical vulnerable coordinate from the network the next time the GE/CCUD extensions' own dependency graph calls for it. Only `network: none` fully closes that path.
- **Correctness (medium, verdict APPROVE)**: the new `rm -rf` RUN wasn't wrapped in the same `ENABLE_JAVA`/`ENABLE_GRADLE_ENTERPRISE` guard as the resolve step; path correctness and multi-stage COPY semantics were confirmed fine.
- **Quality (minor)**: lost the original traceability comment; undocumented why multiple versions of the same artifact (commons-io, plexus-utils) appear in the removal list.

### Fixed
- Wrapped the removal in the same `ENABLE_JAVA && ENABLE_GRADLE_ENTERPRISE` guard.
- Added a fail-loud check: if a listed path still exists *after* `rm -rf`, the build fails.
- Restored traceability + added an explanation for the multi-version groups.
- Documented the network-re-resolution scope limitation directly in the Containerfile comment.
- Added `tests/test-ge-cache-security.sh` (asserts all 13 vulnerable paths absent + a positive check that the GE/CCUD extension jars are still cached), following the existing `tests/test-protoc-precache.sh` pattern.

### One thing I initially got wrong and had to walk back
I first also hard-failed the build if a listed path *didn't exist before* removal (to catch a stale list after a `GE_EXT_VERSION`/`GE_CCUD_VERSION` bump, per the reviewers' staleness concern). Rebuilding to verify this locally surfaced that Maven's transitive resolution here is **not fully deterministic build-to-build** — `plexus-utils:3.5.1`/`4.0.0` were pulled in on one build and absent on the next, with nothing in the repo changed. Hard-failing on that would have caused spurious CI failures unrelated to any real regression, so that specific check is now a warning instead of a fatal error (the post-removal check remains fatal). Documented this explicitly in the Containerfile comment.

Verified again after these changes: rebuilt the `ge-cache` stage twice (once `--no-cache`), inspecting the resulting layer directly both times.

aviadshiber:
## Scope expansion: made the Maven extension mechanism pluggable

Per discussion: the `ge-cache` stage (and everything downstream of it — config schema, entrypoint.sh, tests) hardcoded knowledge of exactly two Maven extensions by name (Gradle Enterprise + its Common Custom User Data extension). That's narrow, single-vendor functionality baked into shared product code — the engineering cost it's carried today (three separate bugs found/fixed in the last day alone) is disproportionate to how niche it is for the general Kapsis audience.

### What changed
- **Containerfile**: `ge-cache` stage renamed to `maven-ext-cache`. `GE_EXT_VERSION`/`GE_CCUD_VERSION` scalars replaced with a `MAVEN_EXTENSIONS` JSON array (mirroring the existing `JAVA_VERSIONS` array-via-`jq` convention) and a `MAVEN_EXTENSIONS_VULNERABLE_PATHS` array for the optional post-resolution cleanup. **Defaults are unchanged** (still GE + CCUD), so this is behavior-preserving out of the box.
- **`configs/build-config.yaml` + every `build-profiles/*.yaml`**: `build_tools.gradle_enterprise: {enabled, extension_version, ccud_version}` → `build_tools.maven_extensions: {enabled, extensions: [{group_id, artifact_id, version}], vulnerable_paths: []}`. **This is a breaking config-schema rename** — anyone with a custom `build-config.yaml` referencing the old key needs to update it (hence the `refactor(build)!:` commit and updated PR title).
- **`scripts/entrypoint.sh`**: the runtime "already cached?" check was hardcoded to `.../gradle-enterprise-maven-extension/1.20/...` specifically — silently wrong for any other version already. Replaced with a generic marker file.
- **`docs/CONFIG-REFERENCE.md`**: removed a `gradle_enterprise: {server_url, build_scans_enabled, remote_cache_disabled}` schema block that was **completely dead** — verified it's never read anywhere in the codebase. Replaced with an accurate description of what Kapsis actually does (pre-caching) vs. doesn't do (GE server config — that's the user's own project's concern).
- **`docs/BUILD-CONFIGURATION.md`**: added a "Maven Extensions" section; also fixed a literal `"Taboola Java development"` string in the profile table to generic wording — a smaller instance of the same over-specificity.
- Renamed `tests/test-ge-cache-security.sh` → `test-maven-extensions-security.sh`, updated `test-maven-auth.sh` and `test-build-config.sh` assertions to match.

### Verification
Rebuilt the `maven-ext-cache` stage in isolation three times:
1. Default config (GE + CCUD) — matches prior behavior exactly, vulnerable paths still removed.
2. Default config again with `--no-cache` — same result (confirms determinism of the *mechanism*, separate from the known non-determinism of *which exact transitive versions* Maven resolves).
3. **A completely unrelated extension** (`org.apache.maven.extensions:maven-build-cache-extension`, nothing to do with Gradle Enterprise) — built and cached correctly with **zero code changes, config only**. This is the actual proof of pluggability, not just a rename.

`tests/test-build-config.sh` (23/23) and `tests/test-maven-extensions-security.sh` pass.

PR #455, opened 2026-07-13
chore: remove org-specific data from public repo, add CLAUDE.md rule

aviadshiber:
## Summary

Kapsis is a public, general-purpose tool. Found and removed remaining Taboola-specific data in tracked files (your personal `configs/aviad-*.yaml` files are already gitignored and were never touched — this is only about what's actually public).

- **`configs/network-allowlist.yaml`**: removed hardcoded `git.taboolasyndication.com` / `artifactory*.taboolasyndication.com` / `ci.taboolasyndication.com`, replaced with commented-out generic examples. Added `configs/network-allowlist-personal.yaml` (gitignored via the existing `configs/*-personal.yaml` pattern) with the real domains restored, so nothing is lost.
- **`scripts/lib/git-remote-utils.sh`**: `detect_git_provider()` had `git.taboolasyndication.com` hardcoded as a special case (needed because that hostname doesn't contain "bitbucket"). Replaced with a `KAPSIS_BITBUCKET_SERVER_HOSTS` env var — same capability, no hardcoded org.
- **`scripts/lib/security.sh`**: reworded a comment naming "a Taboola corporate Mac" to generic phrasing.
- **`docs/CONFIG-REFERENCE.md`, `configs/build-profiles/java-dev.yaml`**: removed "Taboola Java development" wording and several `git.taboolasyndication.com` / personal-username / product-codename examples in doc/config examples.
- **Tests**: same cleanup in fixtures/assertions across 4 test files, plus a new test for the `KAPSIS_BITBUCKET_SERVER_HOSTS` mechanism.
- **`CLAUDE.md`**: added a "Personal/Organization Configuration Separation" section so this is a durable rule, not something to re-discover — where personal/org config belongs, preference for pluggable mechanisms over hardcoding, and placeholder conventions (`git.example.com`, `jon.d`, etc.) for docs/tests going forward.

## Verification
All 4 modified test files pass (95/95 total), shellcheck clean, both YAML files valid. Directly tested `detect_git_provider`'s new env-var-based detection: works with `KAPSIS_BITBUCKET_SERVER_HOSTS` set, correctly falls through to `unknown` without it, doesn't affect generic `bitbucket.*` hostname detection.

No file overlap with #454 (different subsystems entirely) — independent, mergeable in any order.


aviadshiber:
## Added: CLAUDE.md stale-data cleanup

Per follow-up request, also cleaned up CLAUDE.md (554 → 488 lines). Verified claims against the actual repo rather than just trimming for length:

**Real inaccuracies found and fixed:**
- Hardcoded version `(v2.34.2)` — actual latest is v2.38.3. Removed the specific version entirely (releases cut on every merge, so any hardcoded version goes stale within days) — points at `VERSION`/GitHub Releases instead.
- "111 test files" / "18 guides" — actual counts are 122 / 19. Replaced with a count command rather than a number that'll drift again.
- `scripts/lib/` listing was missing 2 files that already exist (`overlay-sandbox.sh`, `worktree-guard.sh`) and duplicated the Key Files table almost entirely — collapsed to a pointer at that table + `ls scripts/lib/`.
- **Dashboard Sync Rule was factually wrong**: said the `AgentStatus` TypeScript interface lives in `dashboard/server/src/types.ts` — it doesn't, that file just re-exports `dashboard/shared/src/index.ts`, which is where it's actually defined and where edits need to go.
- **Config Resolution Order was missing a real step** (`<project>/agent-sandbox.yaml`) that `scripts/launch-agent.sh` actually checks — verified against the source and added it.

**Compressed to breadcrumbs** (verified accurate, but duplicating exact thresholds/timing already in the implementing script's own header comments — Mount Failure Detection, Hung Agent Detection, Commit Failure Detection sections): kept the summary + actionable bits (recovery commands, config keys), replaced multi-paragraph internals with a pointer to the source file, since that copy can't drift out of sync with itself.

**Left alone** (verified accurate, not duplicated, genuinely useful process guidance): Release Artifact Rule, Dashboard Sync Rule's four-surface checklist, Push Fallback, Commit Artifact Filtering, Agent Partial Completion, exit-code table.

PR #463, opened 2026-07-26
fix: never stage .kapsis/ internal files during post-container commit (exit 6 dead-end)

aviadshiber:
## Bug

Kapsis runs an AI agent inside a container against a git worktree, then does a post-container `git commit` + push on the host. During that step, `commit_changes()` stages the agent's changed files "via explicit paths" from `git status --porcelain` — and this swept in Kapsis's own internal files living under `.kapsis/` in the worktree (`.kapsis/README.md`, `.kapsis/gist.txt`, the phase-progress file the agent writes to). These are untracked whenever the target repo's `.gitignore` has no `.kapsis/` entry.

`validate_staged_files()` then correctly detected the staged `.kapsis/` files as suspicious internal files, unstaged them, and — because they're treated as a security-class violation — **aborted the entire commit**:

```
[validate_staged_files:129] Found staged .kapsis/ internal files (should be ignored):
[validate_staged_files:131]   - .kapsis/README.md
[validate_staged_files:131]   - .kapsis/gist.txt
[validate_staged_files:242] Suspicious files removed from staging area; aborting commit so caller can retry with clean files.
[commit_changes:558] Staged file validation failed; aborting commit to prevent unsafe files from reaching the repo
```

Nothing ever "retries with clean files" — this is a dead end. `commit_changes` returns failure, `post_container_git` fails, exit code 6 (`KAPSIS_EXIT_COMMIT_FAILURE`) is surfaced to the caller.

### User-facing impact

The agent's real work is silently lost: no commit, no push, no PR branch, worktree left in a state the caller doesn't know how to recover from — despite the agent having done its job correctly.

## Fix

Stop staging `.kapsis/` in the first place, so the `validate_staged_files` check becomes a harmless backstop instead of a load-bearing abort trigger. In `commit_changes()` (`scripts/post-container-git.sh`):

- The explicit-path staging loop now filters `.kapsis/` out of the changed-files list before calling `git add`.
- The `git add -A` fallback path now excludes `.kapsis/` via pathspec (`':(exclude).kapsis/'`).
- Added a defensive `git reset -q -- .kapsis/` at the top of the function, since `sync_index_from_container()` runs beforehand and copies the container's index verbatim — which could already have `.kapsis/` files staged from inside the container.

`validate_staged_files()` itself is untouched — it still catches literal `~` paths, Kapsis git-mount files, and, as defense in depth, any `.kapsis/` files that somehow still reach the index.

## Tests

`tests/test-commit-failure-handling.sh`:
- Added `test_dot_kapsis_never_staged_and_commit_succeeds`: a worktree with `.kapsis/README.md` + `.kapsis/gist.txt` alongside real legitimate changes now commits successfully (exit 0), the real file lands in the commit, `.kapsis/` never gets staged or committed, and the files remain on disk untouched.
- Updated `test_commit_changes_returns_1_on_security_violation` to exercise a literal `~` path instead of `.kapsis/`, since `.kapsis/` is no longer stageable in the first place — the literal-`~`-path abort behavior is still genuine and unchanged.

Ran the full relevant test files locally (all green):
```
tests/test-commit-failure-handling.sh   14/14
tests/test-validate-staged-files.sh     15/15
tests/test-kapsis-artifact-filter.sh     9/9
tests/test-coauthor-fork.sh             29/29
tests/test-committer-identity.sh        17/17
tests/test-ephemeral-artifact-filtering.sh 9/9
tests/test-push-verification.sh         11/11
```
Also ran `tests/run-all-tests.sh` (full suite) — 616 `[PASS]`, 0 `[FAIL]` before it hit the 10-minute tool timeout partway through (suite includes long-running container/integration tests unrelated to this change).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -317,6 +317,13 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v6
+        with:
+          # Full history + tags: the version-management tests resolve the current
+          # version via `git describe --tags`. The default shallow (depth=1)
+          # checkout has no tags, so describe returns a stale ancestor tag and the
+          # --downgrade tests fail with "no older version found". See version.sh
+          # get_current_version().
+          fetch-depth: 0
 
       - name: Install test dependencies
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -521,6 +521,13 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v6
+        with:
+          # Full history + tags: the version-management tests resolve the current
+          # version via `git describe --tags`. The default shallow (depth=1)
+          # checkout has no tags, so describe returns a stale ancestor tag and the
+          # --downgrade tests fail with "no older version found". See version.sh
+          # get_current_version(). Same fix as the Unit Tests job above.
+          fetch-depth: 0
 
       - name: Configure git for tests
         run: |
--- a/scripts/post-container-git.sh
+++ b/scripts/post-container-git.sh
@@ -536,13 +536,30 @@ commit_changes() {
     # losing the agent's real work with no retry path. Filtering them out of the
     # staging source means validate_staged_files's .kapsis/ check is now just a
     # harmless backstop that should never actually trip.
+    #
+    # ".kapsis/" is a RESERVED worktree-internal directory; a target repo must not
+    # version-control its own directory of that name. It is excluded via THREE
+    # separate mechanisms whose syntaxes differ — keep them in sync if the name
+    # ever changes: (1) the defensive `git reset -- .kapsis/` above, (2) the
+    # `grep -v '^\.kapsis/'` filter just below, and (3) the `:(exclude).kapsis/`
+    # pathspec on the git-add-A fallback. validate_staged_files()'s own
+    # `^\.kapsis/` check is a 4th, final backstop.
     log_info "Staging changes..."
-    local changed_files
-    changed_files=$(git status --porcelain | cut -c4- | grep -v '^\.kapsis/' || true)
+    # Capture porcelain ONCE. The staged_count==0 branch below needs to tell an
+    # ephemeral-only .kapsis/ run (nothing substantive to commit) apart from a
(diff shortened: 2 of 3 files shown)


aviadshiber:
ShellCheck SC2088 on `test-commit-failure-handling.sh:314-315` was **suppressed, not expanded** (commit 47250ab), and that is deliberate. Those lines use a literal, unexpanded `~` on purpose — the fixture in `test_commit_changes_returns_1_on_security_violation` exists to assert that `commit_changes()` detects and rejects a staged literal `~` path as a security violation. Rewriting to `$HOME` would write outside `$TEST_REPO` and make `git add "$HOME/leaked"` fail with 'outside repository', so nothing would stage and the security-abort path would never trigger — silently gutting the test. Added per-line `# shellcheck disable=SC2088` with an inline explanation instead. All 14 tests still pass.

aviadshiber:
Ran the review ensemble (4 reviewers + a claim-verifier) on this PR and pushed fixes for the verified findings in 30ade86.

**Verified and fixed:**

- **Pure-`.kapsis/`-only runs now return rc=2 (ephemeral), not rc=1 (hard failure).** Two reviewers independently flagged that an agent run writing only `.kapsis/` gist progress and no real work would hit `staged_count -eq 0 → return 1` and dead-end as "Commit failed", bypassing the graceful `return 2` path a few lines down. The claim-verifier confirmed the control flow: it's pre-existing (not a regression this PR introduces), but it's the same dead-end class the PR set out to remove, so I closed it. Such a run now routes to rc=2, and the caller reports "no substantive changes". `git status --porcelain` is captured once so the ephemeral-vs-failure decision reads a single snapshot.
- **Two test-coverage gaps (both flagged high).** The defensive `git reset -- .kapsis/` path (`.kapsis/` already staged in the index, as sync_index_from_container leaves it) and the `git add -A -- . ':(exclude).kapsis/'` fallback were both untested. Added regression tests for each: the fallback test forces the fallback deterministically via a quoted non-ASCII filename and verifies it fired. A green result can't be meaningless.
- Documented `.kapsis/` as a reserved worktree-internal directory and enumerated the four independent exclusion layers so none gets removed by accident.

**Deferred to a follow-up:** case-insensitive `.kapsis/` matching. A security reviewer suggested `grep -i`, but that would silently drop a legitimately-distinct `.KAPSIS/` directory on case-sensitive Linux hosts. A correct fix has to be `core.ignorecase`-aware, which is more than this focused PR should carry.

17/17 tests in `test-commit-failure-handling.sh` pass, sibling suites (validate-staged-files, post-container-git) are green, and shellcheck is clean on both files.

PR #466, opened 2026-08-15
Worktree-mode in-container git + hardened host push/PR (fixes dead git + no-PR)

aviadshiber:
## Summary

Worktree sandbox mode had **dead in-container git**: the agent couldn't commit, diff, verify, or open a PR from inside the container, and `/dev`-style runs ended with a pushed branch but **no PR** (root cause: `entrypoint.sh` ran the agent without ever calling `setup_worktree_git`, so `GIT_DIR` was never exported; the sanitized git was also mounted read-only with a read-only objects symlink).

This PR (architecture **B** — see `designs/2026-08-09-worktree-in-container-push-design.md`):

- Gives the agent **working in-container git** for commit/diff/verify.
- Keeps the credentialed **push + PR creation + validation on the host** (reliable and outside the agent's trust domain — an in-container push can't fire on the SIGKILL / virtio-fs-mount-drop failure modes that motivated this, and would duplicate a host path that must exist anyway).
- Fixes the **"no PR"** symptom via a provider-pluggable **host-side post-push hook**.

Design decisions were pressure-tested across **two adversarial review rounds** (see `designs/…-design.md` "Ensemble review log"), which caught and corrected an earlier in-container-push pivot.

## Changes

**Writable in-container git (agent commit/diff/verify)**
- `entrypoint.sh`: run `setup_worktree_git` in worktree mode (was short-circuited by `[[ ==worktree ]] || setup_worktree_git`), so `GIT_DIR` is exported; guard `main` so the script is unit-test-sourceable.
- `launch-agent.sh`: mount the sanitized git **read-write**.
- `worktree-manager.sh`: objects become a **writable dir + `objects/info/alternates`** (borrowing the parent DB read-only) instead of a read-only symlink, so `git commit` can write new objects; `repoint_sanitized_git_objects` rewrites the alternate (legacy symlink path kept as fallback).

**Hardened host push/PR (`post-container-git.sh`, `status.sh`, `launch-agent.sh`)**
- `validate_staged_files`: **case-insensitive** `.kapsis`/`.git-safe`/`.git-objects` exclusion (container FS is case-sensitive) + reject staged **symlinks** (`120000`); gitlinks (`160000`) were already rejected. Fail-closed (no push on violation).
- `push_changes`: **fetch-before-push** divergence guard — refuse a non-fast-forward instead of emitting a bare rejection that stranded work as "committed, no PR".
- Provider-pluggable **`git.post_push_hook`** run on the host after a verified push (PR creation via `gh`/`bkt`/etc., supplied by config; agent-influenced values passed via env only). Outcome recorded as `pr_hook_status` (`skipped|ok|failed:<reason>`) and surfaced, never silent. Unset = legacy PR-instructions behavior.

## Tests

- New `tests/test-worktree-incontainer-git.sh` (GIT_DIR dispatch guard, rw mount).
- Rewrote `tests/test-sanitized-git-objects.sh` to the alternate model + a **functional commit-against-read-only-parent** test.
- Extended `tests/test-post-container-git.sh`: case-insensitive/symlink/gitlink rejection, fetch-before-push (fake bare remote), post-push hook (env, unset-skip, failure-surfaced).
- Affected suites green: worktree-incontainer 4, objects 9, post-container-git 14, post-exit-git 10, status-reporting 30, validate-staged-files 15, commit-failure 17, git-excludes 7, worktree-cleanup 26, coauthor-fork 29, push-verification 11.

## Not covered by unit tests (integration)

Whether `GIT_DIR` is effective for the *agent* process and container egress are exercised only in a real sandbox run — worth a manual verification before merge.

### DIFF
--- a/dashboard/server/src/types.ts
+++ b/dashboard/server/src/types.ts
@@ -37,4 +37,10 @@
 // type (string | null, defined in dashboard/shared/src/index.ts) is
 // unchanged. No schema change here; this comment satisfies dashboard-
 // sync.yml's file-presence check.
+// Worktree in-container git PR (feat/worktree-in-container-push) added a
+// provider-pluggable host-side post-push hook. scripts/lib/status.sh gained
+// AgentStatus.pr_hook_status ("skipped" | "ok" | `failed:${string}` | null) and
+// extended push_status with "diverged" (fetch-before-push guard). Both are
+// defined in dashboard/shared/src/index.ts; this note satisfies dashboard-
+// sync.yml's file-presence check.
 export * from "@kapsis/dashboard-shared";
--- a/dashboard/shared/src/index.ts
+++ b/dashboard/shared/src/index.ts
@@ -35,7 +35,12 @@ export interface AgentStatus {
   error: string | null;
   worktree_path: string | null;
   pr_url: string | null;
-  push_status: "success" | "failed" | "skipped" | "unverified" | null;
+  // Outcome of the provider-pluggable post-push PR hook (git.post_push_hook).
+  // "failed:<reason>" carries the failure reason (e.g. "failed:exit3").
+  // Optional (like transcript_content_missing) so existing status fixtures that
+  // predate this field still satisfy AgentStatus.
+  pr_hook_status?: "skipped" | "ok" | `failed:${string}` | null;
+  push_status: "success" | "failed" | "skipped" | "unverified" | "diverged" | null;
   local_commit: string | null;
   remote_commit: string | null;
   push_fallback_command: string | null;
--- a/dashboard/ui/src/views/AgentDetail.tsx
+++ b/dashboard/ui/src/views/AgentDetail.tsx
@@ -177,6 +177,9 @@ function OverviewTab({ status, health }: { status: AgentStatus; health: AgentHea
         <div className="sub">
           commit: <code>{status.commit_sha?.slice(0, 8) ?? "—"}</code> ({status.commit_status ?? "n/a"})
           <br />push: {status.push_status ?? "n/a"}
+          {status.pr_hook_status && status.pr_hook_status !== "skipped" && (
+            <><br />PR hook: {status.pr_hook_status}</>
+          )}
         </div>
       </div>
       <div className="card">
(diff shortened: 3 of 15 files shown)


aviadshiber:
### Ensemble review round (3 reviewers on the diff) — applied in 98ba1e3

**Fixed in-scope:**
- **Fetch-before-push guard** now evaluates divergence only on a *successful* fetch (captures the exit code, drops the stale `refs/remotes` fallback) so a failed fetch can't falsely refuse a legitimate push; added a first-push-of-new-branch test. (correctness + quality)
- **repoint symlink guard** — refuse a symlinked `objects/info` and `rm -f` the alternates target before writing, so a tampered symlink can't clobber a host file. (security HIGH, new from this diff)
- **Neutralize agent-planted hooks / `core.hooksPath`** in the rw-mounted sanitized dir before any host git op. Defense-in-depth for the writable-git change. (security)
- **Real functional test** for `setup_worktree_git` GIT_DIR export (run in a subprocess to dodge the readonly constant), replacing grep-only coverage of the headline behavior. (quality HIGH)

**Tracked follow-up (out of scope here):** #467 — host post-container git trusts the agent-writable `/workspace/.git`, a config/hook RCE surface. **Pre-existing** (the worktree root is mounted rw on `origin/main` too; verified) and outside this PR's honest-agent threat model, so filed separately rather than folded in. This PR only adds `_neutralize_sanitized_git_hooks` + the repoint guard for the *new* sanitized-dir plant surface.

Cleared by reviewers: in-container commits don't strand the host push; symlink/gitlink mode detection is correct; dashboard type sync (`pr_hook_status`, `"diverged"`, env-var names) matches end-to-end; post-push hook passes agent-influenced values via env only (no shell injection).

aviadshiber:
**Scope expanded per request:** the host-side hardening for #467 is now included in this PR (commit f5a488c), not deferred. `_restore_real_worktree_gitpointer` forces `/workspace/.git` back to the real host-side gitdir before any host git op, so an agent-repointed pointer can't make host git resolve agent-controlled config/hooks (config/hook RCE). Plus `-c core.hooksPath=/dev/null` on the host commit and an adversarial test (repointed pointer + planted post-commit hook does not execute on the host). Closes #467.

